PR #59, opened 2021-12-06
feat: add basic backend configuration code generation

### LINKED ISSUE #56: feat: add basic backend configuration code generation
As defined here: https://github.com/mineiros-io/terrastack/blob/main/docs/backend-config.md
Focused on just the basic usage/feature of generating code inside multiple stacks, but no metadata/globals yet (hence "basic").

katcipis:
Closes #56 

To test manually this script can be used, just a simple scenario but useful to see it working:

```
#!/bin/bash

set -o errexit
set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

mkdir -p envs/prod/stack
terramate init envs/prod/stack

mkdir -p envs/staging/stack
terramate init envs/staging/stack

terramate {
  backend "sometype" {
    env = "staging"
  }
}
EOM

terramate {
  backend "sometype" {
    env = "prod"
  }
}
EOM

terramate generate

echo "generated code for staging"
echo

cat envs/staging/stack/_gen_terramate.tm.tf

echo
echo "generated code for prod"
echo

cat envs/prod/stack/_gen_terramate.tm.tf
```

Output should be:

```
generated code for staging

// GENERATED BY TERRAMATE: DO NOT EDIT

terraform {
  backend "sometype" {
    env = "staging"
  }
}

generated code for prod

// GENERATED BY TERRAMATE: DO NOT EDIT

terraform {
  backend "sometype" {
    env = "prod"
  }
}
```


soerenmartius (approved):
From what I can see: LGTM (not a go expert tho)

katcipis:
> From what I can see: LGTM (not a go expert tho)

Thanks for the review man (L)

katcipis:
The example script shows that code is generated on any dir that has a terramate.tm.hcl file, this happens because how stack detection is done right now, generate lists all stacks to generate code for each of them and it uses the ListStacks function:

https://github.com/mineiros-io/terramate/blob/c24e9e37f1ae45bece2382911e7721c8484fb9eb/list.go#L26

Stack validation is just checking if the file exists, so dirs with a terramate config file, that are not a stack, will be listed as stacks:

```go
			if info.IsDir() {
				stackfile := filepath.Join(path, ConfigFilename)
				st, err := os.Stat(stackfile)
				if err != nil || !st.Mode().IsRegular() {
					return nil
				}

				entries = append(entries, Entry{Dir: path})
```

This PR is still good to go because this fix is on listing logic. For example "terrastack list" will also list non-stack dirs as stacks because they have a config inside.

generate.go
```diff
@@ -0,0 +1,179 @@
...
+
+func generateStackConfig(basedir string, configdir string) ([]byte, error) {
+	if !strings.HasPrefix(configdir, basedir) {
+		// check if we are outside of basedir, time to stop
+		return nil, nil
+	}
+
+	configfile := filepath.Join(configdir, ConfigFilename)
+
+	if _, err := os.Stat(configfile); err != nil {
```
i4ki, line 101: maybe check for `os.IsNotExists(err)` and return the error if the file exists but stat failed whatever the other reason.

generate.go
```diff
@@ -0,0 +1,179 @@
...
+		if tfcode == nil {
+			continue
+		}
+
+		genfile := filepath.Join(stack.Dir, GeneratedTfFilename)
+		errs = append(errs, os.WriteFile(genfile, tfcode, 0666))
+	}
+
+	if err := errutil.Chain(errs...); err != nil {
+		return fmt.Errorf("Generate(%q): %w", basedir, err)
```
i4ki, line 87: I ran on my test stacks (with an "after" block not supported in your branch) and I got:

```
$ terramate generate
2021/12/11 21:22:23 Generate("/home/i4k/src/mineiros/tf-stacks"): parsing config: malformed terramate block: invalid attribute "after": parsing config: malformed terramate block: invalid attribute "after"
```

It successfully generated the code for some stacks.

The error does not show the offending stack and file. I also think we can improve the message because the part `Generate(<path>)` is such internal/debug info. WDYT?

I'd expect an error message like:

```
2021/12/11 21:22:23 failed to generate code for stack "/home/i4k/src/mineiros/tf-stacks/stack-b": parsing config "terramate.tm.hcl": malformed terramate block: invalid attribute "after": parsing config: malformed terramate block: invalid attribute "after"
```

generate.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+// Generate will walk all the directories starting from basedir generating
+// code for any stack it finds as it goes along
+//
+// It will return an error if it finds any invalid terramate configuration files
+// of if it can't generate the files properly for some reason.
+//
+// The provided basedir must be an absolute path to a directory.
+func Generate(basedir string) error {
```
i4ki, line 45: As generate could partially work (generate until error), it'd be cool if you return the list of generated files (alongside the error, it that's the case) so we can have a nice output in the cli, like:

```
$ terramate generate
2021/12/11 21:22:23 failed to generate code for stack "/home/i4k/src/mineiros/tf-stacks/envs/prod/wg": parsing config "terramate.tm.hcl": malformed terramate block: invalid attribute "after": parsing config: malformed terramate block: invalid attribute "after"

Generated files:
[+] envs/prod/vpn/_gen_terramate.tm.tf
[+] envs/prod/vpc/_gen_terramate.tm.tf
[+] envs/dev/vpn/_gen_terramate.tm.tf

Exited with error 1
$
```

i4ki (approved):
lgtm! I left comments for improvements but can be addressed in upcoming PR if they make sense.

generate.go
katcipis: Thanks makes perfect sense, will create a task to address this ASAP :100:

generate.go
katcipis: Will do on a follow up PR

PR #76, opened 2021-12-09
fix: provide proper version on all builds

katcipis:
Right now building/installing from main using:

```
make install
```

Or installing using go install using the latest will always provide the version defined on VERSION, which is misleading (the version is not 0.0.3, it could be something else), also building locally would also provide the version defined on VERSION, which is also not true. 

The current mode of bumping the VERSION is considerably error prone (forgetting to update the file, which is versioned). With a more dynamic/parametric version we can have release automation that builds a binary with the correct version more easily (the automation could be a manual workflow on GH, or something else).

There is also some logic on the cli main to make it work gracefully when people use go install/go run with a specific version, like:

```
go install github.com/mineiros-io/terramate/cmd/terramate@v0.0.3
```

Since some logic from terramate requires a valid semver I used a semver with pre-release info attached to it as the default. 
When building with make build or go build you are going to get a devel version: "0.0.0-devel", Go for example uses "(devel)" as the version build info when there is no tagged release, so it reflects that but using semver.

You can still force a version using "VERSION", like this:

```
VERSION=0.0.4 make build
```

We can add commit info in the future on the default "0.0.0-devel+<commit>", haven't done here yet because it didn't seemed essential and can be done in a next step.


Makefile
```diff
@@ -2,6 +2,8 @@
 SHELL := /bin/bash -o pipefail -o errexit -o nounset
 
 addlicense=go run github.com/google/addlicense@v1.0.0
+version?=0.0.3-$(shell git rev-parse --short HEAD)
```
katcipis, line 5: I dont like the manual 0.0.3 for the same reasons of having the VERSION file, but at least this is used only for make related automation (dev stuff) + it is very easy to define a different version if you want.

cmd/terramate/cli/cli_init_test.go
```diff
@@ -50,6 +49,8 @@ func TestInit(t *testing.T) {
 		want   runResult
 	}
 
+	const tsversion = TerramateTestVersion
+
```
i4ki, line 53: `tmversion` ?

cmd/terramate/cli/cli_init_test.go
```diff
@@ -207,6 +208,7 @@ func TestInit(t *testing.T) {
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			s := sandbox.New(t)
+
```
i4ki, line 211: This empty line is leftover of other changes or added to improve readability?

i4ki (approved):
lgtm

cmd/terramate/cli/cli_init_test.go
katcipis: Im kinda stupid with adding newlines because it looks better to me, but I would not generalize that, can remove :laughing:

katcipis: Or in this case it could have been a leftover, anyway I tried to improve it now

Makefile
mariux: this should be read from VERSION file... automation scripts should never have hardcoded stuff...

cmd/terramate/main.go
```diff
@@ -17,17 +17,47 @@ package main
...
+	defineVersion()
 	err = cli.Run(wd, os.Args[1:], os.Stdin, os.Stdout, os.Stderr)
 	if err != nil {
 		log.Fatal(err)
 	}
 }
+
+func defineVersion() {
+	const (
+		defaultVersion = "v0.0.0-devel"
```
mariux, line 41: why `v`?

mariux (changes requested):
you have conflicts in version naming.. sometimes with `v` sometime without... also please do not hardcode version in makefiles... this can be read from the file you removed ;)

Makefile
katcipis: we could also automate a way to get the latest stable version of the project like this:

```
go mod download -json github.com/mineiros-io/terramate@latest
{
        "Path": "github.com/mineiros-io/terramate",
        "Version": "v0.0.3",
        "Info": "/home/katz/go/pkg/mod/cache/download/github.com/mineiros-io/terramate/@v/v0.0.3.info",
        "GoMod": "/home/katz/go/pkg/mod/cache/download/github.com/mineiros-io/terramate/@v/v0.0.3.mod",
        "Zip": "/home/katz/go/pkg/mod/cache/download/github.com/mineiros-io/terramate/@v/v0.0.3.zip",
        "Dir": "/home/katz/go/pkg/mod/github.com/mineiros-io/terramate@v0.0.3",
        "Sum": "h1:vzsfu8FX6IwF6D4MPeUJUdpqPlmMerPJ68Nt+EtfIsM=",
        "GoModSum": "h1:M7Gn5dOycbD/vShHG+xpBvImUirliXeaGVhjEncMIXg="
}
```

It gets the latest tagged version, so it would automatically assume the most updated latest stable version.

cmd/terramate/main.go
katcipis: because Go uses versions with v prefix, calling: debug.ReadBuildInfo() when you have an actual release will return a version with the v prefix, which then we need to remove, this default is a default for the Go version when it is not available (it is not always available), hence the v.

katcipis:
> you have conflicts in version naming.. sometimes with `v` sometime without... also please do not hardcode version in makefiles... this can be read from the file you removed ;)

1 - we want to move away from having to maintain the file + version it (like getting the latest stable version automatically)
2 - the conflict is caused by Go, semver versions in Go start with "v", the code mentioned is intended to represent a default Go version

i4ki (approved):
:rocket:

cmd/terramate/main.go
katcipis: actually on this case it is inconsistent indeed, when shifting from Go semver and actual semver I got confused, this one never gets the v prefix removed and is wrong, will fix. thanks for the catch :100:

i4ki (approved):
lgtm

katcipis:
> you have conflicts in version naming.. sometimes with `v` sometime without... also please do not hardcode version in makefiles... this can be read from the file you removed ;)

In the end since the cli already provides logic for a default devel version I just removed hardcoded/default version from Makefile. It still can be configured with an env var VERSION, but there is nothing hardcoded and for now the default for unversioned builds will be "0.0.0-devel". (On Go 1.18 commit if info will be added on the build info, so adding +<commit> will be possible/reasonably easy for builds with Go >= 1.18).

Makefile
```diff
@@ -2,6 +2,7 @@
 SHELL := /bin/bash -o pipefail -o errexit -o nounset
 
 addlicense=go run github.com/google/addlicense@v1.0.0
+buildflags=-ldflags "-X github.com/mineiros-io/terramate.Version=$(VERSION)"
```
mariux, line 5: VERSION not defined.

mariux (changes requested):

Makefile
katcipis: In this case it sets the Version to "", which then uses the default (same behavior as not passing any build flag, which needs to work too).

katcipis: It runs this:

https://github.com/mineiros-io/terramate/runs/4486806891?check_suite_focus=true#step:5:6

which produces:

https://github.com/mineiros-io/terramate/runs/4486806891?check_suite_focus=true#step:6:6

which is the default version when a version is not defined.

mariux: which is wrong as this is version 0.0.3 somethng

mariux: all version constraint checks will fail...

PR #80, opened 2021-12-11
refactor: simplify version check/initialization

katcipis:


version.go
```diff
@@ -25,19 +24,7 @@ import (
...
-	tfversionObj, err = tfversion.NewSemver(Version())
-	if err != nil {
-		msg := fmt.Sprintf(
-			"terramate version does not adhere to semver specification: %s",
-			err.Error(),
-		)
-		panic(msg)
-	}
-}
+var tfversionObj *tfversion.Version = tfversion.Must(tfversion.NewSemver(Version()))
```
i4ki, line 27: This way it's gonna panic like:

```
$ ./bin/terramate
panic: Malformed version: x.1

goroutine 1 [running]:
github.com/hashicorp/go-version.Must(...)
	/home/i4k/go/pkg/mod/github.com/hashicorp/go-version@v1.3.0/version.go:104
github.com/mineiros-io/terramate.init()
	/home/i4k/src/mineiros/terramate/version.go:27 +0x72
```
it lacks context, which version? is it a `terramate.tm.hcl` file with a wrong version in it?

katcipis: hmm yeah the extra context seems helpful.

PR #81, opened 2021-12-13
fix: make terramate inherit parent env vars

i4ki:


cmd/terramate/cli/cli.go
```diff
@@ -81,8 +81,15 @@ type cliSpec struct {
 // as far as the parameters are not shared between the Run calls.
 //
 // If a critical error is found an non-nil error is returned.
-func Run(wd string, args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) error {
-	c, err := newCLI(wd, args, stdin, stdout, stderr)
+func Run(
+	wd string,
+	args []string,
+	inheritEnv bool,
```
katcipis, line 87: Maybe it doesn't make sense, just sharing the idea, but wouldn't it make sense to just inject the env, like the args ? then isolated/controlled runs are easier and the "inheriting env" is just passing os.Environ(). Not sure it works on all use cases, so feel free to disregard.

katcipis (approved):
LGTM + commented on env inheriting but probably just a stupid idea

cmd/terramate/cli/cli.go
i4ki: the git wrapper has a config.InheritEnv option, so this was a straightforward fix. But in the future, if we need to inject specific env vars to git, then we can do it. WDYT?

katcipis: Make sense, lets re-assess if anything related to this comes up on the future :100:

PR #83, opened 2021-12-13
spec: hierarchical config + project config

katcipis:
Globals spec is being left as TODO since we have a different task to spec it.
Related: https://app.shortcut.com/mineiros/story/4218/spec-create-spec-for-hierarchical-nature-of-config-files

Tried to explain that we have the concept of overall hierarchy as a form to define semantics, but each different configuration has a different behavior. For hierarchical backend config it is already defined here: https://github.com/mineiros-io/terramate/blob/main/docs/backend-config.md

And for globals it will be coming soon.

For overall project definition, added just 2 initial/basic git related ones, we can add more later as necessary, and these are not hierarchical in nature (defined only once).


README.md
```diff
@@ -67,6 +67,13 @@ to use **latest** to play around, but not recommended for long term automation
 since you won't get the same build result each time you run the install command.
 
 
+## Configuring Your Project
+
+Checkout the docs on [overall configuration](docs/config.md) for details
+on how different kinds of configurations can help you to avoid duplication
+and write solid infrastructure as code.
```
soerenmartius, line 74:
```suggestion
and write solid Infrastructure as Code.
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
+# Terramate Configuration
+
+There is a series of different configurations that can be done
+on Terramate, ranging from avoiding duplication to controlling
+order of execution of stacks.
+
+In order to do so Terramate works with configuration files named
```
soerenmartius, line 7:
```suggestion
In order to do so, Terramate works with configuration files named
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+**terramate.tm.hcl**. These files can be found on any directory
```
soerenmartius, line 8:
```suggestion
**terramate.tm.hcl**. These files can be found in any directory
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+of a Terramate project, but some configurations will be defined
+only once for an entire project, while others provide merge/overriding
+strategies, while others control only behavior specific for
+stacks.
+
+Here is defined the different kinds of configurations and
+how you can learn more about them. But before getting into the different
+kinds of configurations, what would be a Terramate project ?
```
soerenmartius, line 16:
```suggestion
kinds of configurations, what would be a Terramate project?
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+
+A Terramate project is essentially any collection of terraform code
+organized into stacks. You can have all terraform code together in a single
+directory but that would defeat the purpose of most Terramate features, so
+usually you have a set of stacks and maybe a set of terraform modules if they
+are not maintained in a different repository.
```
soerenmartius, line 22:
```suggestion
A Terramate project is essentially any collection of Terraform code
organized into stacks. You can have all Terraform code together in a single
directory but that would defeat the purpose of most Terramate features, so
usually, you have a set of stacks and maybe a set of Terraform modules if they
are not maintained in a different repository.
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+
+It is not a hard requirement for Terramate to work that the project uses git 
+for version control, but features like change detection do depend on git to
+work and will fail if this requirement is not met.
+
+So in general, a Terramate project looks like this:
```
soerenmartius, line 28:
```suggestion
In general, a Terramate project looks like this:
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+
+* A git project
+* The git top level dir is the project base dir
```
soerenmartius, line 31:
```suggestion
* The git top-level dir is the project base dir
```

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+* Stacks will be organized as different directories
+* Configuration may be present on any dir
+
+# Project Configuration
+
+Per project configuration can be defined only once at the project base dir.
+
+Available project wide configurations can be found [here](project-config.md)
```
soerenmartius, line 39:
```suggestion
Available project-wide configurations can be found [here](project-config.md)
```

docs/project-config.md
```diff
@@ -0,0 +1,36 @@
+# Project Config
+
+Terramate does not require any configurations to work, but in case default
+behavior is undesirable these configurations allows you to override it
```
soerenmartius, line 4:
```suggestion
behavior is undesirable, these configurations allow you to override it
```

soerenmartius (changes requested):
great job!

docs/config.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Stack Configuration
+
+Any dir that has no child dirs (a leaf in the project hierarchy) and
```
i4ki, line 43: is this a requirement for the new design? Can't we use the `stack {}` block for the definition of a stack? It seems odd that a stack needs to be a leaf, this would conflict with people's way of organizing IaC.

katcipis: This is being discussed here: https://app.shortcut.com/mineiros/story/4218/spec-create-spec-for-hierarchical-nature-of-config-files#activity-4367

Quite open to changing the spec if we agree on using stack{}

katcipis: Updated the spec according to what was discussed here: https://mineirosio.slack.com/archives/C02GFAAAKPT/p1639504351015800

docs/config.md
```diff
@@ -0,0 +1,68 @@
...
+Per project configuration can be defined only once at the project base dir.
+
+Available project-wide configurations can be found [here](project-config.md)
+
+# Stack Configuration
+
+Before talking about stack specific configuration, lets define what is a
+Terramate stack:
+
+* A dir that has no child dirs (a leaf in the file system tree)
```
mariux, line 46: i would rather say leaf-config/lead-stack as being the last config inside a directory, there can be other directories inside a leaf-stack  e.g. source code of a cloud/lambda function or examples or documentation.. all that is not another stack.

mariux (changes requested):
minor things only.

we should maybe define the difference between root-config, leaf config and intermediate config files.


leaf config might not be needed forever... but for the state being 
- a stack block can only be in a leaf config and 
- a config block can be only on the root config..
- upcoming globals can be anywhere and will be merged top to bottum for example..
- and alreadu defined backend will be replaced top down e.g. lower level backend config will have hgher priority but no merging will be done.
- required version can be anywhere? or just in root? or anywhere but needs to exist in root no matter what?


for upcoming features this merge strategy and precedence strategies needs to be defined for everything that is reflected in the config file.

wdyt?

docs/config.md
```diff
@@ -0,0 +1,68 @@
...
+In general, a Terramate project looks like this:
+
+* A git project
+* The git top-level dir is the project base dir
+* Stacks will be organized as different directories
+* Configuration may be present on any dir
+
+# Project Configuration
+
+Per project configuration can be defined only once at the project base dir.
```
mariux, line 37: :+1:

docs/config.md
```diff
@@ -0,0 +1,68 @@
...
+* Has a valid terramate configuration file (**terramate.tm.hcl**)
+* The terramate configuration file has a `stack {}` block on it
+
+Stacks have configurations that are particular to them, like these:
+
+* [Execution Ordering](execution-order.md)
+
+Even though Terramate metadata is not exclusively about stacks and can be
+used on any hierarchical configuration, it does provide information that can
+be useful to stack configuration, more details [here](metadata.md).
```
mariux, line 56: metadata does only exist in the context of stacks... so if we reference a meta data field in an intermediate or root config, it will be interpreted differently for different stacks...

we could also allow both.... e.g. terramate.stack.path is the path in context of stack.. while terramate.local.path is the current directory  (local is bad naming... could not come up with something better w.g. terramate.dir.path) ... vs. maybe terramate.pwd.path is the current directory (or changed via `--change-dir`) ... all valid information ? wdyt?

docs/config.md
katcipis: Yeah I meant this here: https://mineirosio.slack.com/archives/C02GFAAAKPT/p1639505791023500?thread_ts=1639504351.015800&cid=C02GFAAAKPT

But then I didn't get the concept of "non-leaf config" from @i4ki question/your answer after my proposal, he asked [here](https://mineirosio.slack.com/archives/C02GFAAAKPT/p1639505942025000?thread_ts=1639504351.015800&cid=C02GFAAAKPT):

"but should we fail if someone puts a stack block in a non-leaf dir?"

And you [answered](https://mineirosio.slack.com/archives/C02GFAAAKPT/p1639505973025500?thread_ts=1639504351.015800&cid=C02GFAAAKPT):

"yes.... stack in non-leaf should be an error..."

Since he was talking about "non-leaf dir" it seemed we actually wanted stacks to be on the fs leaf.

docs/project-config.md
```diff
@@ -0,0 +1,36 @@
+# Project Config
+
+Terramate does not require any configurations to work, but in case default
+behavior is undesirable, these configurations allow you to override it
+on a per project basis.
+
+Project wide configuration is done via a **config** block inside the
+**terramate** block:
```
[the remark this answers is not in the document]
katcipis: On the previous overall config docs I already defined that, but no harm on repeating here for clarity :-)

katcipis:
> minor things only.
> 
> we should maybe define the difference between root-config, leaf config and intermediate config files.

I focused on explaining the difference between:

* project config (root-config)
* intermediate config (hierarchical configs)
* stack config (leaf-config)

Maybe I can just add that stacks can't have other stacks inside them, anything other than that is allowed.

> 
> leaf config might not be needed forever... but for the state being
> 
> * a `stack` block can only be in a leaf config and

Will add.

> * a `config` block can be only on the root config..

This one is defined [here](https://github.com/mineiros-io/terramate/pull/83/files#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR35).

With an initial doc for git configs [here](https://github.com/mineiros-io/terramate/pull/83/files#diff-15c9a0f4080d21f318255d23fc1f77a641e5fb83c73d5521446595a912dfe986R1).

> * upcoming `globals` can be anywhere and will be merged top to bottom for example..
> * and already defined backend will be replaced top down e.g. lower level backend config will have higher priority but no merging will be done.

Defined [here](https://github.com/mineiros-io/terramate/blob/main/docs/backend-config.md).

> * `required_version` can be anywhere? or just in root? or anywhere but needs to exist in root no matter what?
> 
> for upcoming features this merge strategy and precedence strategies needs to be defined for everything that is reflected in the config file.

Yup, AFAIK the strategy depends on the nature of the configuration, that is why each configuration docs will define that for itself (like backend config VS globals). Tried to explain that [here](https://github.com/mineiros-io/terramate/pull/83/files#diff-77b3519370ff48f6caf7c316f0ece966f2f0871c40e9381d451d2c5b1c76b40fR60).

> 
> wdyt?

Makes sense :+1:

docs/config.md
katcipis: I was not sure and was leaving each metadata docs to define where they are evaluated, but it does make sense that it will always be evaluated on the context of a stack, will update.

i4ki (approved):
lgtm :rocket:

mariux (approved):
lgtm

PR #85, opened 2021-12-13
fix: module detection not working properly

i4ki:


katcipis (dismissed):
:rocket:

cmd/terramate/cli/cli_test.go
```diff
@@ -75,6 +75,63 @@ source = "%s"
...
+	git.CheckoutNew("change-the-module-2")
+
+	mod2MainTf.Write("# changed")
+
+	git.CommitAll("module 2 changed")
+
+	cli := newCLI(t, s.BaseDir())
+	want := stack.RelPath() + "\n"
+	assertRunResult(t, cli.run(
+		"list", s.BaseDir(), "--changed"),
```
katcipis, line 130: I think we don't need the basedir as a parameter anymore since we changed the design to allow injecting the working dir. BUt definitely not blocking/essential/etc, just sharing.

katcipis (approved):

PR #92, opened 2021-12-16
feat: add support to terraform functions on backend configs

katcipis:
Basically copied code from terradude: https://github.com/mineiros-io/terradude/blob/743c9e390ba6d6e945a87ff6e950050408dc3ff3/dude/dude.go#L30

There is a caveat, I copied the same version of terraform lib: https://github.com/mineiros-io/terradude/blob/743c9e390ba6d6e945a87ff6e950050408dc3ff3/go.mod#L7

The current stable version >= 1.0 broke the API and seems to have made almost everything internal:

https://github.com/hashicorp/terraform

We can dig later on how to have this functionality with terraform >= 1.0 (maybe the lib is somewhere else, didn't dug much yet).


mariux (commented):
that was fast ;) hehe

i4ki (approved):
lgtm

PR #93, opened 2021-12-16
docs: add install improvements

katcipis:
When installing things from @main it seems that Go proxy introduces some caching/staleness, we were not able to get the latest code until we disabled it (me and @mariux ). It didn't seem to be a caching issue, I cleaned the cache with `go clean` and also deleted all local versions of the packages, it was still downloading an old version of the package from the proxy server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,17 @@ go install github.com/mineiros-io/terramate/cmd/terramate@<version>
 ```
 
 Where **<version>** is any terramate [version tag](https://github.com/mineiros-io/terramate/tags),
-or you can just install the **latest** using go install:
+or you can just install the **latest** release:
 
 ```
 go install github.com/mineiros-io/terramate/cmd/terramate@latest
+```
+
+To install the latest code on main:
+
+```
+export GOPROXY=direct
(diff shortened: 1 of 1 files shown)


mariux (approved):
lgtm

README.md
```diff
@@ -50,20 +50,27 @@ go install github.com/mineiros-io/terramate/cmd/terramate@<version>
...
+
 Or if you have the project cloned locally just run:
 
 ```
 make install
 ```
 
 We put great effort into keeping the main branch stable, so it should be safe
-to use **latest** to play around, but not recommended for long term automation
+to use **main** to play around, but not recommended for long term automation
```
i4ki, line 73: isn't just exporting GOPROXY enough?

i4ki (approved):
lgtm

README.md
katcipis: Enough for what ? =P

i4ki: enough to solve the cache issue :laughing:

katcipis: ahhhh, I think so. From my local tests it didn't seemed related to caching at all, I deleted all the cache + all local versions of modules and yet it was downloading an older version x_x.

PR #94, opened 2021-12-16
spec: create spec for global variables

katcipis:
Some files linked on the docs only exist at: https://github.com/mineiros-io/terramate/pull/83
But as soon as the PR is merged it should be OK :-)


mariux (changes requested):
globals should have the following features:
- support expansion of metadata `terramate.*` namespace
- support multiple `globals {}` on the same level, duplicate attributes are considered conflicts an fail.

i4ki (dismissed):
lgtm but the name `globals` seems a bit off. Globals can be seen everywhere, and they have this property that they must be declared in the top-level scope only.

The `globals` in this spec works exactly the same as normal *scoped variables* or *environment variables* but they come from file tree instead of process tree. 

We can call them just *variables* (terramate variables) and each directory creates a scope. So they are scoped variables, and inner directories inherit the scope of outer directories, like scope blocks in programming languages. With this concept, the variables declared at the git root directory would be the `globals` (as they would be visible everywhere in the project) but we don't need to even call them like that.

katcipis:
> lgtm but the name `globals` seems a bit off. Globals can be seen everywhere, and they have this property that they must be declared in the top-level scope only.
> 
> The `globals` in this spec works exactly the same as normal _scoped variables_ or _environment variables_ but they come from file tree instead of process tree.
> 
> We can call them just _variables_ (terramate variables) and each directory creates a scope. So they are scoped variables, and inner directories inherit the scope of outer directories, like scope blocks in programming languages. With this concept, the variables declared at the git root directory would be the `globals` (as they would be visible everywhere in the project) but we don't need to even call them like that.

I think the idea of not using the name **variables** is just to avoid any confusion with terraform variables, since the behavior is different from classic terraform variables AFAIK. Like access now would be global.<name>, if we use var.<name> it matches terraform vars usage and maybe that is misleading.

But the reasoning that it works like hierarchical scope of most languages makes sense. 

WDYT @mariux ?

i4ki:
@katcipis yeah I remember that, in lua global is the default and the `local` keyword is for scoped vars, very weird. I meant that it's what most languages implement.

katcipis:
> @katcipis yeah I remember that, in lua global is the default and the `local` keyword is for scoped vars, very weird. I meant that it's what most languages implement.

Python is also like that, with the inverse behavior, locals are default and globals are explicit (global keyword), but you still can declare a global anywhere =P (not necessarily on top level). Now that I'm thinking about it, all dynamic languages that I know that have globals allow globals to be defined anywhere on the code :thinking:

i4ki:
Ok, I wasn't clear maybe the right word is not "declared" because we often think about the declaration syntax but I'm talking about the semantics. Even if you use `global a = 1` in a function, it sets the value to the top-level scope always, not to the parent scope. The globals is the scope that is always visible to all other scopes, ie there's only 1 unique global list of variables, but terramate globals are affected by scope (or the directory the stack is) hence it's strange.

mariux (approved):
lgtm

mariux:
the bottom line is evaluation starts in the stack based on what is true in the stack... ofc this is exactly the same result as evaluating every step down... as the same rules will overwrite values ;) but on a higher level not all information might be already available so evaluation will just fail until a variable is defined in a lower level... 

e.g.

root: `globals { name = "name-${global.env}" }`
lower-level: `globals { env = "staging" }` 

so in stack it will be `name = "name-staging"`

only this way you do not repeat yourself..... if a variable is not evaluated in the stack and missing, an error should be thrown (for now)...

@katcipis

mariux:
> root: `globals { name = "name-${global.env}" }` 
> lower-level: `globals { env = "staging" }`
> 
> so in stack it will be `name = "name-staging"`

and on higher levels for debugging it could be printed as (ugly) `name = (known in stack)`

PR #96, opened 2021-12-17
feat: add autocomplete support

i4ki:


README.md
```diff
@@ -2,16 +2,18 @@
 
 # Table of Contents
 
+- [Table of Contents](#table-of-contents)
```
katcipis: hmm was this added with mdtoc ? it seems that it should not add this =P (thinking on removing Table Of Contents as a header in general, doesn't help much)

katcipis (approved):
:rocket:

README.md
i4ki: no just reverted those changes... vscode f@#$% up the file when I saved it...

katcipis: ahh it seems to be some sort of automatic TOC generation on vscode too =P

PR #97, opened 2021-12-17
chore: add check for go mod tidyness 

katcipis:
I started the PR with an outdated main on purpose to check if the CI/automation would catch the tidyness problem. Seems to be working:

https://github.com/mineiros-io/terramate/runs/4562025397?check_suite_focus=true#step:4:45


i4ki (approved):
lgtm

PR #102, opened 2021-12-20
fix: dont fail stack list if terramate block is absent

katcipis:
This causes some issues as discussed here: https://mineirosio.slack.com/archives/C02GFAAAKPT/p1640026231072900

Also is breaking global config testing since we can have a terramate HCL file with just globals inside (no terramate block) and that should work.


i4ki (approved):
lgtm

PR #104, opened 2021-12-21
feat: project root

i4ki:
This change creates the concept of "terramate project dir" or "terramate project root". All other features works relative to root, using root configurations, and so on, but the cli interface will continue to work from subdirectories of root.

If a git directory is not found, terramate works but with change detection disabled (an args error is raised if -c/--changed is provided).


mariux (commented):
cool... how is root detected if not in git?

what is precedence?

i4ki:
@mariux 

The precedence is:

1. `git rev-parse --git-dir
2. look for a "root configuration" in current directory and any parent directory until /.

A root configuration is a config with a terramate.config block.

mariux:
> 

if .git is found.... do i still require `terramate.config` block? or is this optional then? could be optional in this case..

i4ki:
@mariux yes, it's optional. ;)

i4ki:
It broke OSX... not sure why yet, investigating now.

i4ki:
fixed. The test function `t.TempDir()` returns symlinked paths (`/var/folders/...` instead of `/private/folders/...`) on OSX and it conflicted with the path returned from git tool for the project's root.

In addition to fix that case, I'm also canonicalizing the paths from `--chdir` in case it also has symlinks.

cmd/terramate/cli/cli.go
```diff
@@ -108,20 +104,27 @@ type cliSpec struct {
...
 		return err
 	}
 	return c.run()
 }
 
+type project struct {
+	root    string
+	wd      string
+	isRepo  bool
+	cfg     hcl.Config // root config
```
katcipis, line 124:
```suggestion
	rootcfg     hcl.Config
```
Would make it more explicitely when referencing it + no need for the docs then.

cmd/terramate/cli/cli.go
```diff
@@ -204,15 +185,50 @@ func newCLI(
...
+	}
+
+	foundRoot, prj, err := lookupProject(wd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to lookup project root from %q: %w", wd, err)
+	}
+
+	if !foundRoot {
+		return nil, fmt.Errorf("project root not found")
+	}
```
katcipis, line 213: Maybe because of how Go map access works , but it feels more intuitive to have:

```
prj, found, err := lookupProject(wd)
```

Instead of the first return value being a boolean. Not a major (or even minor =P) issue of course, just sharing =P.

cmd/terramate/cli/cli.go
```diff
@@ -294,40 +305,27 @@ func (c *cli) initStack(dirs []string) error {
...
+	entries, err := c.listStacks(mgr, c.parsedArgs.Changed)
 	if err != nil {
 		return err
 	}
 
-	trimPart := c.wd + string(os.PathSeparator)
 	for _, entry := range entries {
 		stack := entry.Stack
-		stackdir := strings.TrimPrefix(stack.Dir, trimPart)
+		stackdir, ok := c.showdir(stack.Dir)
```
katcipis, line 325: Just to give some feedback from the perspective of someone new to the code, I haven't got to the definition of showdir yet, but in this context it is kinda hard to intuitively understand what is going on, what would "showdir" do/check I wonder :thinking:

cmd/terramate/cli/cli.go
```diff
@@ -466,26 +467,25 @@ func (c *cli) printMetadata() error {
...
 		c.log("Running on all stacks:")
 	}
 
-	stacks := make([]stack.S, len(entries))
-	for i, e := range entries {
-		stacks[i] = e.Stack
+	relwd := prj.RelPath(c.root(), c.wd())
+	stacks := make([]stack.S, 0, len(entries))
+	for _, e := range entries {
+		if strings.HasPrefix(e.Stack.Dir, relwd) {
```
katcipis, line 486: This kind of filter is done at least on printRunOrder too, I think, I also think it is some kind of filtering going on, maybe there is some way to make it more explicit.

cmd/terramate/cli/cli.go
```diff
@@ -612,6 +614,10 @@ func (c *cli) checkLocalDefaultIsUpdated(g *git.Git) error {
 	return nil
 }
 
+func (c *cli) showdir(dir string) (string, bool) {
+	return prj.ShowDir(c.root(), c.wd(), dir)
+}
```
katcipis, line 619: hmm :thinking: ...starting to think that it is some sort of canonical dir representation, but not sure yet =P

cmd/terramate/cli/cli.go
```diff
@@ -628,3 +634,116 @@ func newGit(basedir string, inheritEnv bool, checkrepo bool) (*git.Git, error) {
...
+			prj.cfg = cfg
+			prj.root = root
+
+			return true, prj, nil
+		}
+	}
+
+	dir := wd
+
+	for {
```
katcipis, line 673:
```suggestion
	for dir != "/" {
```

Debatable but I usually prefer the stopping criteria on the loop declaration (if possible). In this case it seems possible, but maybe Im wrong, disregard at will

cmd/terramate/cli/cli_test.go
```diff
@@ -160,29 +154,27 @@ func TestListAndRunChangedStack(t *testing.T) {
...
-		"--basedir",
-		s.BaseDir(),
 		"--changed",
 		cat,
 		mainTfFileName,
 	), runResult{Stdout: wantRun})
 }
 
 func TestListAndRunChangedStackInAbsolutePath(t *testing.T) {
+	t.SkipNow()
```
katcipis, line 177: Is the test supposed to be skipped or is this a leftover from some other change ?

config/config.go
```diff
@@ -45,3 +47,25 @@ func Exists(path string) bool {
 
 	return info.Mode().IsRegular()
 }
+
+func TryLoadRootConfig(dir string) (found bool, cfg hcl.Config, err error) {
```
katcipis, line 51:
```suggestion
func TryLoadRootConfig(dir string) (cfg hcl.Config, found bool, err error) {
```

Usually on "found" logic the boolean comes after the actual value, like Go maps, but it is not like a rule or pattern, just sharing that it feels odd but probably because I'm not used to it =P. Kinda like actual value first + other complementary information like errors/boolens comes after...but Im probably just stuck with a particular mindset.

config/config.go
```diff
@@ -45,3 +47,25 @@ func Exists(path string) bool {
...
+	path := filepath.Join(dir, Filename)
+	_, err = os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
```
katcipis, line 55: Seems this makes me wonder if it would not be better to model this in the same fashion, like you load the root config and one of the possible errs is "not found", then callers can handle this differently if they want, instead of using the boolean.

generate.go
```diff
@@ -40,33 +41,33 @@ const (
...
+func Generate(root string) error {
+	if !filepath.IsAbs(root) {
+		return fmt.Errorf("project's root %q must be an absolute path", root)
 	}
 
-	info, err := os.Lstat(basedir)
+	info, err := os.Lstat(root)
 	if err != nil {
-		return fmt.Errorf("checking basedir %q: %v", basedir, err)
+		return fmt.Errorf("checking basedir %q: %v", root, err)
```
katcipis, line 58:
```suggestion
		return fmt.Errorf("checking project root %q: %v", root, err)
```

hcl/hcl.go
```diff
@@ -37,11 +37,26 @@ type Config struct {
...
+type GitConfig struct {
+	BaseRef              string // BaseRef is the general base git ref.
+	DefaultBranchBaseRef string // DefaultBranchBaseRef is the baseRef when in default branch.
+	Branch               string // Branch is the default branch.
+	Remote               string // Remote is the default remote.
+}
+
+type RootConfig struct {
+	Git GitConfig
+}
```
katcipis, line 49: we have specs for git config goodies: https://github.com/mineiros-io/terramate/blob/main/docs/project-config.md
we can update the specs if something changed here (I think some names did), or fix the code (I'm quite good with both).

hcl/hcl.go
```diff
@@ -124,48 +139,37 @@ func ParseModules(path string) ([]Module, error) {
...
-	body, ok := f.Body.(*hclsyntax.Body)
-	if !ok {
-		return nil, fmt.Errorf("expected to parse body, got[%v] type[%[1]T]", f.Body)
-	}
-	return body, nil
-}
-
 // Parse parses a terramate source.
-func Parse(fname string, data []byte) (*Config, error) {
+func Parse(fname string, data []byte) (Config, error) {
```
katcipis, line 143: Noice :-)

cmd/terramate/cli/cli.go
i4ki: yeah me too but in this case it needs to check at `/` as well and if the config is still not there then we break the loop. It'd be a perfect case for a `do { ... } while(dir != "/")` haha

hcl/hcl_test.go
```diff
@@ -281,6 +342,225 @@ terramate {
...
+			want: want{
+				config: hcl.Config{
+					Terramate: &hcl.Terramate{
+						RootConfig: &hcl.RootConfig{},
+					},
+				},
+			},
+		},
+		{
+			name: "multiple config blocks - fails",
```
katcipis, line 430: We are missing the multiple git blocks fails test, I think.

hcl/hcl_test.go
```diff
@@ -281,6 +342,225 @@ terramate {
...
+				config: hcl.Config{
+					Terramate: &hcl.Terramate{},
+					Stack:     &hcl.Stack{},
+				},
+			},
+		},
+		{
+			name: "multiple stack blocks",
+			input: `stack{}
+stack {}`,
```
katcipis, line 517: On this input there is not terramate block, is this OK/expected ? (wondering if terramate is always required, or not, when, etc).

hcl/hcl_test.go
```diff
@@ -281,6 +342,225 @@ terramate {
...
+		},
+		{
+			name: "name has interpolation - fails",
+			input: `
+terramate {
+	required_version = ""
+}
+stack {
+	name = "${test}"
+}`,
```
katcipis, line 559: maybe we could extract the terramate default blocks in a way that avoids it being repeated on all tests, so tests only shows stack blocks (terramate block seems to always be the same, so kinda unrelated to what is being tested).

cmd/terramate/cli/cli.go
katcipis: Ah got it, I missed this :+1:

hcl/hcl_test.go
i4ki: yeah good catch, in this case it's failing for multiple stacks, but if there was only 1 declaration then it'd fail with ErrNoTerramateBlock. Will fix.

project/project.go
```diff
@@ -0,0 +1,52 @@
...
+func RelPath(root, dir string) string {
+	d := strings.TrimPrefix(dir, root)
+	if d == "" {
+		d = "/"
+	}
+	return d
+}
+
+// AbsPath returns the absolute dir for project's root.
+func AbsPath(root, dir string) string {
```
katcipis, line 32:
```suggestion
// AbsPath takes the root project dir and a dir path that is relative to the root project dir
// and returns an absolute path (relative to the host root).
func AbsPath(root, dir string) string {
```

Maybe, not sure if it is better =P. Since we have some relative paths that look like root (starting with /) seems interesting to be explicit about being root on the actual host.

run.go
```diff
@@ -16,20 +16,29 @@ package terramate
...
 	"github.com/madlambda/spells/errutil"
+	"github.com/mineiros-io/terramate/project"
 	"github.com/mineiros-io/terramate/stack"
 )
 
 const ErrRunCycleDetected errutil.Error = "cycle detected in run order"
 
-func Run(stacks []stack.S, cmdSpec *exec.Cmd) error {
+func Run(root string, stacks []stack.S, cmdSpec *exec.Cmd) error {
+	wd, err := os.Getwd()
```
katcipis, line 31: Shouldn't Run use the wd defined [here](https://github.com/mineiros-io/terramate/pull/104/files#diff-3a3a4d5e49230d35db58ed019f072d3b3fb7d818e9c255c2208d68bc4bc08578R188) ?

stack/loader.go
```diff
@@ -60,27 +68,20 @@ func (l Loader) Load(dir string) (S, error) {
...
-	s.changed = true
-	return s, nil
+	l.set(stackpath, cfg.Stack)
+	return l.stacks[stackpath], nil
 }
 
 // TryLoad tries to load a stack from directory. It returns found as true
 // only in the case that path contains a stack and it was correctly parsed.
 // It caches the stack for later use.
 func (l Loader) TryLoad(dir string) (stack S, found bool, err error) {
```
katcipis, line 78: Hmm on this Try* method we went with the more classical (return value, boolean, error), instead of (boolean, return value, error). I prefer it :-) (not that it matter much, again feel free to ignore)

stack/stack.go
```diff
@@ -44,25 +44,25 @@ func (s S) String() string {
...
-	l := NewLoader()
+func LookupParent(root, dir string) (S, bool, error) {
+	l := NewLoader(root)
 	return l.lookupParentStack(dir)
 }
 
 // Load a single stack from dir.
-func Load(dir string) (S, error) {
-	l := NewLoader()
+func Load(root, dir string) (S, error) {
```
katcipis, line 58: Im thinking that for path manipulation stuff maybe the stack.S could keep the root inside and be able to provide all sorts of different stack paths required (I think this is spread in multiple places).

katcipis: Like maybe the showdir thing, if it is more to showing stack paths in a nice/cute way, could be a method on the stack itself, like a nice string representation of the type, etc (just wondering/brainstorming)

test/sandbox/sandbox.go
```diff
@@ -77,7 +77,7 @@ type FileEntry struct {
 func New(t *testing.T) S {
 	t.Helper()
 
-	basedir := t.TempDir()
+	basedir := test.CanonPath(t, t.TempDir())
```
katcipis, line 80: maybe we could rename basedir to rootdir on the sandbox, would be more aligned with the new design.

katcipis (approved):
LGTM + bunch of comments but nothing blocking/essential

mariux (changes requested):
there is an endless loop... missing debug logs to help more..

i assume if `dir` in `-C dir` does not contain a `terramate.tm.hcl` it stats it and fails in an endless loop

mariux:
> there is an endless loop... missing debug logs to help more..
> 
> i assume if `dir` in `-C dir` does not contain a `terramate.tm.hcl` it stats it and fails in an endless loop

if i touch the file (create it), it fails with `2021/12/22 20:30:53 failed to lookup project root from "stacks": no "terramate" block found`

mariux:
@i4ki latest commit still fails same way ;)

cmd/terramate/cli/cli.go
```diff
@@ -40,54 +44,46 @@ const (
...
 
 const (
-	defaultRemote       = "origin"
-	defaultBranch       = "main"
-	defaultMainBaseRef  = "HEAD^1"
-	defaultOtherBaseRef = defaultRemote + "/" + defaultBranch
+	defaultRemote        = "origin"
+	defaultBranch        = "main"
+	defaultBaseRef       = defaultRemote + "/" + defaultBranch
+	defaultBranchBaseRef = "HEAD^1"
```
mariux, line 50:
```suggestion
	defaultBranchBaseRef = "HEAD^"
```

i4ki:
thanks @mariux 

fixed it. All the tests use `-C` by default but they were all absolute paths.
When looking for the root config, I use `filepath.Dir()` to get the parent directory, but if `wd` was relative it returns "." when reaching the last part of path.
Now I just make it absolute from the beginning.
Still missing a testcase for this. Will add in a second.

katcipis (dismissed):

cmd/terramate/cli/cli_test.go
i4ki: not sure what to do with this test because it tests the case where basedir and cwd mismatch, remember? Like `cd /tmp; terramate list /home/iac` in this case it reported absolute paths but now `--chdir <path>` is the new way and we chdir() into the directory first.

stack/stack.go
i4ki: I liked this idea, we can improve in other PR.

cmd/terramate/cli/cli_test.go
katcipis: hmm I see, maybe use t.Skip() with a message making this more clear + we assess this soon

katcipis (approved):
lgtm :rocket:

mariux (commented):
lgtm! 

having questions on the flow:
- how do you review such big PRs? since we use squash merge strategy... 
- how do you revert a specific part if something in refactoring breaks?
- how do you track down introduction of a failing behavior? (git bisect will not help, i assume?)

afaics this PR includes at least the following topics:
- refactor: refactor some internals
- refactor: make --changed global option
- feat: add --chdir
- fix: default change base HEAD^
- feat: add terramate.config
- refactor: add project structure
- etc etc etc

i4ki:
I will reply inline each question but first an overall explanation:

This "feature" was not in the initial talks/specs of terramate, so the tool used whatever user provided as basedir or the current directory, and it just walked directories detecting stuff. The concept of root had to be injected into most data structures and functions to avoid a huge (much bigger than this) refactor. The list of features gets bigger each day and I was told features were the most important at this point and refactors can come later, but actually that's not how it works, the original code design doesn't accommodate the new features, so we refactor just what's needed now but it still a lot of code, unfortunately, and no, I'm not happy with the code =(

Now the questions:

> how do you review such big PRs? since we use squash merge strategy...

I'm very _pro_ of small PRs but this change is big and needed, no way around it, no way to do it incrementally because it's fundamental.

Individual commits don't help at all with big refactorings, as this is the case, because lots of commits would not build or tests not pass, etc. Most renowned projects I know do not rely on commits but the full patch that's being contributed. I never used commit messages to aid in review and I don't believe it makes sense but maybe this is just personal taste.

> how do you revert a specific part if something in refactoring breaks?

You can't revert part of a refactor, but only if there are multiple refactors into one, but this is not the case.

This patch probably introduces issues, of course, but if we want solid software at this point, I would say to re-design the code for the current set of features agreed.

> how do you track down introduction of a failing behavior? (git bisect will not help, i assume?)

Git bisects helps at identifying that this is the offended PR.

> afaics this PR includes at least the following topics:
> 
> refactor: refactor some internals
> refactor: make --changed global option
> feat: add --chdir

As I explained in slack when you suggested the `--chdir` option, the old code relies on the basedir concept and it could be passed in the command-line in different ways and very flexible, for example, you can even provide a path outside of your current working tree (cwd), ie outside of the root of you project, but this would not be possible anymore after refactor (I even had to skip one existent test for that), so instead of having a broken cli I had to implement `--chdir` in this PR to support similar use cases (like avoiding cd into dirs).

> fix: default change base HEAD^

This is not even a fix, by the documentation of [git revisions](https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrevgtltngtemegemHEADv1510em) , `HEAD^` is the same as `HEAD^1`. I prefer `HEAD^1` because it's explicit about *first parent*, that's why I used this when I introduced the change base feature. If this is indeed a bugfix, please clarify.

> feat: add terramate.config

This could have been avoided but if so then the only way to detect root would be by having a git repository. As this PR introduces the concept of root I thought it would not harm to have this already. I can easily drop this before merging if you would like.

> refactor: add project structure
> etc etc etc

Don't know what you mean here, but if it's the struct... it's just an implementation detail of the project root idea.

katcipis (dismissed):

katcipis (approved):

PR #108, opened 2021-12-23
feat: add globals on backend config generation

katcipis:
Integrates globals loading into backend config code generation. Didn't test exaustively, so maybe there is some sneaky corner case missing :-).

Example script:

```
#!/bin/bash

set -o errexit
set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p envs/prod/stack
terramate init envs/prod/stack

mkdir -p envs/staging/stack
terramate init envs/staging/stack

echo "listing stacks:"
echo

terramate list

echo

cat > envs/terramate.tm.hcl <<- EOM
terramate {
  backend "gcs" {
    bucket = "base-name-\${global.env}"
    prefix = terramate.path
  }
}
EOM

globals {
  env = "staging"
}
EOM

globals {
  env = "prod"
}
EOM


echo "files before code generation:"

tree

echo

echo "=== generating code ==="
echo

terramate generate

echo "files after code generation:"

tree

echo
echo "generated code for staging:"

cat envs/staging/stack/_gen_terramate.tm.tf

echo
echo "generated code for prod:"

cat envs/prod/stack/_gen_terramate.tm.tf
```


globals.go
```diff
@@ -94,16 +104,13 @@ func (r *rawGlobals) eval(meta StackMetadata) (*Globals, error) {
 		return nil, err
 	}
 
+	globals := newGlobals()
 	// error messages improve if globals is empty instead of undefined
```
i4ki, line 108: maybe now this comment can be removed?

globals.go
```diff
@@ -94,16 +104,13 @@ func (r *rawGlobals) eval(meta StackMetadata) (*Globals, error) {
...
-	empty, err := hclMapToCty(nil)
-	if err != nil {
-		return nil, fmt.Errorf("initializing empty globals: %v", err)
+	if err := globals.addToEvalCtx(evalctx); err != nil {
```
i4ki, line 109: now the global is not empty before the evaluations but contains meta vars in it. Doesn't this change the behavior if no variable gets evaluated?

cmd/terramate/cli/generate_test.go
```diff
@@ -622,6 +622,133 @@ stack {
...
+				},
+			},
+			want: want{
+				res: runResult{
+					Error:        terramate.ErrGenerateLoadingGlobals,
+					IgnoreStdout: true,
+				},
+			},
+		},
 	}
```
i4ki, line 752: I know it's already covered in globals_test.go but maybe having the overriding/shadowing test here as well would be good, as this is an important behavior for the generate+globals.

i4ki (approved):
lgtm but left some not important comments.

globals.go
katcipis: actually no, it is a little odd to add an empty globals to the ctx, and that is done just to improve error messages (doesn't change behavior). I mean...we could remove the comment anyway...but since the "why" is kinda funny on this case I went with the comment.

globals.go
katcipis: globals is empty, the code is basically:

```
globals := newGlobals()
if err := globals.addToEvalCtx(evalctx); err != nil {
}
```

This means:

* Create new/empty globals
* Add empty globals to evalctx

**newGlobals** is just this:

```go
func newGlobals() *Globals {
	return &Globals{
		attributes: map[string]cty.Value{},
	}
}
```

cmd/terramate/cli/generate_test.go
katcipis: makes sense, also wanted to improve the tests overall x_x (bunch of duplication on HCL raw strings). WDYT on doing in a future PR ?

i4ki: Yep, makes sense

i4ki (approved):
lgtm

PR #112, opened 2021-12-27
feat: make terramate blocks optional on stacks

katcipis:
Now terramate{} is no longer required on stacks for them to be detected as stacks, only "stack{}".


i4ki (approved):
lgtm

PR #115, opened 2021-12-27
spec: add locals terraform generation

katcipis:
Basic initial spec to how someone can leverage Terramate metadata + global definitions on Terraform code in a stack. The idea is to debate/evolve on top of this, so not final or complete, just the first idea that came from me and @i4ki talking about this.

### DIFF
--- a/docs/locals-generation.md
+++ b/docs/locals-generation.md
@@ -117,6 +117,44 @@ local.data
 local.more_data
 ```
 
+Now lets extend the exported locals only for **stack-1** by adding
+this to **stacks/stack-1/terramate.tm.hcl**:
+
+```hcl
+export_as_locals {
+  stack_1_only = globals.yet_more_data
+}
+```
+
+Now the generated locals for **stacks/stack-1** will be
+(while **stacks/stack-2** remains unchanged): 
+
+```hcl
+locals {
+  data         = "data"
+  more_data    = "more data"
+  stack_1_only = "YMD"
+}
+```
+
+Now lets override the exported locals for **stack-1** by changing the
+export definition on **stacks/stack-1/terramate.tm.hcl** to:
+
+```hcl
+export_as_locals {
+  more_data    = globals.yet_more_data
+}
+```
+
+Now the generated locals for **stacks/stack-1** will be: 
+
+```hcl
+locals {
+  data         = "data"
+  more_data    = "YMD"
+}
+```
+
 You can also export Terramate [metadata](metadata.md) as locals:
 
 ```hcl
--- a/docs/locals-generation.md
+++ b/docs/locals-generation.md
@@ -83,8 +83,8 @@ that export some locals at **stacks/terramate.tm.hcl**:
 
 ```hcl
 export_as_locals {
-  data      = globals.data
-  more_data = globals.more_data
+  data      = global.data
+  more_data = global.more_data
 }
 ```
 
--- a/docs/locals-generation.md
+++ b/docs/locals-generation.md
@@ -23,8 +23,8 @@ to create them as locals:
 
 ```hcl
 export_as_locals {
-  data      = globals.data
-  more_data = globals.more_data
+  data      = global.data
+  more_data = global.more_data
 }
 ```
 
--- a/docs/locals-generation.md
+++ b/docs/locals-generation.md
@@ -38,7 +38,7 @@ export_as_locals {
 }
 
 export_as_locals {
-  more_data = globals.more_data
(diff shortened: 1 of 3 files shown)


mariux (changes requested):
locals should not defined in context of a stack.. as this would mean code duplication in each and every stack.. it should follow same inheritance as globals to be able to leverage full power..

i4ki (approved):
LGTM :rocket:

katcipis:
> locals should not defined in context of a stack.. as this would mean code duplication in each and every stack.. it should follow same inheritance as globals to be able to leverage full power..

hmm I did miss that from our talks :thinking: 

@mariux do you have any suggestions on how this would look like then ? I imagine it would be something similar to the globals block that can be anywhere then but not very sure about naming/structure, I suppose you already have something in mind.

mariux:
the idea is to keep it DRY as much as possible.... with the current spec i see i copy 

```hcl
terramate {
  stack {
    locals {
      project     = global.project
      environment = global.environment
    }
  }
}
```

into each and every stack...

so not sure but if we would just have a 

```
locals {
  project     = global.project
  environment = global.environment
}
```

block on a higher level it would help not repeating it

same behavior as globals?

mariux:
maybe it would even make sense to omit exporting in the stack..

```hcl
locals {
  project     = global.project
  environment = omit
}
```

this would not export `environment` anymore... so we can exclude locals from export inside a stack..

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+```hcl
+globals {
+  data          = "data"
+  more_data     = "more data"
+  yet_more_data = "YMD"
+}
+```
+
+If we define a [configuration file](config.md) for all stacks 
+export some locals at **stacks/terramate.tm.hcl**:
```
i4ki, line 82:
```suggestion
that export some locals at **stacks/terramate.tm.hcl**:
```

i4ki (commented):
lgtm

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+Exporting using multiple blocks on the same [configuration file](config.md)
+is also allowed, as far as no local is redefined, the previous example could
+also be written like this:
+
+```hcl
+export_as_locals {
+  data      = globals.data
+}
```
mariux, line 38:
```suggestion
export_as_locals {
  data = globals.data
}
```

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+
+export_as_locals {
+  more_data = globals.more_data
```
mariux, line 41:
```suggestion
  more_data = global.more_data
```

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+If we define a [configuration file](config.md) for all stacks 
+that export some locals at **stacks/terramate.tm.hcl**:
+
+```hcl
+export_as_locals {
+  data      = globals.data
+  more_data = globals.more_data
```
mariux, line 87:
```suggestion
  data      = global.data
  more_data = global.more_data
```

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+local.data
+local.more_data
+```
+
+Now lets extend the exported locals only for **stack-1** by adding
+this to **stacks/stack-1/terramate.tm.hcl**:
+
+```hcl
+export_as_locals {
+  stack_1_only = globals.yet_more_data
```
mariux, line 124:
```suggestion
  stack_1_only = global.yet_more_data
```

docs/locals-generation.md
```diff
@@ -0,0 +1,188 @@
...
+}
+```
+
+To use then directly on Terraform you can instruct code generation
+to create them as locals:
+
+```hcl
+export_as_locals {
+  data      = globals.data
+  more_data = globals.more_data
```
mariux, line 27:
```suggestion
  data      = global.data
  more_data = global.more_data
```

mariux (changes requested):
typo in `global.` ;)

docs/locals-generation.md
katcipis: ow man I do this shit all the time x_x (switching global to globals :laughing: ). thanks for catching it

katcipis:
> typo in `global.` ;)

duh

mariux (approved):
lgtm

mariux (approved):
lgtm

PR #116, opened 2021-12-28
refactor: extract code generation to its own package

katcipis:
Also extracted some eval logic that is used by globals/metadata.


i4ki (approved):
lgtm

PR #118, opened 2021-12-28
refactor: extract test.hclwrite

katcipis:
The idea is to have something that can help avoid all the HCL raw strings we have today. It would need some tests for the builder + support to nested blocks (and probably a lot of improvements, it is quite hacky right now), seems doable but here I focused only on extracting what is used today on the globals test. Not sure this is a good idea, so OK if we decide to just drop this.


i4ki (approved):
top! lgtm

PR #119, opened 2021-12-29
feat: terramate stacks globals

katcipis:
Implement the command:

```
terramate stacks globals
```

That will list all globals defined per stack. A quick example:

```sh
#!/bin/bash

set -o errexit
set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-1
mkdir -p stacks/stack-2

terramate init stacks/stack-1
terramate init stacks/stack-2

echo "terramate list"
echo

terramate list

echo

globals {
  testing = "test value"
  object = {
    name = "hello"
    data = "world"
  }
}
EOM

echo "terramate stacks globals"

terramate stacks globals

echo
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -83,6 +83,11 @@ type cliSpec struct {
 		} `cmd:"" help:"show the topological ordering of the stacks"`
 	} `cmd:"" help:"plan execution."`
 
+	Stacks struct {
+		Globals struct {
+		} `cmd:"" help:"list globals for all stacks."`
+	} `cmd:"" help:"stack related commands."`
+
 	Generate struct{} `cmd:"" help:"Generate terraform code for stacks."`
 	Metadata struct{} `cmd:"" help:"shows metadata available on the project"`
 
@@ -224,7 +229,7 @@ func newCLI(
 	}
 
 	if parsedArgs.Changed && !prj.isRepo {
-		return nil, fmt.Errorf("flag --changed provided but no git repository found.")
+		return nil, fmt.Errorf("flag --changed provided but no git repository found")
 	}
 
 	return &cli{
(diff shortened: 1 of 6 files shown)


i4ki (approved):
lgtm but now we have a hcl/printer.go and hcl/formatter.go, maybe we can unify those concepts in a single place.

katcipis:
> lgtm but now we have a hcl/printer.go and hcl/formatter.go, maybe we can unify those concepts in a single place.

I was going to add it as a printer at first, but for what I needed right now having to pass an io.Writer would be just extra work, but not a whole lot of work, by unification you mean this or something else ?

i4ki:
I mean that even if they have different APIs, they're mostly the same so they could be in the same place. Maybe having something like the below we can reuse in the `hcl.PrintConfig()` to print the attributes from a map.

```
func FormatAttributes(attrs map[string]cty.Value) string {
    buf := bytes.Buffer{}
    PrintAttributes(&buf, attrs)
    return string(buf.Bytes())
}
```

But don't need to change now. This can be improved later.

i4ki (approved):
lgtm

katcipis:
> I mean that even if they have different APIs, they're mostly the same so they could be in the same place. Maybe having something like the below we can reuse in the `hcl.PrintConfig()` to print the attributes from a map.
> 
> ```
> func FormatAttributes(attrs map[string]cty.Value) string {
>     buf := bytes.Buffer{}
>     PrintAttributes(&buf, attrs)
>     return string(buf.Bytes())
> }
> ```
> 
> But don't need to change now. This can be improved later.

It was my first idea, but since for now I would not use it it seemed like I was exaggerating or something =P (I'm not sure this is overall useful, at least the way it is, but if it is it makes perfect sense to do that).

PR #120, opened 2021-12-29
chore: remove duplicated stack listing

katcipis:
On generation we are listing stacks twice unnecessarily (we can just use the metadata).


i4ki (approved):
lgtm

PR #121, opened 2021-12-30
feat: change detection on stack subdirectories

i4ki:


test/sandbox/sandbox.go
```diff
@@ -366,6 +372,12 @@ func (se StackEntry) RelPath() string {
 	return se.DirEntry.relpath
 }
 
+// CreateDir creates a directory inside the stack directory. The relpath must be
+// relative to the stack directory.
+func (se StackEntry) CreateDir(relpath string) DirEntry {
+	return newDirEntry(se.t, se.abspath, relpath)
+}
+
```
katcipis, line 380: Since StackEntry is composed by DirEntry:https://github.com/mineiros-io/terramate/blob/1772847e0eacb20b370ff20fb5db58a2afe167a3/test/sandbox/sandbox.go#L60

Do we need the method defined again on it ? I think syntatic sugar for composition would just call the DirEntry method.

katcipis (dismissed):
LGTM + possible duplicated method implementation on sandbox that can be avoided

test/sandbox/sandbox.go
i4ki: I forgot StackEntry just embedded DirEntry. =)

katcipis (approved):

PR #122, opened 2021-12-30
move init command inside the stacks command

i4ki:


soerenmartius (approved):
LGTM from what I can see and taking into consideration that my Go knowledge isn't that great! Thanks for the contribution @i4ki ! Appreciated!

katcipis (approved):

PR #123, opened 2021-12-30
feat: add load of export_as_locals

katcipis:
Includes the core logic but no code generation, which will be done in a different PR. It is very similar to globals, but evaluation is simpler, so there is no attempt to extract duplication here yet, maybe in a future PR.

The spec: https://github.com/mineiros-io/terramate/blob/main/docs/locals-generation.md

### DIFF
--- a/export_as_locals.go
+++ b/export_as_locals.go
@@ -27,6 +27,8 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+const ErrExportAsLocalsRedefined errutil.Error = "export_as_locals attribute redefined"
+
 // ExportAsLocals represents a export_as_locals block.
 type ExportAsLocals struct {
 	attributes map[string]cty.Value
@@ -83,10 +85,14 @@ func loadStackExportAsLocals(rootdir string, cfgdir string) (unEvalExportAsLocal
 
 	for _, block := range blocks {
 		for name, attr := range block.Body.Attributes {
-			// TODO(katcipis): test behavior
-			//if globals.has(name) {
-			//return nil, fmt.Errorf("%w: global %q already defined in configuration %q", ErrGlobalRedefined, name, cfgpath)
-			//}
+			if exportLocals.has(name) {
+				return unEvalExportAsLocals{}, fmt.Errorf(
+					"%w: export_as_locals %q already defined in configuration %q",
+					ErrExportAsLocalsRedefined,
+					name,
+					cfgpath,
+				)
+			}
 			exportLocals.expressions[name] = attr.Expr
 		}
 	}
--- a/export_as_locals.go
+++ b/export_as_locals.go
@@ -43,5 +43,9 @@ func LoadStackExportAsLocals(rootdir string, sm StackMetadata, g *Globals) (Expo
 }
 
 func (e ExportAsLocals) Attributes() map[string]cty.Value {
-	return nil
+	attrcopy := map[string]cty.Value{}
+	for k, v := range e.attributes {
+		attrcopy[k] = v
+	}
+	return attrcopy
 }
--- a/export_as_locals.go
+++ b/export_as_locals.go
@@ -87,7 +87,7 @@ func loadStackExportAsLocals(rootdir string, cfgdir string) (unEvalExportAsLocal
 			//if globals.has(name) {
 			//return nil, fmt.Errorf("%w: global %q already defined in configuration %q", ErrGlobalRedefined, name, cfgpath)
 			//}
-			exportLocals.add(name, attr.Expr)
+			exportLocals.expressions[name] = attr.Expr
 		}
 	}
 
@@ -112,13 +112,9 @@ type unEvalExportAsLocals struct {
 func (r unEvalExportAsLocals) merge(other unEvalExportAsLocals) {
 	for k, v := range other.expressions {
 		if !r.has(k) {
-			r.add(k, v)
+			r.expressions[k] = v
 		}
 	}
-}
-
-func (r unEvalExportAsLocals) add(name string, expr hclsyntax.Expression) {
-	r.expressions[name] = expr
 }
 
 func (r unEvalExportAsLocals) has(name string) bool {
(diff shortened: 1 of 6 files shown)


soerenmartius (approved):
That's a decent amount of work! Thanks man, very good progress! I thank you for the input but will leave the review to someone who's a bit more familiar with Go :)

katcipis:
> That's a decent amount of work! Thanks man, very good progress! I thank you for the input but will leave the review to someone who's a bit more familiar with Go :)

thanks for the feedback/review :100:

export_as_locals.go
```diff
@@ -0,0 +1,180 @@
...
+// it reaches rootdir, loading export_as_locals and merging them appropriately.
+//
+// More specific definitions (closer or at the stack) have precedence over
+// less specific ones (closer or at the root dir).
+//
+// Metadata and globals for the stack are used on the evaluation of the
+// export_as_locals blocks.
+//
+// The rootdir MUST be an absolute path.
+func LoadStackExportAsLocals(rootdir string, sm StackMetadata, g *Globals) (ExportAsLocals, error) {
```
i4ki, line 49: Why are they called `LoadStack*` if they load from any kind of configuration?

i4ki: By having `ExportAsLocals` in the function names and types seems like the logic is doing this, when in fact this is just the name of the block...

I would call this just `LoadStackLocals()`, as it loads the locals values that are going to be generated later.

export_as_locals.go
```diff
@@ -0,0 +1,180 @@
...
+	if !filepath.IsAbs(rootdir) {
+		return ExportAsLocals{}, fmt.Errorf("%q must be an absolute path", rootdir)
+	}
+
+	unEvalExport, err := loadStackExportAsLocals(rootdir, sm.Path)
```
i4ki, line 54: `unEvalExport` is funny :smile:

First, you have symbols or variables and after evaluation you get values. So maybe:
```
localVars, err := loadStackLocals(rootdir, sm.Path)
localValues, err := localVars.eval(sm, g)
```
as globals and metadata are kinds of variables.

export_as_locals.go
```diff
@@ -0,0 +1,180 @@
...
+			return newUnEvalExportAsLocals(), nil
+		}
+		return loadStackExportAsLocals(rootdir, parentcfg)
+	}
+
+	if err != nil {
+		return unEvalExportAsLocals{}, err
+	}
+
+	exportLocals := newUnEvalExportAsLocals()
```
i4ki, line 85: `newUnEvalExportAsLocals()` again newUnEval is strange and AsLocal looks like converting something. Of course the context of the file is `export_as_locals` but it doesn't mean we need to embed this everywhere.

i4ki (approved):
LGTM but left comments regarding the naming of choice.

The block is called `export_as_locals` to be clear the symbols will be generated as terraform locals, but it doesn't mean types and functions must have `exportAsLocals` as this is just the user-facing block name.

We're adding the ability to generate TF locals from terramate variables (globals and metadata at this moment), so from very high level, the generate has the task of "load locals variable mapping" and generate TF locals. Having this in mind, I would call the file `locals_decl.go` or `locals_mapping.go` and the functions to load the locals definitions could be `LoadLocalsMap()` or `LoadLocalsDecl()` or something like that.

Just sharing that the whole naming feels strange, it's like we are mixing syntax with the semantics.

katcipis:
> LGTM but left comments regarding the naming of choice.
> 
> The block is called `export_as_locals` to be clear the symbols will be generated as terraform locals, but it doesn't mean types and functions must have `exportAsLocals` as this is just the user-facing block name.

I hated the naming, was just not able to think on something better yet =P. +1 on improving.

> 
> We're adding the ability to generate TF locals from terramate variables (globals and metadata at this moment), so from very high level, the generate has the task of "load locals variable mapping" and generate TF locals. Having this in mind, I would call the file `locals_decl.go` or `locals_mapping.go` and the functions to load the locals definitions could be `LoadLocalsMap()` or `LoadLocalsDecl()` or something like that.

hmm maybe LoadStackLocals, Im torn because I also wanted to make it clear the whole concept that locals use exported data from Terramate (that is the whole point), just talking about "loading locals" doesn't make that very clear IMHO, and the concept exists independent of syntax, exported as locals are a thing, they are exported from Terramate and imported on Terraform code on the stack.

> 
> Just sharing that the whole naming feels strange, it's like we are mixing syntax with the semantics.

Hmm yeah maybe, I think I was too literal, but the semantic concept of exporting Terramate data as locals exist, it is not just because of the syntax, we are literally exporting data from Terramate configs into stack Terraform code. At this stage we are not generating the final code yet, but indeed just loading, but we are loading "exported locals".

Maybe the naming should be around "exported locals"... hmm thinking. I will merge this and address this on a PR ASAP, just to make sure I can keep building generate. Thanks for the review/feedback :heart:

export_as_locals.go
katcipis: > Why are they called `LoadStack*` if they load from any kind of configuration?

Loading exported locals definitions depends on metadata + globals, and both are defined only on the context of a stack, there is no way to load locals that don't belong to a stack, unless we were loading un-evaluated information. It is pretty much the same as globals/metadata, they can be used/defined anywhere on a project but evaluation is always on the context of a stack.

katcipis: > By having `ExportAsLocals` in the function names and types seems like the logic is doing this, when in fact this is just the name of the block...
> 
> I would call this just `LoadStackLocals()`, as it loads the locals values that are going to be generated later.

Maybe `LoadStackExportedLocals()` ?

export_as_locals.go
katcipis: man I hated this uneval naming thing..but at the time I was completely stuck on not being able to think on better names, thanks :100:

export_as_locals.go
katcipis: Man adding the "As" everywhere was the dumbest idea I ever had in my life..honestly..WTF. +1 on improving.

PR #125, opened 2021-12-31
fix: tests broken missing stacks on init

katcipis:
Missing fix on tests for "stacks init" instead of just "init"


mariux (approved):
lgtm

PR #126, opened 2021-12-31
refactor: naming improvements

katcipis:
According to discussions from the PR: https://github.com/mineiros-io/terramate/pull/123/files


exported_locals_test.go
```diff
@@ -46,6 +46,12 @@ func TestExportAsLocals(t *testing.T) {
 	export_as_locals := func(builders ...hclwrite.BlockBuilder) *hclwrite.Block {
 		return hclwrite.NewBuilder("export_as_locals", builders...)
 	}
+	// exported_locals is used only as a syntatic sugar to build testcase.want
+	// HCL blocks, which are also just an easy way to describe the wanted
+	// result since the actual result is a map of cty.Values (quite annoying to build manually)
+	exported_locals := func(builders ...hclwrite.BlockBuilder) *hclwrite.Block {
```
i4ki, line 52: underscored names? why not `exportedLocals()` ?

i4ki (approved):

exported_locals_test.go
katcipis: I just copied from the exported_as_locals, but on that case it makes more sense since the idea was to match the syntax on HCL :laughing: (overall not sure the idea is good anyway). Will fix.

i4ki (approved):
lgtm

PR #130, opened 2022-01-03
refactor: generate backend config on its own file

katcipis:
The idea is to create different files for different generated code, so backend config should be generated as _gen_backend_cfg.tm.tf and locals are saved on _gen_locals.tm.tf. There is no detection of duplicated backend configs yet, so generating code with a newer version of terramate, if you already have code generated from a previous version, will result on duplicated backends on terraform, one in the old file + one in the new one. 

Example:

```sh
#!/bin/bash

set -o errexit
set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p envs/prod/stack
terramate stacks init envs/prod/stack

mkdir -p envs/staging/stack
terramate stacks init envs/staging/stack

echo "listing stacks:"
echo

terramate stacks list

echo

cat > envs/terramate.tm.hcl <<- EOM
terramate {
  backend "gcs" {
    bucket = "base-name-\${global.env}"
    prefix = terramate.path
  }
}
EOM

globals {
  env = "staging"
}
EOM

globals {
  env = "prod"
}
EOM


echo "files before code generation:"

tree

echo

echo "=== generating code ==="
echo

terramate generate

echo "files after code generation:"

tree

echo
echo "generated code for staging:"

cat envs/staging/stack/_gen_backend_cfg.tm.tf

echo
echo "generated code for prod:"

cat envs/prod/stack/_gen_backend_cfg.tm.tf
```


mariux (approved):
lgtm

PR #132, opened 2022-01-04
chore: add basic pre-commit hook

katcipis:


Makefile
```diff
@@ -68,6 +68,21 @@ install:
 clean:
 	rm -rf bin/*
 
+## Runs pre-commit hook
+.PHONY: pre-commit
+pre-commit:
+	./hack/hooks/pre-commit
```
mariux, line 74: why hack?

hack/hooks/pre-commit
```diff
@@ -0,0 +1,19 @@
...
+echo
+echo "=== linting ==="
+make lint
+
+echo
+echo "=== testing ==="
+make test
+
+echo "=== building ==="
+make build
```
mariux, line 19: a pre-commit should not build..

Makefile
katcipis: you mean why use the dir hack overall to store the pre-commit script or using the one at hack to run the make target ? On the target I used the script at hack since it is the one that is always there independent if you did the local setup of the hook or not, but can change.

hack/hooks/pre-commit
katcipis: hmm the idea was to be sure that I'm not commiting something that doesn't build (it is possible to pass tests + linter but build is broken, but it is rare).

Makefile
i4ki: I think he meant: Why call this directory "hack"?

katcipis: I think I started using hack when I copied the idea from kubernetes/docker(now moby)/some other projects. But I dont have a strong opinion about this, could also be scripts or whatever else we prefer (don't love tha name TBH). I remember at the time, but a long time ago, like 2015, I found a bunch of project using hack. Did some digging now and it seems that it is divided between scripts and hack. No strong opinion on either of the names, starting to think that scripts is better.

* https://github.com/kubernetes/kubernetes/tree/master/hack
* https://github.com/moby/moby/tree/master/hack
* https://github.com/etcd-io/etcd/tree/main/hack
* https://github.com/docker/cli/tree/master/scripts
* https://github.com/tailscale/tailscale/tree/main/scripts
* https://github.com/cockroachdb/cockroach/tree/master/scripts

WDYT @mariux @i4ki ?

i4ki: The `hack` directory commonly contains helpers for the developers, quick e2e scripts, or alias to commands often used while developing, things not every developer uses, and so on.
If we call it `scripts` it kinda looks like users can rely on them, and I guess this is not the case (at least not atm).

About the origin of the name, I could be wrong but I think it evolved from the idea of a `HACKING.txt` file that lots of opensource projects had in the past, where the project authors explained how to "hack the code", useful stuff contributors would need while coding, scripts to find mem leaks, races, etc, scripts to debug, etc, but formatted in txt file instead of ready to use in a directory.

i4ki: examples of the `HACKING` file:

- nmap: https://github.com/nmap/nmap/blob/master/HACKING
- gcc: I'm sure they had but now renamed to several `contrib` files per module.
- gccgo: https://go.googlesource.com/gofrontend/+/master/HACKING
- boost lib: https://www.boost.org/doc/libs/1_39_0/tools/build/v2/hacking.txt
- etc

katcipis: > The `hack` directory commonly contains helpers for the developers, quick e2e scripts, or alias to commands often used while developing, things not every developer uses, and so on. If we call it `scripts` it kinda looks like users can rely on them, and I guess this is not the case (at least not atm).

hmm yeah that makes sense, although at least for cockroachdb it seems that they use scripts exactly to what would be hack:

https://github.com/cockroachdb/cockroach/tree/master/scripts

Like benchmark running, git blame, etc. Doesn't mean we have to do the same...it is just that I don't think there is much convergence on what each means. But I do like your approach and would be OK with keeping it hack.

> 
> About the origin of the name, I could be wrong but I think it evolved from the idea of a `HACKING.txt` file that lots of opensource projects had in the past, where the project authors explained how to "hack the code", useful stuff contributors would need while coding, scripts to find mem leaks, races, etc, scripts to debug, etc, but formatted in txt file instead of ready to use in a directory.

Hmm as not aware of this,thanks for sharing :heart:

katcipis: > examples of the `HACKING` file:
> 
> * nmap: https://github.com/nmap/nmap/blob/master/HACKING
> * gcc: I'm sure they had but now renamed to several `contrib` files per module.
> * gccgo: https://go.googlesource.com/gofrontend/+/master/HACKING
> * boost lib: https://www.boost.org/doc/libs/1_39_0/tools/build/v2/hacking.txt
> * etc

Me as soon as I saw the boost link: 

d

mariux:
@i4ki @katcipis for pre commits we normally use [pre-commit](https://pre-commit.com/) is this also possible for go projects?

katcipis:
> @i4ki @katcipis for pre commits we normally use [pre-commit](https://pre-commit.com/) is this also possible for go projects?

Probably, but given that the current pre commit is just about calling a few make targets not so sure what would be the advantage. Right now we need Go + Make, now we would need Go + Make + Python + Pip to have git hooks :thinking:

Makefile
i4ki: hahaha just sent the first results for a HACKING.txt google dork

PR #134, opened 2022-01-04
feat: terramate generate should not overwrite files that are not terramate managed

katcipis:
For now using a perfect match with the generated header as a way to detect if the file is a terramate generated file or not. If it is not a terramate generated file now we fail generation for the specific stack + file remains unchanged.


mariux (approved):
lgtm

PR #135, opened 2022-01-04
fix: globals references using try won't work

katcipis:


globals_test.go
```diff
@@ -299,6 +299,46 @@ func TestLoadGlobals(t *testing.T) {
...
 		},
+		{
+			name:   "stack with globals referencing globals using try",
+			layout: []string{"s:stack"},
+			globals: []globalsBlock{
+				{
+					path: "/stack",
+					add: globals(
+						str("field", "value"),
+						expr("field_try", `try(global.field, "error")`),
```
mariux, line 310: not sure we should support this case... hmm... we could ofc.... it's not supported by terraform and in current implementation it will always be true or error with unknown var...

globals_test.go
```diff
@@ -299,6 +299,46 @@ func TestLoadGlobals(t *testing.T) {
...
+		},
+		{
+			name:   "stack with globals referencing metadata using try",
+			layout: []string{"s:stack"},
+			globals: []globalsBlock{
+				{
+					path: "/stack",
+					add: globals(
+						expr("path_try", `try(terramate.path, "error")`),
+						expr("undefined_try", `try(terramate.undefined, "undefined")`),
```
mariux, line 331: that's not how try works in terraform ;)

mariux (changes requested):

globals_test.go
katcipis: Indeed, when I looked into the docs I saw this

"try evaluates all of its argument expressions in turn and returns the result of the first one that does not produce any errors."

Imagined that reference to undefined would just be an error.

But missed this:

"The try function will not catch errors relating to constructs that are probably invalid even before dynamic expression evaluation, such as a malformed reference or a reference to a top-level object that has not been declared"

Sorry for the noise

globals_test.go
katcipis: I'm stupid, lets focus on tf behavior

globals_test.go
mariux: all good.. we can actually implement it the way we want.... for now i think it's better to carp out if a variable is not defined at all on any level... wdyt?

PR #136, opened 2022-01-05
fix: evaluate only globals already known

mariux:
- detect unknown namespaces
- detect unknown variables in global namespace
- fix evaluation order
- detect global variables that could not be evaluated (cycles)
- set new context in each loop execution

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path/filepath"
 
+	hhcl "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/madlambda/spells/errutil"
 	"github.com/mineiros-io/terramate/config"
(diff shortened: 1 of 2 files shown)


katcipis:
Tested with:

```sh
#!/bin/bash

set -o errexit
set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p dir/stack

cat > dir/stack/terramate.tm.hcl <<- EOM
globals {
  team = {
    members = ["aaa"]
  }
  members = global.team.members
  members_try = try(global.team.members, [])
}

export_as_locals {
  members = global.members
  members_try = global.members_try
}

stack {}
EOM

terramate generate

echo
echo "generated code"
cat dir/stack/_gen_locals.tm.tf

echo
echo "listing globals"
terramate stacks globals
```

And it seems to be working now, thanks :100:

katcipis (approved):

kassianh (approved):

PR #137, opened 2022-01-05
test: add some tests on globals eval with try()

katcipis:
Adding some tests to cover the fix for try() usage, it was failing before but after I merged the fix it started to work, so it seems to cover the fix reasonably well.


mariux (approved):
lgtm

PR #138, opened 2022-01-05
fix: wrong error message about missing symlink

mariux:
- fix: wrong error message about missing symlink

includes #140

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -199,10 +199,7 @@ func newCLI(
 			Msgf("failed to parse cli args: %v", args)
 	}
 
-	logLevel := parsedArgs.LogLevel
-	logFmt := parsedArgs.LogFmt
-
-	configureLogging(logLevel, logFmt, stderr)
+	configureLogging(parsedArgs.LogLevel, parsedArgs.LogFmt, stderr)
 
 	wd, err := os.Getwd()
 	if err != nil {
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -204,51 +204,46 @@ func newCLI(
 
 	configureLogging(logLevel, logFmt, stderr)
 
-	logger.Trace().
-		Msg("Get working directory.")
-	wd := parsedArgs.Chdir
-	if wd == "" {
-		wd, err = os.Getwd()
+	wd, err := os.Getwd()
+	if err != nil {
+		log.Fatal().
+			Err(err).
+			Msg("Getwd() failed")
+	}
+
+	if parsedArgs.Chdir != "" {
+		logger.Debug().
+			Str("wd", wd).
+			Str("dir", parsedArgs.Chdir).
+			Msg("Changing working directory")
+		err = os.Chdir(parsedArgs.Chdir)
 		if err != nil {
 			logger.Fatal().
+				Str("wd", wd).
+				Str("dir", parsedArgs.Chdir).
 				Err(err).
-				Msg("failed to get working directory")
+				Msg("Changing working directory failed")
 		}
+
+		wd, err = os.Getwd()
+		if err != nil {
+			log.Fatal().
+				Err(err).
+				Msg("Getwd() failed")
+		}
+	}
+
+	wd, err = filepath.EvalSymlinks(wd)
+	if err != nil {
+		log.Fatal().
+			Err(err).
+			Msg("EvalSymlinks() failed")
 	}
 
 	logger.Trace().
 		Str("wd", wd).
-		Msg("Get absolute filepath for working dir")
-	wd, err = filepath.Abs(wd)
-	if err != nil {
-		logger.Fatal().
-			Str("wd", wd).
-			Err(err).
-			Msg("getting absolute path")
-	}
+		Msg("Running in directory")
 
-	logger.Trace().
-		Msgf("Evaluate symbolic links for %q.", wd)
-	wd, err = filepath.EvalSymlinks(wd)
-	if err != nil {
-		logger.Fatal().
-			Str("wd", wd).
-			Err(err).
-			Msg("failed evaluating symlinks")
-	}
-
-	logger.Trace().
-		Msgf("Change working directory to %q.", wd)
-	err = os.Chdir(wd)
-	if err != nil {
-		logger.Fatal().
-			Str("wd", wd).
-			Err(err).
-			Msg("failed to change working directory")
-	}
-
-	logger.Trace().
-		Msgf("Look up project in %q.", wd)
 	prj, foundRoot, err := lookupProject(wd)
 	if err != nil {
 		logger.Fatal().
@@ -763,11 +758,10 @@ func (c *cli) runOnStacks() {
 			Err(err)
 	}
 
-	if c.parsedArgs.Changed {
-		c.log("Running on changed stacks:")
-	} else {
-		c.log("Running on all stacks:")
-	}
+	log.Info().
+		Str("wd", c.wd()).
+		Bool("changed", c.parsedArgs.Changed).
+		Msg("Running command in stacks reachable from working directory")
 
 	logger.Trace().
 		Msg("Filter stacks by working directory.")
--- a/manager.go
+++ b/manager.go
@@ -15,6 +15,7 @@
 package terramate
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -200,21 +201,27 @@ func listChangedFiles(dir string, gitBaseRef string) ([]string, error) {
 		return nil, fmt.Errorf("listing untracked files: %v", err)
 	}
 
-	if len(untracked) > 0 {
-		log.Fatal().
-			Strs("files", untracked).
-			Msg("repository has untracked files")
-	}
-
 	uncommitted, err := g.ListUncommitted()
 	if err != nil {
 		return nil, fmt.Errorf("listing uncommitted files: %v", err)
 	}
 
+	if len(untracked) > 0 {
+		log.Error().
+			Strs("files", untracked).
+			Err(errors.New("repository has untracked files")).
+			Msg("Ensuring clean git status")
+	}
+
 	if len(uncommitted) > 0 {
-		log.Fatal().
+		log.Error().
 			Strs("files", uncommitted).
-			Msg("repository has uncommitted files")
+			Err(errors.New("repository has uncommitted files")).
+			Msg("Ensuring clean git status")
+	}
+
+	if len(uncommitted)+len(untracked) > 0 {
+		return nil, errors.New("git status is not clean")
 	}
 
 	baseRef, err := g.RevParse(gitBaseRef)
--- a/manager.go
+++ b/manager.go
@@ -25,6 +25,8 @@ import (
 	"github.com/mineiros-io/terramate/git"
 	"github.com/mineiros-io/terramate/hcl"
 	"github.com/mineiros-io/terramate/stack"
+
+	"github.com/rs/zerolog/log"
 )
 
 type (
@@ -199,7 +201,9 @@ func listChangedFiles(dir string, gitBaseRef string) ([]string, error) {
 	}
 
 	if len(untracked) > 0 {
-		return nil, fmt.Errorf("repository has untracked files: %v", untracked)
+		log.Fatal().
+			Strs("files", untracked).
+			Msg("repository has untracked files")
 	}
 
 	uncommitted, err := g.ListUncommitted()
@@ -208,7 +212,9 @@ func listChangedFiles(dir string, gitBaseRef string) ([]string, error) {
 	}
 
 	if len(uncommitted) > 0 {
-		return nil, fmt.Errorf("repository has uncommitted files: %v", uncommitted)
+		log.Fatal().
+			Strs("files", uncommitted).
+			Msg("repository has uncommitted files")
 	}
 
 	baseRef, err := g.RevParse(gitBaseRef)
--- a/run.go
+++ b/run.go
@@ -15,32 +15,30 @@
 package terramate
 
 import (
-	"fmt"
-	"os"
 	"os/exec"
 	"path/filepath"
 
-	"github.com/mineiros-io/terramate/project"
 	"github.com/mineiros-io/terramate/stack"
+
+	"github.com/rs/zerolog/log"
 )
 
 func Run(root string, stacks []stack.S, cmdSpec *exec.Cmd) error {
-	wd, err := os.Getwd()
-	if err != nil {
-		return fmt.Errorf("error getting current directory: %w", err)
-	}
-
 	for _, stack := range stacks {
 		cmd := *cmdSpec
 
-		stackdir, _ := project.FriendlyFmtDir(root, wd, stack.Dir)
-		fmt.Fprintf(cmd.Stdout, "[%s] running %s\n", stackdir, &cmd)
+		log.Info().
+			Str("stack", stack.Dir).
+			Str("cmd", cmd.String()).
+			Msg("Running command in stack")
+
 		cmd.Dir = filepath.Join(root, stack.Dir)
+
 		err := cmd.Run()
+
 		if err != nil {
 			return err
 		}
-		fmt.Fprintf(cmd.Stdout, "\n")
 	}
 
 	return nil
--- a/cmd/terramate/cli/cli_order_test.go
+++ b/cmd/terramate/cli/cli_order_test.go
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// FIXME(katcipis): improve CLI tests isolation
-//go:build ignore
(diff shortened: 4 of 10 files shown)


run.go
```diff
@@ -15,32 +15,30 @@
...
-	}
-
 	for _, stack := range stacks {
 		cmd := *cmdSpec
 
-		stackdir, _ := project.FriendlyFmtDir(root, wd, stack.Dir)
-		fmt.Fprintf(cmd.Stdout, "[%s] running %s\n", stackdir, &cmd)
+		log.Info().
+			Str("stack", stack.Dir).
+			Str("cmd", cmd.String()).
```
mariux, line 32: @katcipis @i4ki how can we show the user command here instead of the calculated executable?

mariux:
@katcipis @i4ki how to fix tests?

katcipis:
> @katcipis @i4ki how to fix tests?

It seems that output is being writen on "stderr" (it is injected on the cli logic) on the tests, but on the tests the log level should be the default "none" no ? Overall it is failing because the tests are pedantic with output being on stderr/stdout, unless the test explicitely commands to ignore them.

run.go
katcipis: After exec.Cmd is created we don't have that info anymore AFAIK, it already stores on the calculated executable. we would need to log this before when the exec.Cmd is being created (probably being created with: https://pkg.go.dev/os/exec#Command).

cmd/terramate/cli/cli.go
```diff
@@ -205,52 +200,46 @@ func newCLI(
...
-	wd, err = filepath.Abs(wd)
-	if err != nil {
-		return nil, fmt.Errorf("getting absolute path of %q: %w", wd, err)
-	}
+	if parsedArgs.Chdir != "" {
+		logger.Debug().
+			Str("cwd", cwd).
+			Str("dir", parsedArgs.Chdir).
+			Msg("Changing working directory")
+		err = os.Chdir(parsedArgs.Chdir)
```
katcipis, line 219: it was a previous design decision to test the cli logic as a function, so we avoid calling anything that changes the process working dir, since that changes the overall test process environment, so things breaking are probably related to os.Getwd/os.Chdir calls.

One alternative discussed at the time, that would improve isolation, is to test the cli by building/running the terramate binary, then since it is run as a separate process from the test itself this would not be an issue.

manager.go
```diff
@@ -199,7 +201,9 @@ func listChangedFiles(dir string, gitBaseRef string) ([]string, error) {
 	}
 
 	if len(untracked) > 0 {
-		return nil, fmt.Errorf("repository has untracked files: %v", untracked)
+		log.Fatal().
```
katcipis, line 204: As mentioned here: https://github.com/mineiros-io/terramate/pull/141
Calling Fatal on the middle of some computation in Go is usually not a good idea, specially because any added defers will not be executed on this case (and they could do important cleanup).

cmd/terramate/cli/cli.go
```diff
@@ -197,53 +197,46 @@ func newCLI(
...
+				Err(err).
+				Msg("Changing working directory failed")
+		}
 
-	logger.Trace().
-		Msgf("Change working directory to %q.", wd)
-	err = os.Chdir(wd)
-	if err != nil {
-		return nil, fmt.Errorf("failed to change working directory to %q: %w", wd, err)
+		cwd, err = os.Getwd()
```
katcipis, line 225:
```suggestion
		cwd = parsedArgs.Chdir
```

Should be safe since we just called "os.Chdir" with success (no need to call Getwd again, I think).

katcipis (commented):
Some comments regarding calling Fatal inside the terramate.Manager.

manager.go
```diff
@@ -208,7 +212,9 @@ func listChangedFiles(dir string, gitBaseRef string) ([]string, error) {
 	}
 
 	if len(uncommitted) > 0 {
-		return nil, fmt.Errorf("repository has uncommitted files: %v", uncommitted)
+		log.Fatal().
```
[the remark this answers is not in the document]
mariux: isn't this just initialization... and not core logic... this should be checked before even calling list or run and not within the action?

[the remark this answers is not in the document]
mariux: how can i change this to Error and return?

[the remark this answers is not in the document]
mariux: what is the concept for logging and error state?

run.go
mariux: why is this object created outside of the run function?

cmd/terramate/cli/cli.go
mariux: getwd returns the absoulte path.... chdir can be relative and is invalid after the chdir was executed

manager.go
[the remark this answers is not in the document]
i4ki: @mariux code in library packages must never fatal/abort. It's the client responsibility to decide if it should abort given some error. So in this case you can make:

```
    log.Error().
        .Strs("files", uncommited).
        .Msgf("directory %q has uncommited files", dir)

    return nil, errors.New("repository has uncommitted files")
```

The list of uncommitted files don't need to be in the error as they will not be used as decision making in the client, they just need to be logged. So I think something like the above is enough.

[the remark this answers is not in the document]
mariux: but then we need to duplicate error messages?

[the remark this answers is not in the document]
mariux: adjusted now we end up with a duplicate message in the end. ;) but fine for now, i would guess as we cannot ignore final message until all errors are printed?

run.go
i4ki: just to avoid passing stdin, stdout, stderr as arguments to `terramate.Run()`, as they are needed for tests. But you can change this, not big deal. You can change the Run to something like:

```
func Run(root string, stacks []stack.S, command []string, stdin io.Reader, stdout, stderr io.Writer) error
```

or create a run type containing the cmd spec (cmd, args, standard I/O, etc):

```
func Run(root string, stacks []stack.S, cmdSpec YourOwnCmdSpec) error {
```

manager.go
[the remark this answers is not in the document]
i4ki: unfortunately, or you dont log errors as they will be handled in the cli as `warn` or `fatal` anyway.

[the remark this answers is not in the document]
mariux: refactored to not duplicate the same message ;)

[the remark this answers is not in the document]
mariux:
```
2022-01-06T18:41:40+01:00 ERR Ensuring clean git status error="repository has untracked files" files=["x"]
2022-01-06T18:41:40+01:00 ERR Ensuring clean git status error="repository has uncommitted files" files=["stacks/organization/teams/teamA/terramate.tm.hcl"]
2022-01-06T18:41:40+01:00 FTL Aborting execution error="git status is not clean"
```

kassianh (approved):
lgtm :)

cmd/terramate/cli/cli.go
```diff
@@ -197,53 +197,46 @@ func newCLI(
...
-		Str("stack", wd).
-		Logger()
-
-	logger.Trace().
-		Msgf("Evaluate symbolic links for %q.", wd)
-	wd, err = filepath.EvalSymlinks(wd)
+	cwd, err := os.Getwd()
 	if err != nil {
-		return nil, fmt.Errorf("failed evaluating symlinks for %q: %w", wd, err)
+		log.Fatal().
```
kassianh, line 206: Is this supposed to be Fatal? Reading through the conversation further up I got the impression it should be Error not Fatal.

cmd/terramate/cli/cli.go
```diff
@@ -197,53 +197,46 @@ func newCLI(
...
+		}
 
-	logger.Trace().
-		Msgf("Change working directory to %q.", wd)
-	err = os.Chdir(wd)
-	if err != nil {
-		return nil, fmt.Errorf("failed to change working directory to %q: %w", wd, err)
+		cwd, err = os.Getwd()
+		if err != nil {
+			log.Fatal().
```
kassianh, line 227: And this

kassianh: Also shouldnt it be `logger` not `log`?

i4ki (dismissed):
The evalSymlinks is needed because our tests pass a `--chdir <path>` from `t.TempDir()` and in the OSX platform this function returns paths inside the `/var` link that points to `/private/var` whereas the paths returned from git command are all resolved, so they mismatch and tests fails. If you don't want to automatically resolve user paths by default, then you need to fix the tests to always resolve the temp dir passed as `--chdir` before calling terramate. The other changes are unrelated to this problem, but fine for me.

Having said that, the fix can be just:

```
diff --git a/cmd/terramate/cli/cli.go b/cmd/terramate/cli/cli.go
index 970c5b7..d8b2d60 100644
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -202,8 +202,11 @@ func newCLI(

        configureLogging(logLevel, logFmt, stderr)

-       log.Trace().
+       logger := log.With().
                Str("action", "newCli()").
+               Logger()
+
+       logger.Trace().
                Msg("Get working directory.")
        wd := parsedArgs.Chdir
        if wd == "" {
@@ -213,23 +216,25 @@ func newCLI(
                }
        }

-       logger := log.With().
-               Str("action", "newCli()").
-               Str("stack", wd).
-               Logger()
+       logger.Trace().
+               Msgf("Get absolute file path of %q.", wd)
+       wd, err = filepath.Abs(wd)
+       if err != nil {
+               return nil, fmt.Errorf("getting absolute path of %q: %w", wd, err)
+       }

        logger.Trace().
-               Msgf("Evaluate symbolic links for %q.", wd)
-       wd, err = filepath.EvalSymlinks(wd)
+               Msgf("stating directory %q", wd)
+       _, err = os.Stat(wd)
        if err != nil {
-               return nil, fmt.Errorf("failed evaluating symlinks for %q: %w", wd, err)
+               return nil, err
        }

        logger.Trace().
-               Msgf("Get absolute file path of %q.", wd)
-       wd, err = filepath.Abs(wd)
+               Msgf("Evaluate symbolic links for %q.", wd)
+       wd, err = filepath.EvalSymlinks(wd)
        if err != nil {
-               return nil, fmt.Errorf("getting absolute path of %q: %w", wd, err)
+               return nil, fmt.Errorf("failed evaluating symlinks for %q: %w", wd, err)
        }

        logger.Trace().
```

cmd/terramate/cli/cli.go
mariux: if this failed the system is broken anyway... so there is no way around a fatal error here in my opinion... also this is initialization... there should be no defers going on... this should never fail in tests

cmd/terramate/cli/cli.go
```diff
@@ -197,53 +197,46 @@ func newCLI(
...
-		return nil, fmt.Errorf("getting absolute path of %q: %w", wd, err)
-	}
+	if parsedArgs.Chdir != "" {
+		logger.Debug().
+			Str("cwd", cwd).
+			Str("dir", parsedArgs.Chdir).
+			Msg("Changing working directory")
+		err = os.Chdir(parsedArgs.Chdir)
+		if err != nil {
+			logger.Fatal().
```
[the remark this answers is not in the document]
mariux: this is also initialization there should be zero side effect with defers and so on...  also this should never fail in tests..

cmd/terramate/cli/cli.go
mariux: same as first getwd.. ;)

mariux:
> The evalSymlinks is needed because our tests pass a `--chdir <path>` from `t.TempDir()` and in the OSX platform this function returns paths inside the `/var` link that points to `/private/var` whereas the paths returned from git command are all resolved, so they mismatch and tests fails. If you don't want to automatically resolve user paths by default, then you need to fix the tests to always resolve the temp dir passed as `--chdir` before calling terramate. The other changes are unrelated to this problem, but fine for me.

i would really not fix tests with extra code to be honest... a chdir should not fail as long as the directory exists and is accessible by the caller.. 

evaluating symlink and changing to the evaluated dir should be avoided by all means as there are things like automounter that dismount unused mounts... so when not in the actual mounted path the automounter might not notice it still being used and just dismount. automounter works with symlinks in that case... so the implementation will fail on systems depending on short-lived mounts.. (terraform can run for multiple hours)

so i would argue to fix the tests in this case.

run.go
mariux: gotcha... for UX it would still be great to show the users command and not the evaluated executable... i will create a ticket so this can be fixed in the future... fine for now, i guess.. ;)

i4ki:
> i would really not fix tests with extra code to be honest... a chdir should not fail as long as the directory exists and is accessible by the caller..

yeah, chdir will not fail, it's just that user provides `--chdir /var/tmp/something` but git returns paths with `/private/var/tmp/something` and we have functions like `cli.filterStacksByWorkingDir()` that filters by cwd but then they don't match.

> evaluating symlink and changing to the evaluated dir should be avoided by all means as there are things like automounter that dismount unused mounts... so when not in the actual mounted path the automounter might not notice it still being used and just dismount. automounter works with symlinks in that case... so the implementation will fail on systems depending on short-lived mounts.. (terraform can run for multiple hours)

This is only valid if someone has a mounted path containing a symlink to somewhere outside the mount, so in this case you use the symlink just to keep the mount up but the process do not use it.

Something like:

```
/mnt/data/var -> /var
```

If that's the case, if you provide `/mnt/data/var` as `--chdir` the terramate (and terraform) will never use the `/mnt/data` mount anyway.

Now this phrase:

> evaluating symlink and changing to the evaluated dir should be avoided by all means...

this is very dangerous (as a general statement) and very easy to exploit if the symlink is in a globally writable directory like `/tmp`, even more if the tool is going to be run in CIs as in this case it probably run as root and so it can go anywhere in the filesystem. There are lots of attacks possible in this case (like TOCTOU) and the paths could even be protected and vulnerable. This would be a big issue if someone uses `terramate` automated in a backend app (maybe a cloud orchestration product).

But maybe makes sense in this case (really don't know)... if that's the case then we need to change the "project detection" to not use `git rev-parse --git-dir`.

Terraform seems to evaluate the symlinks at some entrypoints, not sure if all of them: 
https://github.com/hashicorp/terraform/search?q=EvalSymlinks

So I would say we have two options:

1. Keep evaluating user's symlink but then someone could not pin a `mount` by providing a symlink in an unused mount point.
2. Dont evaluate (never) and don't use git to discover the git repository path.

I'm ok with both, but `1` could enable some attack vectors. If you chose `2`, just remove this part: https://github.com/mineiros-io/terramate/blob/main/cmd/terramate/cli/cli.go#L952:L988

You don't need to fix the tests, we can do it if they fail.

i4ki:
> i would really not fix tests with extra code to be honest... a chdir should not fail as long as the directory exists and is accessible by the caller..

yeah, chdir will not fail, it's just that user provides `--chdir /var/tmp/something` but git returns paths with `/private/var/tmp/something` and we have functions like `cli.filterStacksByWorkingDir()` that filters by cwd but then they don't match.

> evaluating symlink and changing to the evaluated dir should be avoided by all means as there are things like automounter that dismount unused mounts... so when not in the actual mounted path the automounter might not notice it still being used and just dismount. automounter works with symlinks in that case... so the implementation will fail on systems depending on short-lived mounts.. (terraform can run for multiple hours)

This is only valid if someone has a mounted path containing a symlink to somewhere outside the mount, so in this case you use the symlink just to keep the mount up but the process do not use it.

Something like:

```
/mnt/data/var -> /var
```

If that's the case, if you provide `/mnt/data/var` as `--chdir` the terramate (and terraform) will never use the `/mnt/data` mount anyway.

Now this phrase:

> evaluating symlink and changing to the evaluated dir should be avoided by all means...

this is very dangerous (as a general statement) and very easy to exploit if the symlink is in a globally writable directory like `/tmp`, even more if the tool is going to be run in CIs as in this case it probably run as root and so it can go anywhere in the filesystem. There are lots of attacks possible in this case (like TOCTOU) and the paths could even be protected and vulnerable. This would be a big issue if someone uses `terramate` automated in a backend app (maybe a cloud orchestration product).

But maybe makes sense in this case (really don't know)... if that's the case then we need to change the "project detection" to not use `git rev-parse --git-dir`.

Terraform seems to evaluate the symlinks at some entrypoints, not sure if all of them: 
https://github.com/hashicorp/terraform/search?q=EvalSymlinks

So I would say we have two options:

1. Keep evaluating user's symlink but then someone could not pin a `mount` by providing a symlink in an unused mount point.
2. Dont evaluate (never) and don't use git to discover the git repository path.

I'm ok with both, but `1` could enable some attack vectors. If you chose `2`, just remove this part: https://github.com/mineiros-io/terramate/blob/main/cmd/terramate/cli/cli.go#L952:L988

You don't need to fix the tests, we can do it if they fail.

i4ki:
The project's testing is now ready for `Fatal()` and `Chdir()` calls anywhere, so you can enable the tests again.

katcipis (approved):
:rocket:

PR #141, opened 2022-01-06
fix: avoid calling log.Fatal on core logic

katcipis:
The [PR](https://github.com/mineiros-io/terramate/pull/138) introduces log.Fatal calls inside the core logic (manager, cli), that will abort the tests suddenly making the output of the tests to disappear (the test program is exited abruptly). From [zerolog docs](https://pkg.go.dev/github.com/rs/zerolog#Logger.Fatal):

"Fatal starts a new message with fatal level. The os.Exit(1) function is called by the Msg method, which terminates the program immediately."

One reason to not use Fatal/os.Exit on the middle of some logic is what happened right now, makes it harder to test (the test caller will just exit), but it also can be problematic if there are deferred calls on the stack that will cleanup things if something went wrong.

To ensure that you will execute deferred calls + being able to test you need to return an error or panic, panic will execute all defers + will show properly with stack info on the test when it fails + it can be recovered from on a test for inspection.
There seems to be some other change in behavior, but now the tests fails with error messages in a more graceful manner (they are not aborted mid execution).

Since the tests are not ready to handle new data on stderr I also set the level to "fatal" for the tests for now, we can re-assess later.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -206,9 +206,7 @@ func newCLI(
 
 	cwd, err := os.Getwd()
 	if err != nil {
-		log.Fatal().
-			Err(err).
-			Msg("Getwd() failed")
+		return nil, fmt.Errorf("Getwd() failed: %v", err)
 	}
 
 	if parsedArgs.Chdir != "" {
@@ -218,18 +216,12 @@ func newCLI(
 			Msg("Changing working directory")
 		err = os.Chdir(parsedArgs.Chdir)
 		if err != nil {
-			logger.Fatal().
-				Str("cwd", cwd).
-				Str("dir", parsedArgs.Chdir).
-				Err(err).
-				Msg("Changing working directory failed")
+			return nil, fmt.Errorf("Changing workind dir failed: %v", err)
 		}
 
 		cwd, err = os.Getwd()
 		if err != nil {
-			log.Fatal().
-				Err(err).
-				Msg("Getwd() failed")
+			return nil, fmt.Errorf("Getwd() failed: %v", err)
 		}
 	}
 
--- a/cmd/terramate/cli/cli_test.go
+++ b/cmd/terramate/cli/cli_test.go
@@ -392,7 +392,7 @@ func (ts tscli) run(args ...string) runResult {
 
 	allargs := []string{}
 	if ts.chdir != "" {
-		allargs = append(allargs, "--chdir", ts.chdir)
+		allargs = append(allargs, "--chdir", ts.chdir, "--log-level", "fatal")
 	}
 
 	allargs = append(allargs, args...)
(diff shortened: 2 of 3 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -206,9 +206,7 @@ func newCLI(
 
 	cwd, err := os.Getwd()
 	if err != nil {
-		log.Fatal().
-			Err(err).
-			Msg("Getwd() failed")
+		return nil, fmt.Errorf("Getwd() failed: %v", err)
```
mariux, line 209: but this should be zero log and executed here to show correct line in coffee . how to do this?

katcipis: hmm there are some alternatives. Some libraries embed automatically on error instances this information, so we could keep using the error and have line of code info. Another alternative is use log.Panic(), it will produce the message you want + a panic stack trace.

mariux: makes not much sense to me.. it's a fatal thing and not a broken thing.. ;) 

as this is initialization there should not be any cleanup to do anyway.. why not just exit?

katcipis: Just returning the error will also just exit with an error status code. The fastest way right now would be to log with an error and still return the error, calling Exit here (or fatal) will not work well with how the cli tests are written today, we can change the tests to work with code calling Exit, but would need some time for this (change the test approach, probably not much hopefully).

mariux: but this is not using zerolog then.. which is now the logging used here..

katcipis: Yeah it would not use it for Fatal/exiting. I'm going to work on changing the tests so calling exit/log.Fatal would be OK anywhere on the code, hopefully it won't take much time.

mariux: btw tests are still failing.. ;)

katcipis:
Will change the cli test approach so it wont fail/abort when exit/fatal is called.

cmd/terramate/cli/cli.go
mariux: we can change it to error but how to return then?

katcipis: yeah there is something else related to os.Getwd/Chdir (which was changed from original code on main) that is breaking the tests, Im taking a look into it. What this was "fixing" is the test aborting all of a sudden and then there was no error message at all or even the name of the test that failed etc.

mariux: maybe the initialization should go to main.. not sure why we need to test commandline parsing.. and chdir..

katcipis: It is supposed to an end to end test, like ensuring that running:

```
terramate init
```

Works, in an e2e manner. We tried to do this extracting all the logic to cli.Run (so main doesn't to anything), but then some side effect operations can cause problems since just calling a function cli.Run is not as isolated as running a separate process. Overall doing e2e tests with a separate process will increase isolation of the tests, so an aborting terramate can't kill the entire test process.

PR #144, opened 2022-01-06
tests: remove logging on tests output

katcipis:
For now it is disabled, someone can change it to debug if required but usually just the error message is enough to help (at least for me). The default is tracing, so there is so much output when a test fails that it is hard to check the error message itself.


i4ki (approved):
lgtm

PR #145, opened 2022-01-06
refactor: make e2e tests use terramate binary

i4ki:
In addition to the test refactor, the cli logging was changed to fatal in case of errors instead of returning it to main because this change makes it impossible to check using the sentinel errors and then there's no benefit for keeping the `Fatal()` in the `main()` function, far away from when it happens. So now all the `cli` package code can be moved to main package, but we can do this in a separate PR.


i4ki:
@katcipis look at the coverage stats =(

The `go build` does not support a `-cover` flag, then searching for workarounds I found this:
https://github.com/golang/go/issues/33463#issuecomment-518160521

But not sure if supports passing args... very ugly and probably also will need a "coverage merge magic".
Ideas?

kassianh:
Looks fine to me but it's also a bit beyond me..

katcipis:
> @katcipis look at the coverage stats =(
> 
> The `go build` does not support a `-cover` flag, then searching for workarounds I found this: [golang/go#33463 (comment)](https://github.com/golang/go/issues/33463#issuecomment-518160521)
> 
> But not sure if supports passing args... very ugly and probably also will need a "coverage merge magic". Ideas?

It is not awesome but I'm OK with just letting it drop for now (the coverage), we already plan on extracting tests that are unrelated to cli stuff. Then for some cli dependent logic we could also extract it in a way that can be tested more isolated, leaving the e2e tests for very basic full integration (which still drops coverage maybe, but not as much).

Or we can go with the hack, but it is so very hacky :laughing:

katcipis (dismissed):
LGTM + some suggestions on how to improve the logging fatal calls, but can also be done on another PR.

katcipis (dismissed):
Thanks man :100:

katcipis (approved):

PR #146, opened 2022-01-07
chore: add release automation

katcipis:
Im going for this workflow:

* Create PR to change version from -dev to final one, like 0.0.9
* Reviewed/merged the PR
* Run locally on main: `make release/tag` it will automatically use the VERSION file (also used by the code itself)
* Release workflow is triggered by tag creation + goreleaser does it magic

It also generates the changelog automatically. For homebrew support it is a little more involved, maybe we can do it later, but at least for now we should get darwin binaries + linux/etc.

Not married with this particular workflow/no strong opinions, so feel free to suggest alternatives :-)

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,32 @@
+builds:
+  - main: ./cmd/terramate
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+      - darwin
+      # - windows (when we add win support)
+      #
+archives:
+  - replacements:
+      darwin: Darwin
+      linux: Linux
+      #windows: Windows (when we add win support)
+      386: i386
+      amd64: x86_64
+
+checksum:
+  name_template: 'checksums.txt'
+
+snapshot:
+  name_template: "{{ incpatch .Version }}-next"
+
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^spec:'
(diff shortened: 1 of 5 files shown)


.goreleaser.yaml
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+builds:
+  - main: ./cmd/terramate
+    env:
+      - CGO_ENABLED=0
```
katcipis, line 18: Why disabling CGO here: https://goreleaser.com/limitations/cgo/

i4ki (approved):
lgtm

Makefile
```diff
@@ -68,6 +68,13 @@ install:
 clean:
 	rm -rf bin/*
 
+## creates a new release tag
+.PHONY: release/tag
+release/tag: VERSION?=v$(shell cat VERSION)
```
katcipis, line 73: The tag needs a "v" on the prefix because of how Go handles versions (v<semver>, like the current releases: https://github.com/mineiros-io/terramate/releases).

mariux (commented):
lgtm

PR #147, opened 2022-01-07
test: disable logs on test run

katcipis:
We can allow it to be configured if we ever test log output, but for now AFAIK we don't use log output on tests for validation, so we can just disable it to avoid tests failing or having to ignore stderr on a bunch of tests.

### DIFF
--- a/cmd/terramate/cli/cli_runner_test.go
+++ b/cmd/terramate/cli/cli_runner_test.go
@@ -71,7 +71,7 @@ func (ts tscli) run(args ...string) runResult {
 		allargs = append(allargs, "--chdir", ts.chdir)
 	}
 	if len(args) > 0 {
-		// Avoid failing test when calling just terramate with no params
+		// Avoid failing test when calling just terramate with no args
 		allargs = append(allargs, "--log-level", "fatal")
 	}
 	allargs = append(allargs, args...)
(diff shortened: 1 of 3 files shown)


cmd/terramate/cli/cli_runner_test.go
```diff
@@ -70,7 +70,10 @@ func (ts tscli) run(args ...string) runResult {
 	if ts.chdir != "" {
 		allargs = append(allargs, "--chdir", ts.chdir)
 	}
-
+	if len(args) > 0 {
+		// Avoid failing test when calling just terramate with no params
```
katcipis, line 74:
```suggestion
		// Avoid failing test when calling just terramate with no args
```

cmd/terramate/cli/cli_runner_test.go
```diff
@@ -70,7 +70,10 @@ func (ts tscli) run(args ...string) runResult {
 	if ts.chdir != "" {
 		allargs = append(allargs, "--chdir", ts.chdir)
 	}
-
+	if len(args) > 0 {
+		// Avoid failing test when calling just terramate with no args
+		allargs = append(allargs, "--log-level", "fatal")
+	}
```
mariux, line 76: added #148 as alternative.... if no stderr/stdout is set expectation could be not to test for it?

mariux:
merged #148 now.. feel free to revert.. an use this one..

PR #149, opened 2022-01-10
fix: stacks listed twice

i4ki:
The root cause was stacks listed twice and the incorrect one was missing the changed flag, so the run-order put them in a map, randomly overwriting the entry, so in the end stacks could have a changed flag or not, so the run-order was broken.

It's a bit hard to reproduce because it only happens when a stack detected as changed also had a module that changed, so I'm going to provide a testcase later this week.


katcipis (approved):
Thanks for the quick fix :rocket:

PR #150, opened 2022-01-10
fix: explicitly mark as changed stacks that changed due to module changes

i4ki:
This affects run-order of changed stacks due to module changes.


katcipis (approved):

soerenmartius:
Cool!

PR #156, opened 2022-01-10
refactor: extract code generation from writing to file

katcipis:
The idea is to have internal functions that only generate code, but don't write them, so we can use it to detect if generated code is updated by generating code + comparing with code present on stack. This PR only extracts the functions, no new functionality added.


generate/generate.go
```diff
@@ -381,6 +425,22 @@ func sortedAttributes(attrs hclsyntax.Attributes) []*hclsyntax.Attribute {
 	return sorted
 }
 
+func writeGeneratedCode(target string, code []byte) error {
+	logger := log.With().
+		Str("action", "writeGeneratedCode()").
+		Str("target", target).
```
mariux, line 431:
```suggestion
		Str("file", target).
```

mariux (changes requested):

generate/generate.go
```diff
@@ -216,48 +230,64 @@ func generateStackLocals(
...
-	return os.WriteFile(genfile, tfcode, 0666)
+	logger.Debug().Msg("Stack has backend config, saving generated code.")
+
+	if err := writeGeneratedCode(targetBackendCfgFile, tfcode); err != nil {
+		return fmt.Errorf(
+			"stack %q: %w: saving code at %q",
+			stackpath,
+			err,
+			targetBackendCfgFile,
+		)
```
mariux, line 269: should be log.Error() with context

generate/generate.go
```diff
@@ -216,48 +230,64 @@ func generateStackLocals(
...
+	}
+
+	logger.Debug().Msg("Saved stack generated code.")
```
mariux, line 272: no need to log same information before and after an action.
```suggestion
```

generate/generate.go
```diff
@@ -295,7 +325,21 @@ func loadStackBackendConfig(root string, configdir string, evalctx *eval.Context
...
+	logger.Trace().Msg("Add stack metadata evaluation namespace.")
+
+	if err := stackMetadata.SetOnEvalCtx(evalctx); err != nil {
+		return nil, fmt.Errorf("stack %q: %v", stackpath, err)
+	}
+
+	logger.Trace().Msg("Add global evaluation namespace.")
+
+	if err := globals.SetOnEvalCtx(evalctx); err != nil {
+		return nil, fmt.Errorf("stack %q: %v", stackpath, err)
```
mariux, line 342: should be log.Error() with context.

generate/generate.go
```diff
@@ -295,7 +325,21 @@ func loadStackBackendConfig(root string, configdir string, evalctx *eval.Context
...
-		return loadStackBackendConfig(root, filepath.Dir(configdir), evalctx)
+		return generateBackendCfgCode(root, stackpath, stackMetadata, globals, filepath.Dir(configdir))
+	}
+
+	evalctx := eval.NewContext(stackpath)
+
+	logger.Trace().Msg("Add stack metadata evaluation namespace.")
+
+	if err := stackMetadata.SetOnEvalCtx(evalctx); err != nil {
+		return nil, fmt.Errorf("stack %q: %v", stackpath, err)
```
mariux, line 336: should be log.Error() with context.

generate/generate.go
```diff
@@ -216,48 +230,64 @@ func generateStackLocals(
...
-	}
-
-	logger.Debug().
-		Str("genfile", genfile).
-		Msg("Load stack backend config.")
-	tfcode, err := loadStackBackendConfig(root, stackpath, evalctx)
+	logger.Trace().Msg("Generating code.")
+	tfcode, err := generateBackendCfgCode(root, stackpath, stackMetadata, globals, stackpath)
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrBackendConfigGen, err)
```
mariux, line 253: should be log.Error() with context.

generate/generate.go
```diff
@@ -158,44 +141,75 @@ func Do(root string) error {
...
+	if len(stackLocalsCode) == 0 {
+		logger.Debug().Msg("Stack has no locals to be generated, nothing to do.")
+		return nil
+	}
+
+	logger.Debug().Msg("Stack has locals, saving generated code.")
+
+	if err := writeGeneratedCode(targetLocalsFile, stackLocalsCode); err != nil {
+		err = errutil.Chain(ErrExportingLocalsGen, err)
+		return fmt.Errorf(
```
mariux, line 174: should be log.Error() with context.

generate/generate.go
```diff
@@ -158,44 +141,75 @@ func Do(root string) error {
...
 
-	logger.Trace().
-		Msg("Get generated file path.")
-	genfile := filepath.Join(stackpath, LocalsFilename)
-	if err := checkFileCanBeOverwritten(genfile); err != nil {
-		return err
+	stackLocalsCode, err := generateStackLocalsCode(root, stackpath, stackMetadata, globals)
+	if err != nil {
+		err = errutil.Chain(ErrExportingLocalsGen, err)
+		return fmt.Errorf("stack %q: %w", stackpath, err)
```
mariux, line 162: should be log.Error() with context.

mariux (changes requested):

katcipis:
@mariux what do you mean with the comments "should be log.Error() with context." ? We should call "log.Fatal().Err()" ? That wouild make it very hard to test the generate function, since today we can test error situations by calling the function + checking the error. If the idea is to call fatal to abort I would suggest we return the error here, and the error has context inside of it + call log.Fatal on the cli code (which called this function).

Or would it be something else ?

generate/generate.go
i4ki, line 162: I see you're spliting the err return multiple times, maybe if you create a helper alias it improves to keep it one line:
```suggestion
		return errorf("stack %q: %w", stackpath, errchain(ErrExportingLocalsGen, err))
```
but feel free to ignore also, just a thought.

generate/generate.go
```diff
@@ -216,48 +230,63 @@ func generateStackLocals(
 		localsBody.SetAttributeValue(name, localsAttrs[name])
 	}
 
-	logger.Debug().
-		Msg("Write file.")
 	tfcode := AddHeader(gen.Bytes())
```
i4ki, line 233: I know this line is not changed, but this `AddHeader()` looks like it's doing something completely different. Maybe rename to `PrependHeaderBytes()` ?

i4ki (approved):
lgtm

generate/generate.go
katcipis: this stayed on multiple lines more because of the previous code dealing with err lists. wil fix.

mariux:
> If it is about logging on the "error" level, it will create some duplication since I think the cli code calling functions also logs errors as Fatal, but the Error log entry will have some more detailed info indeed, so we will get:
> 
>     * log level error inside generate with internal details
> 
>     * log level fatal on cli given that generate returned error
> 
> 
> Is that what you want/mean here ?

it should not be Fatal ofc as this would end the program? 

But shouldn't we log the Error where it happens? as we only have access to the context at this place... also if we add Caller() the line of the error would be reported wrongly if not triggered at the right line of code? 

So this should be something like `log.Error().Str("stack", stack).Err(err).Msg("failed whatever")`

katcipis:
> > If it is about logging on the "error" level, it will create some duplication since I think the cli code calling functions also logs errors as Fatal, but the Error log entry will have some more detailed info indeed, so we will get:
> > ```
> > * log level error inside generate with internal details
> > 
> > * log level fatal on cli given that generate returned error
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is that what you want/mean here ?
> 
> it should not be Fatal ofc as this would end the program?
> 
> But shouldn't we log the Error where it happens? as we only have access to the context at this place...

Usually context is added on the error instance itself, and there are different ways to do that, like a plain formatted string, or having an actual error type with its fields etc that can be inspected by the caller, etc. The disadvantage of logging where an error happened is that it may happen that the error is logged multiple times, but at different levels on the stack (if we log + return the error).

 also if we add Caller() the line of the error would be reported wrongly if not triggered at the right line of code?
> 
> So this should be something like `log.Error().Str("stack", stack).Err(err).Msg("failed whatever")`

Hmm and it would be OK having that + on the cli code `log.Fatal().Err(err)` will be called, with the same error instance ?

i4ki (approved):
:rocket:

PR #157, opened 2022-01-10
feat: run fail if unclean repository

i4ki:


manager.go
```diff
@@ -23,13 +23,16 @@ import (
...
 
+	"github.com/madlambda/spells/errutil"
 	"github.com/mineiros-io/terramate/git"
 	"github.com/mineiros-io/terramate/hcl"
 	"github.com/mineiros-io/terramate/project"
 	"github.com/mineiros-io/terramate/stack"
 	"github.com/rs/zerolog/log"
 )
 
+const ErrDirtyRepo errutil.Error = "the repository is not clean"
```
mariux, line 34: wdyt?

```suggestion
const ErrDirtyRepo errutil.Error = "ERR_GIT_REPOSITORY_NOT_CLEAN"
```

manager.go
```diff
@@ -457,6 +390,117 @@ func (m *Manager) moduleChanged(
...
+func checkRepoIsClean(g *git.Git) error {
+	logger := log.With().
+		Str("action", "checkRepoIsClean()").
+		Logger()
+
+	logger.Debug().
+		Msg("Get list of untracked files.")
+	untracked, err := g.ListUntracked()
+	if err != nil {
+		return fmt.Errorf("listing untracked files: %v", err)
```
mariux, line 475: should be log.Error() with context.

manager.go
```diff
@@ -457,6 +390,117 @@ func (m *Manager) moduleChanged(
...
+	}
+
+	logger.Debug().
+		Msg("Get list of uncommitted files in dir.")
+	uncommitted, err := g.ListUncommitted()
+	if err != nil {
+		return fmt.Errorf("listing uncommitted files: %v", err)
```
mariux, line 482: should be log.Error() with context.

manager.go
```diff
@@ -457,6 +390,117 @@ func (m *Manager) moduleChanged(
...
+	}
+
+	if len(untracked) > 0 {
+		logger.Warn().
+			Strs("files", untracked).
+			Msg("repository has untracked files")
```
mariux, line 488: in case of run this should be error level how will this work?

mariux (dismissed):

manager.go
i4ki: As it's the caller who decides if it's a warning or an error, we can change those WARN to TRACE and return the error and let the caller log a WARN or an ERR/FTL. 

Currently it shows this:
Screenshot_2022-01-11_10-01-48

manager.go
i4ki: Do you mean you want the error messages / logs to be machine-oriented by default? If so, then we need to change a lot of stuff because in Go the error messages are human-friendly and we have been following this pattern since the beginning.

katcipis: Some examples of human-friendly error sentinels from Go stdlib/os: https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/internal/oserror/errors.go;drc=refs%2Ftags%2Fgo1.17.6;l=13

We can do something different, just explaining the rationale behind error sentinels that consist of a simple human readable message (with no upper cases,etc). It is usually done that way so it can be composed with other errors, forming a longer phrase explaining the whole issue.

Some more info [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) (as mentioned, logging is a different topic, but here we are talking about error values/sentinels).

katcipis (dismissed):

manager.go
mariux: gotcha.... ;) sorry for the noise... go.... ;)

katcipis (approved):
:rocket:

PR #162, opened 2022-01-12
feat: add generate check

katcipis:
This adds the core function on the package generate, it is still not used by run or other cli commands, so should not affect behavior.


generate/generate.go
```diff
@@ -141,6 +98,93 @@ func Do(root string) error {
...
+				StackDir: metadata.Path,
+				Filename: LocalsFilename,
+			})
+		}
+
+		return nil
+	})
+
+	if len(errs) > 0 {
+		err := fmt.Errorf("checking outdated code: %v : and %d other errors", errs[0], len(errs)-1)
```
katcipis, line 181: Inspired on: https://github.com/golang/go/blob/3d3f5d912b776424e50be276bc3a4ae02bf8d143/src/go/scanner/errors.go#L93

But in a more hacky way, maybe we need to model something like this ErrorList from Go (hcl.Diagnose is a worse name IMHO, but it is quite similar, including the Error implementation which is almost identical =P).

generate/generate.go
```diff
@@ -440,34 +484,110 @@ func writeGeneratedCode(target string, code []byte) error {
...
+			errs = append(errs, fmt.Errorf(
+				"stack %q: %w: %v",
+				stackpath,
+				ErrLoadingGlobals,
+				err))
+			continue
+		}
+
+		logger.Trace().Msg("Calling iterator.")
+		if err := iter(stackpath, stackMetadata, globals); err != nil {
```
i4ki, line 587: funny haha

generate/generate_check_test.go
```diff
@@ -0,0 +1,206 @@
...
+	stack2Dir := "/" + stack2.RelPath()
+
+	got, err := generate.Check(s.RootDir())
+	assert.NoError(t, err)
+	assertOutdatedEquals(t, got, []generate.Outdated{})
+
+	// Checking detection when there is no config generated yet
+	// for both locals and backend config
+	stack1.CreateConfig(
+		hcldoc(
```
i4ki, line 43: cool :)

test/sandbox/sandbox.go
```diff
@@ -202,6 +202,14 @@ func (s S) LoadMetadata() terramate.Metadata {
 	return meta
 }
 
+// Generate generates code for all stacks on the sandbox
+func (s S) Generate() {
```
i4ki, line 206: cool :)

i4ki (approved):
left some comments, but feel free to ignore or address in the other PRs.

generate/generate.go
```diff
@@ -141,6 +98,93 @@ func Do(root string) error {
...
+		metadata terramate.StackMetadata,
+		globals *terramate.Globals,
+	) error {
+		logger := log.With().
+			Str("action", "generate.Check()").
+			Str("path", root).
+			Str("stack", stackpath).
+			Logger()
+
+		logger.Trace().Msg("Checking for outdated backend cfg code on stack.")
```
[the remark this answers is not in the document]
katcipis: ah plain mistake on my side, sorry for that (I moved code around a lot =P)

generate/generate.go
katcipis: yeah, at least here it is internal to the function and I convert it to a single err instance, that is why I'm OK with it =P . What would you suggest ?

generate/generate.go
```diff
@@ -440,34 +484,110 @@ func writeGeneratedCode(target string, code []byte) error {
...
+	}
+
+	if !info.IsDir() {
+		return fmt.Errorf("project's root %q is not a directory", root)
 	}
 
 	return nil
 }
+
+type stackIterator func(
```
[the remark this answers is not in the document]
katcipis: hmm yeah..but also calling it just callback seems off...will try to think about something

generate/generate.go
katcipis: I think I was remembering some iter stuff from C++ :laughing: (but it is not what is happening here, so naming is off indeed..but I was thinking about it =P)

generate/generate.go
katcipis: overall for dealing with N operations, where some may fail and some may not, like they are orthogonal to each other, we could just return a slice of Results, and then each result may have an error (failed) or success (err will be nil). It is one of the ideas I had one of these days for generate...but Im not sure and don't wanted to rock the boat with this PR :laughing:

i4ki: it's ok how it's right now. I meant the idea of having `if err.HasError()` or something like that...

i4ki (approved):
lgtm

PR #163, opened 2022-01-13
feat: add check for outdated generated code on terramate run

katcipis:
Changed the design to make it easier to check only a specific set of stacks, was thinking way too much on terms of "check all the project", but that doesn't work well with cwd etc. Certainly space for future improvements, but should work. Here is an example script:

```sh
#!/bin/bash

set -o nounset

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-1
terramate stacks init stacks/stack-1

mkdir -p stacks/stack-2
terramate stacks init stacks/stack-2


echo "terramate run: should work:"
echo

terramate run cat terramate.tm.hcl

echo

cat > stacks/stack-1/terramate.tm.hcl <<- EOM
stack {}

export_as_locals {
  path = terramate.path
}
EOM

echo
echo "terramate run: should fail for all stacks:"
echo

terramate run cat terramate.tm.hcl

echo
echo "terramate run: should fail for stack-1:"
echo

terramate run -C stacks/stack-1 cat terramate.tm.hcl

echo
echo "terramate run: should work for stack-2:"
echo

terramate run -C stacks/stack-2 cat terramate.tm.hcl

cat > stacks/stack-2/terramate.tm.hcl <<- EOM
stack {}

terramate {
  backend "test" {}
}

export_as_locals {
  path = terramate.path
}
EOM

echo
echo "terramate run: should fail for all stacks:"
echo

terramate run cat terramate.tm.hcl

echo
echo "terramate generate: updating code"

terramate generate

echo "terramate generate: code updated"
echo "terramate run: should work for all stacks:"
echo

terramate run cat terramate.tm.hcl
```


mariux (commented):
:heart:

cmd/terramate/cli/cli_runner_test.go
```diff
@@ -116,6 +116,8 @@ func assertRun(t *testing.T, got runResult) {
 func assertRunResult(t *testing.T, got runResult, want runExpected) {
 	t.Helper()
 
+	assert.EqualInts(t, want.Status, got.Status, "exit status mismatch")
```
i4ki, line 119: I think checking the status later is better because you need to see the stdout/stderr to understand the issue...

i4ki (dismissed):

cmd/terramate/cli/cli_runner_test.go
katcipis: Well when it failed for me it was worse because the output check then made no sense at all without me knowing that the exit status was 0 instead of error....maybe we need to do all checks with Error to avoid stopping on the first error that happens here, felt like I had only half the information and on my case it was the half that helped me less.

i4ki (approved):

PR #166, opened 2022-01-14
refactor: move metadata loading to stack pkg

i4ki:


metadata_test.go
```diff
@@ -149,14 +138,22 @@ func TestLoadMetadata(t *testing.T) {
...
-			metadata, err := terramate.LoadMetadata(s.RootDir())
-
+			stackEntries, err := terramate.ListStacks(s.RootDir())
 			if tcase.wantErr != nil {
 				assert.IsError(t, err, tcase.wantErr)
 			}
 
-			if diff := cmp.Diff(tcase.want, metadata, cmpopts.IgnoreUnexported(tcase.want)); diff != "" {
-				t.Fatalf("want %+v != got %+v.\ndiff:\n%s", tcase.want, metadata, diff)
+			gotMetadata := []stack.Metadata{}
```
i4ki, line 146: I kept this `metadata_test.go` just to make sure metadata is loaded in the same way as before but this better to be moved to stack tests in the future.

exported_locals.go
```diff
@@ -187,13 +188,15 @@ func (r exportedLocalExprs) eval(meta StackMetadata, globals *Globals) (Exported
 
 	logger.Trace().
 		Msg("Add proper namespace for stack metadata evaluation.")
-	if err := meta.SetOnEvalCtx(evalctx); err != nil {
+	err := evalctx.SetNamespace("terramate", meta.ToCtyMap())
```
katcipis, line 191: I think this will also be duplicated on globals eval now (why I had the SetOnEvalCtx, which I didn't liked anyway :laughing: ). Maybe we could have our own evalctx instance that is more domain specific and has something like:

```
evalctx.SetMetadata(meta)
evalctx.SetGlobals(globals)
```

And the namespace names + how to translate from meta/globals to hcl eval stays there. It could do the ToCtyMap() for stack.Metadata, so it would not need any Cty specific knowledge. This was not well handled before either...just thinking if now we can do something better (I don't like spreading these cty stuff around x_x..but didn't had a good idea for that).

i4ki: yeah we can do this in a separate PR, not sure yet where this guy fits.

katcipis: man..me neither... I was honestly annoyed...like..where the fuck does this fit ...no good idea yet :laughing:

generate/generate_check_test.go
```diff
@@ -28,14 +29,16 @@ func TestCheckReturnsOutdatedStackFilenames(t *testing.T) {
 	stack1 := s.CreateStack("stacks/stack-1")
 	stack2 := s.CreateStack("stacks/stack-2")
 
-	stack1Dir := stack1.RelPath()
-	stack2Dir := stack2.RelPath()
+	stack1val, err := tmstack.Load(s.RootDir(), stack1.Path())
+	assert.NoError(t, err)
+	stack2val, err := tmstack.Load(s.RootDir(), stack2.Path())
+	assert.NoError(t, err)
```
katcipis, line 35: Maybe something like this, leveraging the sandbox:

```go
	stack1 := s.LoadStack(stackEntry1.Path())
	stack2 := s.LoadStack(stackEntry2.Path())
```

Now that the test also has stack instances we could call the sandbox stack an "stackEntry", or else it is a little confusing.

globals_test.go
```diff
@@ -792,15 +794,20 @@ func TestLoadGlobalsErrors(t *testing.T) {
...
-
 			for _, c := range tcase.configs {
 				path := filepath.Join(s.RootDir(), c.path)
 				test.AppendFile(t, path, config.Filename, c.body)
 			}
 
-			for _, stackMetadata := range metadata.Stacks {
-				_, err := terramate.LoadStackGlobals(s.RootDir(), stackMetadata)
+			stackEntries, err := terramate.ListStacks(s.RootDir())
+			// TODO(i4k): this better not be tested here.
```
katcipis, line 803: Indeed =P

metadata_test.go
katcipis: I was wondering about this, thanks for the comment :100:

test/sandbox/sandbox.go
```diff
@@ -193,25 +194,31 @@ func (s S) Git() Git {
...
-	meta, err := terramate.LoadMetadata(s.RootDir())
+	err := generate.Do(s.RootDir(), s.RootDir())
 	assert.NoError(s.t, err)
-	return meta
 }
 
-// Generate generates code for all stacks on the sandbox
-func (s S) Generate() {
+// LoadStacks load all stacks from sandbox rootdir.
+func (s S) LoadStacks() []stack.S {
```
katcipis, line 206: Ah we even have a LoadStacks, so maybe also having a LoadStack for a single stack makes sense (one of the tests seemed like could benefit from it).

katcipis (dismissed):
LGTM, thanks for the improvements. Left some non-blocks/potentially stupid comments. Can ignore or address on future PRs.

generate/generate.go
```diff
@@ -183,7 +180,7 @@ func CheckStack(root string, stackDir string) ([]string, error) {
 func writeStackLocalsCode(
 	root string,
 	stackpath string,
-	stackMetadata terramate.StackMetadata,
+	stackMetadata stack.Metadata,
```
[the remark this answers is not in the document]
i4ki: Si, me gusto mucho! :laughing:

katcipis (approved):

PR #167, opened 2022-01-14
fix: print metadata filtered by cwd

i4ki:


cmd/terramate/cli/cli_metadata_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+stack "/somedir/stack3":
+	terramate.name="stack3"
+	terramate.path="/somedir/stack3"
+	terramate.description=""
+
+stack "/somedir/stack4":
+	terramate.name="stack4"
+	terramate.path="/somedir/stack4"
+	terramate.description=""
```
katcipis, line 123: We can add later, but here we are also missing the scenario where wd is a dir with no stacks inside (like we tested for globals listing).

katcipis (approved):
Thanks for adding tests :100:

PR #168, opened 2022-01-14
fix: print globals filtered by cwd

i4ki:


cmd/terramate/cli/stacks_globals_test.go
```diff
@@ -125,6 +127,67 @@ stack "/stacks/stack-1":
...
+			},
+			want: runExpected{
+				Stdout: `
+stack "/stack3":
+	str = "stack3-string"
+`,
+			},
+		},
+		{
+			name: "three stacks with globals, wd = stacks",
```
katcipis, line 163: Maybe add a test case when wd has no stacks inside, like you have stacks with globals but wd is some internal dir that is empty (should be empty result etc)

katcipis (dismissed):
LGTM + suggestion for a test scenario

katcipis (approved):

PR #170, opened 2022-01-14
refactor: move e2e tests to specific package

katcipis:
As we are moving away from the previous cli package/cli function testing to full e2e the idea is to make this more explicit (+ we are thinking on merging code on main anyway, so the cli package may not exist soon). Not so sure about the name TBH. Maybe just "tests" and the e2e could be implicit, or just e2e and the testing part is implicit =P.


i4ki (approved):
lgtm

PR #173, opened 2022-01-18
fix: make absolute paths be relative to project root

i4ki:


cmd/terramate/e2etests/run_test.go
```diff
@@ -390,7 +465,7 @@ func TestRunOrderNotChangedStackIgnored(t *testing.T) {
 	stackMainTf := stack.CreateFile(mainTfFileName, "# some code")
 	stackConfig := hcl.NewConfig(terramate.DefaultVersionConstraint())
 	stackConfig.Stack = &hcl.Stack{
-		After: []string{stack2.Path()},
+		After: []string{project.RelPath(s.RootDir(), stack2.Path())},
```
katcipis, line 468: Hmm doesn't [stack2.RelPath()](https://github.com/mineiros-io/terramate/blob/c6a9617c61c504357e9b0287f57d5ced1e0aa7a4/test/sandbox/sandbox.go#L365) work here ?

katcipis (approved):
LGTM + tested on the current fail scenario and seems to fix + just a comment on how to get the relpath of stack on tests, but ignorable/possibly missing something.

cmd/terramate/e2etests/run_test.go
katcipis: I suspected that could be the case... hmm need to represent better all this path stuff x_x.

PR #177, opened 2022-01-19
refactor: use values instead of pointers on sandbox

katcipis:
Some objects on the sandbox were being handled as pointers but they seem like all the other objects, immutable/copiable. Also the method receivers are always values (because the objects are immutable). So I went with migrating to value so it is consitent/makes sense, or else it felt like we are being random about when it is pointer or value =P (I personally like immutability and guaranteeing that by copying, and the structs are quite small). But not a strong opinion, would also be OK with all pointers, but in this case it seems cool to make it explicit that no mutation is ever happening.


i4ki (approved):
lgtm

PR #179, opened 2022-01-20
refactor: make stack independent of the hcl config block

i4ki:


init.go
```diff
@@ -130,26 +130,21 @@ func Init(root, dir string, force bool) error {
 
 	logger.Debug().
 		Str("configFile", stackfile).
-		Msg("Create stack file.")
-	f, err := os.Create(stackfile)
+		Msg("Create new configuration.")
```
kassianh, line 133: We can probably reduce duplication of debug statements here by using a sub-logger when `Str("configFile", stackfile)` is first introduced a bit further up in the function. I missed this when adding logging.

stack/loader.go
```diff
@@ -105,9 +107,11 @@ func (l Loader) TryLoad(dir string) (stack S, found bool, err error) {
 	}
 
 	logger.Debug().
-		Msg("Set stack path and stack config.")
-	l.set(stackpath, cfg.Stack)
-	return l.stacks[stackpath], true, nil
+		Msg("Creates a new stack")
```
kassianh, line 110:
```suggestion
		Msg("Create a new stack.")
```

Just being nitpicky

kassianh (commented):
Few suggestions where I'm comfortable making them, otherwise, looks good to me!

hcl/hcl.go
```diff
@@ -85,11 +88,48 @@ const (
 	ErrStackInvalidRunOrder     errutil.Error = "invalid stack execution order definition"
 )
 
-func NewConfig(reqversion string) Config {
+// NewConfig creates a new HCL config with absdir as config directory path.
+func NewConfig(absdir string, reqversion string) (Config, error) {
```
katcipis, line 92: Is the invariant of the path being absolute guaranteed somewhere else ? Since it is checking if it exists here + if it is a dir I imagine we are missing a check if the path is absolute (type documents the path is absolute, but I could pass relative and it would work I think, but could be missing something stupid as usual).

hcl/hcl.go
```diff
@@ -85,11 +88,48 @@ const (
...
+
+// Save the configuration file using filename inside config directory.
+func (c Config) Save(filename string) error {
+	cfgpath := filepath.Join(c.absdir, filename)
+	f, err := os.Create(cfgpath)
+	if err != nil {
+		return fmt.Errorf("saving configuration file %q: %w", cfgpath, err)
+	}
+
+	defer f.Close()
```
katcipis, line 124: If the file fails to close, indicating that maybe not everything was written, it will fail silently (it is extremelly annoying to handle this on all Close calls...on services I used to have a closer that just logs errors and use it as a wrapper everywhere). Also I think linters can warn about this, golangci-lint default options are indeed way too nice =P.

hcl/hcl.go
```diff
@@ -214,7 +254,11 @@ func Parse(fname string, data []byte) (Config, error) {
 		)
 	}
 
-	var tmconfig Config
+	tmconfig, err := NewConfig(filepath.Dir(fname), "")
+	if err != nil {
+		return Config{}, fmt.Errorf("parsing HCL file %q: %w", fname, err)
```
katcipis, line 259:
```suggestion
		return Config{}, fmt.Errorf("creating config %q: %w", fname, err)
```

I think NewConfig doesn't to any parsing, but could have missed something :thinking:

init.go
```diff
@@ -101,22 +106,18 @@ func Init(root, dir string, force bool) error {
 	}
 
 	if isInitialized && !force {
-		logger.Trace().
-			Str("configFile", stackfile).
-			Msg("Stack is initialized annd not forced.")
+		logger.Trace().Msg("Stack is initialized annd not forced.")
```
katcipis, line 109:
```suggestion
		logger.Trace().Msg("Stack is initialized and not forced.")
```

init.go
```diff
@@ -128,28 +129,20 @@ func Init(root, dir string, force bool) error {
...
-	logger.Debug().
-		Str("configFile", stackfile).
-		Msg("Create stack file.")
-	f, err := os.Create(stackfile)
+	logger.Debug().Msg("Create new configuration.")
+
+	cfg, err := hcl.NewConfig(dir, DefaultVersionConstraint())
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to initialize stack: %w", err)
```
katcipis, line 136:
```suggestion
		return fmt.Errorf("failed to create new stack config: %w", err)
```

hcl/hcl.go
katcipis: I think we have other closes like this, so maybe we can address this on the future. But it should be addressed someday (the linter would help catch all of them).

stack/loader.go
```diff
@@ -44,45 +44,16 @@ func NewLoader(root string) Loader {
...
-		return s, nil
-	}
-
-	fname := filepath.Join(dir, config.Filename)
-
-	logger.Debug().
-		Str("configFile", fname).
-		Msg("Parse config file.")
-	cfg, err := hcl.ParseFile(fname)
+	stack, found, err := l.TryLoad(dir)
```
katcipis, line 47: Wondering if instead of TryLoad/Load we could just have Load and a NotExist error, like os.IsNotExist. But unrelated to the PR, just sharing.

stack/loader.go
```diff
@@ -109,13 +81,14 @@ func (l Loader) TryLoad(dir string) (stack S, found bool, err error) {
...
 	}
+
 	fname := filepath.Join(dir, config.Filename)
 
 	logger.Debug().
 		Str("configFile", fname).
 		Msg("Parse config file.")
-	cfg, err := hcl.ParseFile(fname)
 
+	cfg, err := hcl.ParseFile(fname)
```
katcipis, line 91: Noice

stack/loader.go
```diff
@@ -133,10 +106,11 @@ func (l Loader) TryLoad(dir string) (stack S, found bool, err error) {
 		return S{}, false, fmt.Errorf("stack %q is not a leaf stack", dir)
 	}
 
-	logger.Debug().
-		Msg("Set stack path and stack config.")
-	l.set(stackpath, cfg.Stack)
-	return l.stacks[stackpath], true, nil
+	logger.Debug().Msg("Create a new stack")
+
```
katcipis, line 110: Noice

stack/stack.go
```diff
@@ -40,26 +52,47 @@ type (
...
+func (s S) IsChanged() bool { return s.changed }
+
+// SetChanged sets the changed flag of the stack.
 func (s *S) SetChanged(b bool) { s.changed = b }
 
-func (s S) String() string {
-	return s.Name()
-}
+// String representation of the stack.
+func (s S) String() string { return s.Name() }
```
katcipis, line 95: hmm maybe add more info to the string representation ? Would be good to know if it is changed when printing a stack, maybe :thinking: (I can just print with +v/#v too...so not sure, just wondering).

katcipis (approved):
LGTM + the usual =P

hcl/hcl.go
i4ki: not sure how to properly validate that if both fs and prj paths are absolute. It'd help catch `some/dir` but we rarely use them. I think we should talk about this again, it seems wrong what we're doing like documenting the path must be absolute and doing checks everywhere... But yeah I will add the check here to be consistent with other places. Thanks

hcl/hcl.go
i4ki: yes this should be handled, I forgot to fix this.

hcl/hcl.go
i4ki: actually I think it should not be required to be an absolute path here because otherwise, it'll make it impossible to call `hcl.ParseFile("something.hcl")`. I will change the docs.

katcipis: Doesn't [filepath.IsAbs](https://pkg.go.dev/path/filepath#IsAbs) work here ? (regarding "not sure how to properly validate that if both fs and prj paths are absolute")

stack/stack.go
i4ki: not sure either, but if we format this here it's going to show everywhere. We have to talk about this I guess.

PR #180, opened 2022-01-20
refactor: make explicit the usage of fs and project's path

i4ki:


kassianh (approved):
lgtm!

cmd/terramate/cli/cli.go
```diff
@@ -468,13 +468,13 @@ func (c *cli) printStacks() {
...
 		stack := entry.Stack
-		stackRepr, ok := c.friendlyFmtDir(stack.Dir)
+		stackRepr, ok := c.friendlyFmtDir(stack.PrjAbsPath())
 		if !ok {
 			continue
 		}
 
 		logger.Debug().
-			Str("stack", c.wd()+stack.Dir).
+			Stringer("stack", stack).
```
katcipis, line 477: For some reason it felt like The Stringer would be a Batman villain or something :laughing: (Oh no, The Stringer !!!)

cmd/terramate/cli/cli.go
```diff
@@ -988,13 +988,13 @@ func (c *cli) filterStacksByWorkingDir(stacks []terramate.Entry) []terramate.Ent
 
 	logger.Trace().
 		Msg("Get relative working directory.")
-	relwd := prj.RelPath(c.root(), c.wd())
+	relwd := prj.PrjAbsPath(c.root(), c.wd())
```
katcipis, line 991: hmm it reads kinda off, relwd = PrjAbsPath, it is not an actual relpath

generate/generate.go
```diff
@@ -69,7 +68,7 @@ func Do(root string, workingDir string) error {
 		stack stack.S,
 		globals *terramate.Globals,
 	) error {
-		stackpath := project.AbsPath(root, stack.Dir)
+		stackpath := stack.AbsPath()
```
katcipis, line 71: Noice :100:

project/project.go
```diff
@@ -21,27 +21,27 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// RelPath returns the dir relative to project's root.
-func RelPath(root, dir string) string {
+// PrjAbsPath converts the file system absolute path absdir into an absolute
+// project path.
```
katcipis, line 25:
```suggestion
// PrjAbsPath converts the file system absolute path absdir into an absolute
// project path on the form /path/on/project relative to the given root.
```

project/project.go
```diff
@@ -21,27 +21,27 @@ import (
...
-		Str("action", "RelPath()").
-		Str("dir", dir).
+		Str("action", "PrjAbsPath()").
+		Str("dir", absdir).
 		Str("root", root).
 		Msg("Trim path to get relative dir.")
 
-	d := strings.TrimPrefix(dir, root)
-
+	d := strings.TrimPrefix(absdir, root)
```
katcipis, line 33: Wouldn't it be safer to use [filepath.Rel](https://pkg.go.dev/path/filepath#Rel) and then just prepend a filepath.Separator ? Right now if absdir is not relative to root it would just return absdir itself (nothing trimmed). Maybe this sort of invariant is guaranteed somewhere else.

run/run.go
```diff
@@ -44,15 +43,15 @@ func (c *Cmd) String() string {
 func (c Cmd) Run(root string, stacks []stack.S) error {
 	for _, stack := range stacks {
 		cmd := exec.Command(c.Path, c.Args...)
-		cmd.Dir = filepath.Join(root, stack.Dir)
+		cmd.Dir = stack.AbsPath()
```
katcipis, line 46: Noice :100:

katcipis (approved):
LGTM + some considerations that are non-blocking

cmd/terramate/cli/cli.go
i4ki: hahahahahaha

project/project.go
i4ki: Not sure if [filepath.Rel](https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/path/filepath/path.go;l=266) is what we want here, as it deals with lots of file system details like Windows volumes... and it also fails but in our case (if there's no bug) the `root` is always a substring of `absdir`. If we make this fail for a programming error, it'll clumsy the code and give strange error messages (user seeing an error that's not his fault). Maybe we can panic here if `absdir` is not really a superstring of `root` ?

katcipis: If we guarantee that on another level it is also OK to not check again, like it is the caller responsability to ensure that.

katcipis (approved):

PR #182, opened 2022-01-20
feat: implements the `wants` attribute for pulling other stacks.

i4ki:


manager.go
```diff
@@ -387,6 +387,40 @@ func (m *Manager) moduleChanged(
...
+		wanted = append(wanted, s)
+	}
+
+	visited := map[string]struct{}{}
+
+	for len(wantedBy) > 0 {
+		for _, s := range wantedBy {
+			ws, err := m.stackLoader.LoadAll(m.root, s.AbsPath(), s.Wants()...)
+			if err != nil {
+				return nil, err
```
katcipis, line 405:
```suggestion
				return nil, fmt.Errorf("calculating wanted stacks: %v", err)
```

manager.go
```diff
@@ -387,6 +387,40 @@ func (m *Manager) moduleChanged(
...
+	for _, s := range stacks {
+		wantedBy[s.PrjAbsPath()] = s
+		wanted = append(wanted, s)
+	}
+
+	visited := map[string]struct{}{}
+
+	for len(wantedBy) > 0 {
+		for _, s := range wantedBy {
+			ws, err := m.stackLoader.LoadAll(m.root, s.AbsPath(), s.Wants()...)
```
katcipis, line 403:
```suggestion
			wantedStacks, err := m.stackLoader.LoadAll(m.root, s.AbsPath(), s.Wants()...)
```

manager.go
```diff
@@ -387,6 +387,40 @@ func (m *Manager) moduleChanged(
...
+			if err != nil {
+				return nil, err
+			}
+
+			for _, wws := range ws {
```
katcipis, line 408:
```suggestion
			for _, wantedStack := range wantedStacks {
```

I know you like short names, but the ws/wws thing seems to have gotten too far IMHO :rofl:. But feel free to ignore.

stack/loader.go
```diff
@@ -103,25 +103,23 @@ func (l Loader) Set(dir string, s S) {
 	l.stacks[dir] = s
 }
 
-// LoadAll loads all the stacks in the dirs directories. If dirs are relative
-// paths, then basedir is used as base.
-func (l Loader) LoadAll(root string, basedir string, dirs ...string) ([]S, error) {
+// LoadAll loads all the stacks in the dirs directories. If dirs are project
+// paths, then absBaseDir is used as base.
+func (l Loader) LoadAll(root string, absBaseDir string, dirs ...string) ([]S, error) {
```
katcipis, line 108: I imagine there is good reason for this and Im missing something, but can't we work just with root + dirs is always projRelPaths ? or paths relative to root ? It is considerably complicated to my limited CPU that we have root, which a lot of stuff is relative to, and then another absBaseDir..that maybe things are going to be relative to ? and then paths...that may or may not be relative to absBaseDir ? (docs mention "if", so I imagine it is not always relative to absBaseDir). For example, Im wondering something like:

1 - "dirs" could be absolute ? (not very clear from docs)
2 - "dirs" could be relative to absBaseDir ? (clear from docs)
3 - "dirs" could be relative to root ? (not clear from docs)

Maybe when stacks are created/loaded they should normalize their after/before/wants in a single place so it is always project abs path relative to root (like /etc/etc), so any further logic can always count on that, or else this complexity (having /path/1 and ../../path/1, etc) spreads around, like here. I could be way off since you have been working more on this than me...Im just sharing that I got confused, but I do get confused rather easily :laughing:

katcipis: This is unrelated to the PR itself, just sharing the confusion/feelings.

test/sandbox/sandbox.go
```diff
@@ -152,6 +152,8 @@ func (s S) BuildTree(layout []string) {
 				cfg.Stack.After = specList(t, name, value)
 			case "before":
 				cfg.Stack.Before = specList(t, name, value)
+			case "wants":
+				cfg.Stack.Wants = specList(t, name, value)
```
katcipis, line 156: Unrelated to the PR, but thinknig that maybe specList should be parseListSpec ? (or parseStrList ?)

katcipis (dismissed):
LGTM :rocket: + non-blocking stuff / me being stupid / the usual

stack/loader.go
katcipis: It is kinda like we have 2 ways to represent the same thing and we are allowing that to spread, maybe in this case we can allow users to have 2 ways to reference the same thing but then internally we normalize to a single one. Just sharing some thoughts.

cmd/terramate/e2etests/run_test.go
```diff
@@ -441,11 +441,177 @@ stack-z
...
+	type testcase struct {
+		name   string
+		layout []string
+		wd     string
+		want   runExpected
+	}
+
+	for _, tc := range []testcase{
+		{
+			/* TODO(i4k): stack-a wants stack-a should error */
```
i4ki, line 459: should? @katcipis @mariux

stack/loader.go
i4ki: I improved the docs, let me know if it makes more sense now.

cmd/terramate/e2etests/run_test.go
```diff
@@ -441,11 +441,177 @@ stack-z
...
+			name: "stack-a wants stack-b (from inside stack-a)",
+			layout: []string{
+				`s:stack-a:wants=["/stack-b"]`,
+				`s:stack-b`,
+			},
+			wd: "/stack-a",
+			want: runExpected{
+				Stdout: "stack-a\nstack-b\n",
+			},
+		},
```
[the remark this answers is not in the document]
i4ki: done, thanks!

cmd/terramate/e2etests/run_test.go
katcipis: yeah I saw this and my first intuition  was that it should fail, it is kind odd for a stack to want itself. But on the other hand, if you want yourself the run set would be...yourself...and since you are not talking about order...seems OK, it is not like a cycle because since there is no ordering you are just adding things on a set...in this case adding yourself.

stack/loader.go
```diff
@@ -103,25 +103,24 @@ func (l Loader) Set(dir string, s S) {
 	l.stacks[dir] = s
 }
 
-// LoadAll loads all the stacks in the dirs directories. If dirs are relative
-// paths, then basedir is used as base.
-func (l Loader) LoadAll(root string, basedir string, dirs ...string) ([]S, error) {
+// LoadAll loads all the stacks in the dirs project paths. If dirs are relative,
+// then wd is used as working directory to compute their absolute paths.
+func (l Loader) LoadAll(root string, wd string, dirs ...string) ([]S, error) {
```
katcipis, line 108:
```suggestion
// LoadAll loads all the stacks in the dirs paths. For each dir in dirs: 
// - If it is relative, it will be considered relative to wd, path = wd + dir
// - If it is absolute, it will be considered absolute in relation to the given root, path = root + dir
func (l Loader) LoadAll(root string, wd string, dirs ...string) ([]S, error) {
```

katcipis: it did improve, thanks man. But was still confused about the role of root, so tried to make that clear also.

mariux (commented):
:heart: lgtm

cmd/terramate/e2etests/run_test.go
mariux: could be a warning.. as we allow cyclic references this is not a hard fail?

i4ki: ok, added a warning.

katcipis (approved):

PR #185, opened 2022-01-21
chore: improve git wrapper isolation

katcipis:
The current implementation make sit easy to inherit env from parent process, making easy for changes on environment to change
behavior (we already have problems with git hook env leaking to tests and breaking them). A design that makes sharing explicit seems more proper here.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -38,6 +38,9 @@ type (
 		WorkingDir string
 
 		// Env is the environment variables to be passed over to git.
+		// If it is nil it means no environment variables should be passed.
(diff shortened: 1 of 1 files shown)


i4ki (approved):
lgtm

i4ki:
was this really a chore? :thinking:

katcipis:
> was this really a chore? thinking

hmm it is not a feature I think because it didn't affect the cli itself, not a refactoring because it changes behavior, so I just went with chore..but TBH IDK :laughing:

PR #187, opened 2022-01-24
feat: integrate generate config loading with code generation

katcipis:
This integrates the config loading logic on code generation + outdated code checking. Here is a example script on how it looks like with a config created on the root of the project:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
    generate {
      locals_filename = "locals.tf"
      backend_config_filename = "backend.tf"
    }
  }
  backend "local" {
    whatever = terramate.path
  }
}

export_as_locals {
  path = terramate.path
}
EOM

mkdir -p stacks/stack-{1,2,3}
terramate stacks init stacks/stack-{1,2,3}

echo "initial tree"
tree

echo
terramate generate

echo "tree after generating code"
tree
```

When the filenames are changed the old ones will be ignored on code generation + outdated detection. So it is the responsibility of the dev to delete old generated files (we could add something to help with that also, but not in place right now).

Will create follow up tasks for automation that can help detect/cleanup automatically files dont belong to any configuration anymore (old/outdated files).


i4ki (approved):
LGTM

PR #188, opened 2022-01-25
docs: initial spec for terraform code generation

katcipis:
Started with the basics so we can discuss basic syntax/behavior/constraints. Still need to add definitions on overriding/hierarchical behavior, for now there is only a single example defining things on root and generating code for all stacks.

### DIFF
--- a/docs/terraform-generation.md
+++ b/docs/terraform-generation.md
@@ -23,7 +23,7 @@ block. For now, only the following is not allowed:
 * References to locals on the form `local.name`
 
 Basically there is no support for partial evaluation (yet), so anything defined
-needs to be evaluated on the context of the code generation and the final generated
+needs to be evaluated in the context of the code generation, and the final generated
 code will have the results of the evaluation.
 
 Each `export_as_terraform` block requires a label. This label is part of the identity
--- a/docs/terraform-generation.md
+++ b/docs/terraform-generation.md
@@ -1,7 +1,7 @@
 # Terraform Code Generation
 
 Terramate provides a way to integrate its [globals](globals.md) and
-[metadata](metadata.md) on Terraform code by allowing you to generate
+[metadata](metadata.md) in Terraform code by allowing you to generate
 arbitrary terraform code that leverages Terramate data.
 The generated code can then be composed/referenced by any Terraform code
 inside a stack.
--- a/docs/terraform-generation.md
+++ b/docs/terraform-generation.md
@@ -67,7 +67,7 @@ backend "local" {
 ```
 
 To generate provider/Terraform configuration for all stacks we can add
-on the root configuration:
+in the root configuration:
 
 ```hcl
 export_as_terraform "provider" {
--- a/docs/terraform-generation.md
+++ b/docs/terraform-generation.md
@@ -1,0 +1,118 @@
+# Terraform Code Generation
+
+Terramate provides a way to integrate its [globals](globals.md) and
+[metadata](metadata.md) on Terraform code by allowing you to generate
+arbitrary terraform code that leverages Terramate data.
+The generated code can then be composed/referenced by any Terraform code
+inside a stack.
+
+This feature is a superset (and generalization) of 
+[locals generation](locals-generation.md) and
+[backend config generation](backend-config.md).
+
+To make use of [globals](globals.md) and [metadata](metadata.md) define
+a `export_as_terraform` block on a [Terramate configuration file](config.md)
+defining the code you want to generate. The code may include:
+
+* Blocks, sub blocks,etc 
+* Attributes initialized by literals
+* Global references
+* Metadata references
+* Expressions using interpolation, functions, etc
+
+Mostly of what you can do on Terraform can be defined here, for now the
+following is disallowed:
+
+* References to variables on the form `var.name`
+* References to locals on the form `local.name`
+
+Basically there is no support for partial evaluation (yet), so anything defined
+needs to be evaluated on the context of the code generation.
+
+Each `export_as_terraform` requires a label. This label is part of the identity
+of the block and is also used as a default to which filename will be used when
+code is generated. Given a label `x` the filename will be `x.tf`. The labels are
+also used to configure a different behavior for the filenames, as can be seen
+[here](todo-docs-for-config).
+
+Now lets jump to some examples.
+
+Given these globals defined on the root of the project:
+
+```hcl
+globals {
+  backend_data = "backend_data"
+  provider_data = "provider_data"
+  provider_version = "0.6.6"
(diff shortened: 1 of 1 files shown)


docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
+# Terraform Code Generation
+
+Terramate provides a way to integrate its [globals](globals.md) and
+[metadata](metadata.md) on Terraform code by allowing you to generate
```
kassianh, line 4:
```suggestion
[metadata](metadata.md) in Terraform code by allowing you to generate
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+arbitrary terraform code that leverages Terramate data.
```
kassianh, line 5:
```suggestion
arbitrary Terraform code that leverages Terramate data.
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+The generated code can then be composed/referenced by any Terraform code
+inside a stack.
+
+Terraform code generation starts with the definition of a `export_as_terraform`
+block on a [Terramate configuration file](config.md) defining the code you
```
kassianh, line 10:
```suggestion
block in a [Terramate configuration file](config.md) defining the code you
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+want to generate inside the block. The code may include:
+
+* Blocks, sub blocks,etc
```
kassianh, line 13:
```suggestion
* Blocks, sub blocks, etc 
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+block on a [Terramate configuration file](config.md) defining the code you
+want to generate inside the block. The code may include:
+
+* Blocks, sub blocks,etc 
+* Attributes initialized by literals
+* Terramate Global references
+* Terramate Metadata references
+* Expressions using interpolation, functions, etc
+
+Mostly of what you can do on Terraform can be done on a `export_as_terraform`
```
kassianh, line 19:
```suggestion
Most of what you can do in Terraform can be done in a `export_as_terraform`
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+block, for now only the following is disallowed:
```
kassianh, line 20:
```suggestion
block. For now, only the following is not allowed:
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+
+* References to variables on the form `var.name`
+* References to locals on the form `local.name`
+
+Basically there is no support for partial evaluation (yet), so anything defined
+needs to be evaluated on the context of the code generation and the final generated
```
kassianh, line 26:
```suggestion
needs to be evaluated in the context of the code generation and the final generated
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+code will have the results of the evaluation.
+
+Each `export_as_terraform` block requires a label. This label is part of the identity
+of the block and is also used as a default to which filename will be used when
```
kassianh, line 30:
```suggestion
of the block and is also used as a default for which filename will be used when
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+code is generated. Given a label `x` the filename will be `_gen_terramate_x.tf`. The labels are
+also used to configure different filenames for each block if the default names are
+undesired, more details on how to configure this can be checked [here](todo-docs-for-config).
```
kassianh, line 33:
```suggestion
undesired. More details on how to configure this can be checked [here](todo-docs-for-config).
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+globals {
+  backend_data = "backend_data"
+  provider_data = "provider_data"
+  provider_version = "0.6.6"
+  terraform_version = "1.1.3"
+}
+```
+
+We can define the generation of a backend configuration for all
+stacks by defining a `export_as_terraform` blocks on the root
```
kassianh, line 50:
```suggestion
stacks by defining a `export_as_terraform` block in the root
```

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
...
+Which will generate code for all stacks using the filename `_gen_terramate_backend.tf`:
+
+```hcl
+backend "local" {
+  param = "backend_data"
+}
+```
+
+To generate provider/Terraform configuration for all stacks we can add
+on the root configuration:
```
kassianh, line 70:
```suggestion
in the root configuration:
```

kassianh (commented):
Lgtm. Just left a few suggestions.

docs/terraform-generation.md
```diff
@@ -0,0 +1,115 @@
+# Terraform Code Generation
+
+Terramate provides a way to integrate its [globals](globals.md) and
```
i4ki, line 3: This sentence is a bit confusing, maybe because it says `Terramate ... Terraform ... Terramate ...`.

What about:

```
The tool supports the generation of arbitrary Terraform code using 
both [globals](globals.md) and [metadata](metadata.md) from Terramate.
```

but for me reading this, it looks like `globals` and `metadata` is already implicit in the Terramate context, so maybe this is enough:

```
Terramate supports the generation of arbitrary Terraform code using 
both [globals](globals.md) and [metadata](metadata.md).
```

i4ki (changes requested):
The spec looks good but I think we should clarify the generation strategy. About the output code, the spec only says: `and the final generated code will have the results of the evaluation`.

I think it should be made clear that the final output code will contain only literals (number, string, etc), so even if the code uses TF functions to manipulate the variables, they will not be in the final code. 

So, if user provides `file(globals.filename_used_everywhere)`, the output will contains the file content, which means changing the file on disk WILL NOT trigger a IaC change anymore.

katcipis:
> The spec looks good but I think we should clarify the generation strategy. About the output code, the spec only says: `and the final generated code will have the results of the evaluation`.
> 
> I think it should be made clear that the final output code will contain only literals (number, string, etc), so even if the code uses TF functions to manipulate the variables, they will not be in the final code.
> 
> So, if user provides `file(globals.filename_used_everywhere)`, the output will contains the file content, which means changing the file on disk WILL NOT trigger a IaC change anymore.

Gave it a shot at improving it, take a look now. Thanks for the feedback :heart:

i4ki (approved):
lgtm

PR #189, opened 2022-01-26
docs: add hierarchical behavior to tf code gen spec

katcipis:
Add some basic on how it will behave with multiple configs on different levels inside the project.


docs/terraform-generation.md
```diff
@@ -114,4 +114,111 @@ terraform {
 }
 ```
 
-TODO: Define overriding behavior (hierarchies).
+## Hierarchical Code Generation
+
+Terraform code generation can be defined anywhere inside a project, from a specific
+stack, which defines code generation only for the specific stack, to parent dirs
+or even the project root, which then has the potential to affect code generation
+to multiple or all stacks (as seen on the previous example).
```
kassianh, line 122:
```suggestion
to multiple or all stacks (as seen in the previous example).
```

docs/terraform-generation.md
```diff
@@ -114,4 +114,111 @@ terraform {
...
+
+This does raise the question of how code generation configuration is merged/overridden
+as Terramate navigate the project loading and evaluating configuration in order to
```
kassianh, line 125:
```suggestion
as Terramate navigates the project loading and evaluating configuration in order to
```

docs/terraform-generation.md
```diff
@@ -114,4 +114,111 @@ terraform {
...
+
+```hcl
+terraform {
+  required_version = "overriden"
+}
+```
+
+Since the `stacks/stack-1` configuration is overriding the previous
+definition at `stacks`. Any other stack under `stacks` would remain
+with the configuration defined on the parent dir `stacks`.
```
kassianh, line 193:
```suggestion
with the configuration defined in the parent dir `stacks`.
```

kassianh (commented):
Just a few minor changes, LGTM!

docs/terraform-generation.md
katcipis: I have problems with on/in :laughing:

kassianh (approved):

PR #190, opened 2022-01-27
feat: add core logic to generate terraform code per stack

katcipis:
Not yet integrated on code generation itself, but integrating on the next step should be reasonably easy. This PR add a function that for a single stack:

* Reads configs
* Merge them, overriding when necessary
* Evaluate the final result with globals and metadata

Which is the core building block for the overall code generation for N stacks.

Specs for the feature can be found here:

* https://github.com/mineiros-io/terramate/pull/188
* https://github.com/mineiros-io/terramate/pull/189


kassianh (approved):
lgtm

generate/exportedtf/exportedtf.go
```diff
@@ -0,0 +1,216 @@
...
+// StackTf represents all exported terraform code for a stack,
+// mapping the exported code name to the actual Terraform code.
+type StackTf struct {
+	tfcode map[string]Body
+}
+
+// Body represents exported Terraform code from a single block.
+// Is contains parsed and evaluated code on it.
+type Body struct {
+	body []byte
```
thiesen, line 43: what about calling this field `content`?

generate/exportedtf/exportedtf.go
```diff
@@ -0,0 +1,216 @@
...
+	logger.Trace().Msg("loading export_as_terraform blocks.")
+
+	exportBlocks, err := loadExportBlocks(rootdir, stackpath)
+	if err != nil {
+		return StackTf{}, fmt.Errorf("loading exported terraform code: %w", err)
+	}
+
+	evalctx, err := newEvalCtx(stackpath, sm, globals)
+	if err != nil {
+		return StackTf{}, fmt.Errorf("preparing to eval exported terraform code: %v", err)
```
thiesen, line 96: why are we using `%v` instead of `%w` as in line 91?

generate/exportedtf/exportedtf.go
katcipis: The type itself is already Body and this is just the private part of the body itself, so I don't see the point of calling each a different name :thinking: . I would then rename both to Content, but Body is shorter and it seems clear enough...or do you think body is confusing ? (I wrapped the []byte only to make it immutable, could also just pass around []byte).

generate/exportedtf/exportedtf.go
katcipis: You use %w for wrapping when you intend to expose the error for further inspection and use %v or %s if you want the error to be opaque and conceil details (only the error str representation will be added on the new error, but no further type info). More details can be found [here](https://go.dev/blog/go1.13-errors).

Overall there is several different ways to model errors so they can be inspected for details, sentinels are the simplest ones with some tradeoffs on coupling or even leaking details (as mentioned on the blog post). So I try to use wrapping/sentinels only when I want to expose a clear distinction on the nature of a failure. For the creation of the eval ctx I didn't thought it would be useful, but we could add a sentinel if you think it makes sense (like ErrEval).

katcipis: Ah I also add Sentinels usually as I test, since I use then on the tests to check for specific errors. I didn't tested eval ctx creation so far, so that is also a reason it doesn;t have a wrapping. My default is to not wrap/use %v, so I just add the wrap when Im intentionally wanting to expose/test something through the API.

generate/exportedtf/exportedtf.go
katcipis: Tried to improve the names, not extremely happy with them overall

thiesen (approved):
lgtm!

I'm gonna steal some of this stuff eventually :grimacing:

katcipis:
> lgtm!
> 
> I'm gonna steal some of this stuff eventually grimacing

We can come up with better stuff together, none of this is very "reference" x_x (usual feelings of I hate my code)

kassianh (approved):

PR #193, opened 2022-01-31
feat: terramate generate: add overall tf code generation

katcipis:
Introduces generate_hcl code generation on the cli, including outdated code detection. Here is an example to showcase how to use + expected results:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}

globals {
    tf_version = "1.1.3"
    gh_provider_version = "4.19.1"
}

generate_hcl "provider.tf" {

    provider "github" {
      owner = global.gh_owner
    }

    terraform {
      required_providers {
        github = {
          source  = "integrations/github"
          version = global.gh_provider_version
        }
      }
    }

    terraform {
      required_version = global.tf_version
    }
}

generate_hcl "main.tf" {
  module "repositories" {
    source = "../../../modules/terramate-github-repositories"

    repositories        = global.repositories
    repository_defaults = global.repository_defaults
    member_map          = global.member_map
  }
}
EOM

# Now add some per stack globals

cat > stacks/stack-1/terramate.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-1"
            visibility = "private"
        }
    ]
    repository_defaults = {
        config = "stack-1"
    }
    member_map  = {
        "first.lastname" = "powerpuffGirls69"
    }
}
stack {}
EOM

cat > stacks/stack-2/terramate.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-2"
            visibility = "public"
        }
    ]
    repository_defaults = {
        config = "stack-2"
    }
    member_map  = {
        "first.lastname" = "onePunchMan666"
    }
}
stack {}
EOM


echo "initial tree"
tree

echo
terramate generate

echo "tree after generating code"
tree

echo
echo "checking generated main.tf files"
terramate run -- cat main.tf

echo
echo "checking generated provider.tf files"
terramate run -- cat provider.tf
```


kassianh (commented):
Lgtm but also a bit over my head.

generate/generate.go
```diff
@@ -129,47 +132,198 @@ func CheckStack(root string, stack stack.S) ([]string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Trace().Msg("Loaded exported terraform code, checking")
+
+	outdated := []string{}
+
+	for name, tf := range loadedStackTf.ExportedCode() {
+		exportedTfFilename := name
```
thiesen, line 221: non-blocking: would it make sense to rename `name` to `exportedTfFilename` in the loop so we don't need to assign a new variable here?

generate/generate.go
```diff
@@ -129,47 +132,198 @@ func CheckStack(root string, stack stack.S) ([]string, error) {
...
+		logger.Trace().Msg("checking if code is updated.")
+
+		tfcode := PrependHeader(tf.String())
+		currentTfCode, err := loadGeneratedCode(targetpath)
+		if err != nil {
+			return nil, err
+		}
+
+		if tfcode != string(currentTfCode) {
+			logger.Trace().Msg("Outdated code detected.")
```
thiesen, line 237: non-blocking:

```suggestion
			logger.Trace().Msg("Outdated terraform code detected.")
```

generate/generate_check_test.go
```diff
@@ -23,7 +23,95 @@ import (
...
 )
 
-func TestCheckReturnsOutdatedStackFilenames(t *testing.T) {
+func TestCheckReturnsOutdatedStackFilenamesForExportedTf(t *testing.T) {
+	s := sandbox.New(t)
+
+	stackEntry := s.CreateStack("stacks/stack")
+	stack := stackEntry.Load()
+
+	assertOutdated := func(want []string) {
```
thiesen, line 32: non-blocking: maybe `assertOutdatedStacks` would make clearer what we're asserting?

generate/generate_hcl_test.go
```diff
@@ -14,62 +14,339 @@
...
+				path := filepath.Join(s.RootDir(), cfg.path)
+				test.AppendFile(t, path, config.Filename, cfg.add.String())
+			}
 
-func labels(labels ...string) hclwrite.BlockBuilder {
-	return hclwrite.Labels(labels...)
-}
+			workingDir := filepath.Join(s.RootDir(), tcase.workingDir)
+			err := generate.Do(s.RootDir(), workingDir)
+			assert.IsError(t, err, tcase.wantErr)
```
thiesen, line 240: are we actually setting `wantErr` in any of the test cases? if this is not the case, shouldn't this assertion fail?

generate/genhcl/genhcl.go
```diff
@@ -31,24 +31,24 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// StackHCLs represents all exported terraform code for a stack,
-// mapping the exported code name to the actual Terraform code.
+// StackHCLs represents all generated HCL code for a stack,
+// mapping the generated code name to the actual HCL code.
```
thiesen, line 35: `mapping the generated code name` are we referring to the actual code content (like a blok name) or is it the filename?

thiesen (approved):
dude, looks great to me! :mage_man: 

I added some questions but nothing should be blocking - just https://github.com/mineiros-io/terramate/pull/193/files#r798043896 but I assume I'm missing something since the tests are green)

generate/generate.go
katcipis: Indeed we can, this was a left over from a previous version of the code that used the name to pass to a function, I removed the function but I was refactoring quickly, so I just removed the function call..but now it doesn't make sense :laughing: . Nice catch :heart:

generate/generate.go
katcipis: Used HCL since it is not about terraform anymore (but the code may be confusing because Im still renaming things to match generate_hcl instead of export_as_terraform).

generate/generate_check_test.go
katcipis: since the function is defined inside the test (scope is quite small) I thought a smaller name was OK, but will think on a better name if it was confusing.

generate/generate_hcl_test.go
katcipis: If we never set wantErr, meaning that we are not expecting a failure, and no failure happens, this is the same as:

```go
assert.IsError(t, nil, nil)
```

Which works, because nil == nil (they match). If we get an unexpected error, it fails, and if we define wantErr then it would fail if the got err doesn't match.

generate/genhcl/genhcl.go
katcipis: It is the filename, will improve this

generate/generate_hcl_test.go
katcipis: But since we are not using it here we can remove it, I usually have some error handling tests but so far haven't added at this level. But here you can see how it works:

https://github.com/mineiros-io/terramate/blob/ad148a17de74f6f7b6e4594c69188f920f265483/generate/genhcl/genhcl_test.go#L577

When you set the error, it will check for a specific match with the sentinel wrapped inside the error, when you don't set it it will check if both errors are nil, since only a nil error "is" another nil error.

katcipis: Simpler example: https://go.dev/play/p/aMat5TWAU7t

katcipis:
> dude, looks great to me! mage_man
> 
> I added some questions but nothing should be blocking - just https://github.com/mineiros-io/terramate/pull/193/files#r798043896 but I assume I'm missing something since the tests are green)

Thanks for the review man :heart: 

Tried to explain and provide some examples about errors.Is, the assert.IsError is a wrapper around errors.Is, so they behave the same way.

Also improved the code with all your suggestions, thanks again :heart:

generate/generate_hcl_test.go
thiesen: got it! thanks for taking the time to explain! :heart:

thiesen (approved):
lgtm! :rocket:

PR #194, opened 2022-01-31
refactor: split code generation tests

katcipis:
Splitting the backend generation from locals generation. Since we are also adding terraform code generation it was getting quite big. It would also make it easy to remove older generation methods that get deprecated by terraform code gen.


thiesen (approved):
lgtm!

PR #199, opened 2022-02-03
feat: add limited scope traversal expr on generate_hcl

katcipis:
We are not going to support function calls or loops for now (they are allowed when evaluation is in play, but not on the expression copying for now). We also dont provide support for traversals inside objects, only directly on attributes inside blocks.

Current behavior:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}

globals {
}

generate_hcl "provider.tf" {

    provider "github" {
      owner = global.gh_owner
      some_local = local.should_work
      something = traversing.is.fun
    }
}
EOM

terramate generate

echo
echo "checking generated provider.tf files"
terramate run -- cat provider.tf
```

Produces:

```
generating code for all stacks

checking generated provider.tf files
2022-02-03T17:53:01+01:00 INF Running command in stacks reachable from working directory action=runOnStacks() changed=false workingDir=/tmp/tmp.8ne5q8416u
2022-02-03T17:53:01+01:00 INF Running command in stack cmd="cat provider.tf" stack=stack-1
// TERRAMATE: GENERATED AUTOMATICALLY DO NOT EDIT
// TERRAMATE: originated from generate_hcl block on /stacks/terramate.tm.hcl

provider "github" {
  owner      = "the-awesome-mineiros"
  some_local = local.should_work
  something  = traversing.is.fun
}
2022-02-03T17:53:01+01:00 INF Running command in stack cmd="cat provider.tf" stack=stack-2
// TERRAMATE: GENERATED AUTOMATICALLY DO NOT EDIT
// TERRAMATE: originated from generate_hcl block on /stacks/terramate.tm.hcl

provider "github" {
  owner      = "the-awesome-mineiros"
  some_local = local.should_work
  something  = traversing.is.fun
}
```


hcl/hcl.go
```diff
@@ -514,6 +532,27 @@ func CopyBody(target *hclwrite.Body, src *hclsyntax.Body, evalctx *eval.Context)
 	return nil
 }
 
+func getUnknownScopeTraversal(expr hclsyntax.Expression, evalctx *eval.Context) (*hclsyntax.ScopeTraversalExpr, bool) {
+	logger := log.With().
+		Str("action", "parseUnknownTraversal").
```
thiesen, line 537: non-blocking: I believe it'd easier to log the exact name of the function as you're already doing in other cases 

```suggestion
		Str("action", "getUnknownScopeTraversal").
```

thiesen (approved):
:rocket:

hcl/hcl.go
katcipis: You are right, this happened because when I refactor using gorename it doesn't fix things like the log entries :laughing: . Thanks for catching this.

katcipis: We are not consistent about it, but I like to add <package>.<func> for functions, since just the function is easier to get clashes (different packages, same function names).

thiesen (approved):

PR #201, opened 2022-02-03
fix: avoid aborting when we cant get branch name

katcipis:
On some CI envs we can't get the branch name easily, not sure if we really need them TBH but since we want to get this working ASAP I tried to keep the behavior the same when we have the branch name and then on CI the check won't be performed for now. With more time we can improve/review the overall git checks.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -1165,11 +1165,22 @@ func (p *project) setDefaults(parsedArgs *cliSpec) error {
 				Msg("Get current branch.")
 			branch, err := gw.CurrentBranch()
 			if err != nil {
-				return fmt.Errorf("failed to get current git branch: %v", err)
-			}
-
-			if branch == gitOpt.DefaultBranch {
-				baseRef = gitOpt.DefaultBranchBaseRef
+				// ON CI envs we don't have a clean way to get the branch name
+				// git symbolic-ref will just fail. but we don't want a hard
+				// fail on this case. We need to re-assess this, if we really
+				// need the branch name, is there some other way to solve this ? etc.
+				//
+				// More info on git branch names on GHA:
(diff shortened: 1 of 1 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -897,51 +897,60 @@ func (c *cli) checkLocalDefaultIsUpdated(g *git.Git) error {
...
+	}
+
+	c.logerr("current branch %q is the default branch, checking if it is updated.", branch)
+	c.logerr("retrieving info from remote branch: %s/%s ...", defaultRemote, defaultBranch)
+
+	logger.Trace().
+		Msg("Fetch remote reference.")
+	remoteRef, err := g.FetchRemoteRev(defaultRemote, defaultBranch)
+	if err != nil {
+		return fmt.Errorf("checking local branch %q is update: %v", branch, err)
```
thiesen, line 929:
```suggestion
		return fmt.Errorf("checking if local branch %q is updated: %v", branch, err)
```

cmd/terramate/cli/cli.go
```diff
@@ -897,51 +897,60 @@ func (c *cli) checkLocalDefaultIsUpdated(g *git.Git) error {
...
+	}
+	c.logerr("retrieved info from remote branch: %s/%s.", defaultRemote, defaultBranch)
+
+	logger.Trace().
+		Msg("Get local commit ID.")
+	localCommitID, err := g.RevParse(branch)
+	if err != nil {
+		return fmt.Errorf("checking local branch %q is update: %v", branch, err)
```
thiesen, line 937:
```suggestion
		return fmt.Errorf("checking if local branch %q is updated: %v", branch, err)
```

thiesen (approved):
LGTM!

PR #202, opened 2022-02-04
feat: terramate generate: cleanup old files

katcipis:
When we generate code now any old code is automatically removed, so if you change configuration file names it should create the config with the new names and remove the old ones. Changed the design to generate all code for all configs first and then removing old generated files, and then saving the generated code.

Example of change that would leave an outdated config file, but now is automatically removed:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}

globals {
}

generate_hcl "provider.tf" {
    provider "github" {
      owner = global.gh_owner
      some_local = local.should_work
      something = traversing.is.fun
    }
}
EOM

terramate generate

echo "project tree"
tree

globals {
}

generate_hcl "new_provider.tf" {
    provider "github" {
      owner = global.gh_owner
      some_local = local.should_work
      something = traversing.is.fun
    }
}
EOM

terramate generate

echo "project tree"
tree
```


kassianh (approved):
lgtm :) but again a bit over my head

thiesen (approved):
lgtm!

PR #204, opened 2022-02-07
fix: use default logging before parsing cli args

katcipis:
Now logging done before configuring the log library will have the same default level/format as the ones configured for the cli options. Now we get this:

```
terramate wrong
2022-02-07T18:54:57+01:00 FTL failed to parse cli args: [wrong] error="unexpected argument wrong" action=newCli()
```

Instead of JSON (which is not the default).


kassianh (approved):
lgtm

PR #205, opened 2022-02-07
fix: outdated code detection wont ignore empty generate_hcl

katcipis:
The following snippet should work, but before this fix it doesn't:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}

generate_hcl "empty.tf" {}
EOM

terramate run -- echo hi
```

In this scenario there is no way to solve the problem, calling terramate generate won't generate any code and yet the outdated code detection keeps detecting the project as outdated. This PRs fixes this so detection properly ignores when the generated code is empty (empty block) + file does not exist on fs (it shouldn't). 

There is less testing on export_as_locals and backend config because we are converging to replacing them with generate_terraform (backend config is already replaced by generate_hcl, but not export_as_locals), but the fix itself is also applied on them.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -611,7 +611,7 @@ func checkFileCanBeOverwritten(path string) error {
 }
 
 // loadGeneratedCode will load the generated code at the given path.
-// It returns an error if it can't read the file of if the file is not
+// It returns an error if it can't read the file or if the file is not
 // a Terramate generated file.
 //
 // The returned boolean indicates if the file exists, so the contents of
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -286,7 +286,7 @@ func backendConfigOutdatedFiles(
 
 	logger.Trace().Msg("Checking for outdated backend cfg code on stack.")
 
-	if string(genbackend) != string(currentbackend) {
+	if string(genbackend) != currentbackend {
 		logger.Trace().Msg("Detected outdated backend cfg.")
 		return []string{cfg.BackendCfgFilename}, nil
 	}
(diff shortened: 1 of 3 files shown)


generate/generate.go
```diff
@@ -594,11 +606,18 @@ func writeGeneratedCode(target string, code string) error {
...
 
-func loadGeneratedCode(path string) ([]byte, error) {
+// loadGeneratedCode will load the generated code at the given path.
+// It returns an error if it can't read the file of if the file is not
+// a Terramate generated file.
+//
+// The returned boolean indicates if the file exists, so the contents of
+// the file + true is returned if a file is found, but if no file is found
+// it will return an empty string and false indicating that the file doesn't exist.
+func loadGeneratedCode(path string) (string, bool, error) {
```
katcipis, line 620: Could just use the empty string as also meaning "file dont exist" since we never generate empty files on terramate + empty files wont be detected as being generated by terramate. But I usually dislike overloading the meaning of something ("" means file exists and is empty AND file doesn't exist)...so this seemed better even though not strictly necessary.

kassianh (approved):
lgtm

generate/generate.go
```diff
@@ -307,33 +311,37 @@ func generatedHCLOutdatedFiles(
...
 		logger.Trace().Msg("Checking if code is updated.")
-
-		tfcode := prependGenHCLHeader(genHCL.Origin(), genHCL.String())
-		currentTfCode, err := loadGeneratedCode(targetpath)
+		currentHCLcode, codeFound, err := loadGeneratedCode(targetpath)
 		if err != nil {
 			return nil, err
 		}
+		if !codeFound && genHCL.String() == "" {
+			logger.Trace().Msg("Updated since file not found and generated_hcl is empty")
```
thiesen, line 338: shouldn't it be **not updated** instead?

thiesen: ahhh! I got it - it means that the file is not outdated, right? sorry for the noise

generate/generate.go
```diff
@@ -594,11 +606,18 @@ func writeGeneratedCode(target string, code string) error {
 }
 
 func checkFileCanBeOverwritten(path string) error {
```
thiesen, line 608: non-blocking (just a question):  wouldn't it make sense to return a bool from `checkFileCanBeOverwritten` instead of an error?

generate/generate.go
```diff
@@ -594,11 +606,18 @@ func writeGeneratedCode(target string, code string) error {
...
-	_, err := loadGeneratedCode(path)
+	_, _, err := loadGeneratedCode(path)
 	return err
 }
 
-func loadGeneratedCode(path string) ([]byte, error) {
+// loadGeneratedCode will load the generated code at the given path.
+// It returns an error if it can't read the file of if the file is not
```
thiesen, line 614:
```suggestion
// It returns an error if it can't read the file or if the file is not
```

thiesen (approved):
lgtm!

PR #206, opened 2022-02-07
draft

katcipis:


.github/workflows/ci.yml
```diff
@@ -62,6 +62,20 @@ jobs:
       - name: make build
         run: make build
 
+  release_dry_run:
```
katcipis, line 65: We were doing the release dry run also on mac, which is unnecessary, so I already moved this to do the dry run only once.

soerenmartius (approved):
very good stuff! LGTM!

PR #207, opened 2022-02-08
fix: terramate generate: different code gen strategies with same config filename overwrite each other

katcipis:
Now this:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack
terramate stacks init stacks/stack

terramate {
  config {
    generate {
      locals_filename = "file.tf"
    }
  }
}
generate_hcl "file.tf" {
    block {}
}
export_as_locals {
    something = "hi"
}
EOM

terramate generate
```

Will fail like this:

```
2022-02-08T16:11:25+01:00 FTL generating code. error="failed to generate code: conflicting config detected: two configurations produce same file \"file.tf\""
```

Instead of overwriting each other. The error message can be improved on the future with information about the origin of the config.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -895,6 +895,11 @@ func parseGenerateConfig(block *hclsyntax.Block) (GenerateConfig, error) {
 		}
 	}
 
+	if cfg.LocalsFilename != "" && cfg.LocalsFilename == cfg.BackendCfgFilename {
+		return GenerateConfig{}, fmt.Errorf(
+			"terramate.config.generate: locals and backend cfg filenames have the same %q value", cfg.LocalsFilename)
+	}
+
 	return cfg, nil
 }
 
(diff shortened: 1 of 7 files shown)


hcl/hcl.go
```diff
@@ -895,6 +895,11 @@ func parseGenerateConfig(block *hclsyntax.Block) (GenerateConfig, error) {
 		}
 	}
 
+	if cfg.LocalsFilename != "" && cfg.LocalsFilename == cfg.BackendCfgFilename {
+		return GenerateConfig{}, fmt.Errorf(
+			"terramate.config.generate: locals and backend cfg filenames have the same %q value", cfg.LocalsFilename)
```
thiesen, line 900:
```suggestion
			"terramate.config.generate: locals and backend cfg files have the same name %q", cfg.LocalsFilename)
```

PR #208, opened 2022-02-08
fix: outdated detection: detect stale files from previous configurations

katcipis:
Right now when a config is removed, the file left there is not detected as outdated, since we only check if code matches with what is defined on the current configs. Now we list all generated code and check if any of the generated code existing on the filesystem has no config matching with it, making it outdated too.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -228,26 +228,56 @@ func CheckStack(root string, stack stack.S) ([]string, error) {
 		return nil, fmt.Errorf("checking for outdated code: %v", err)
 	}
 
+	logger.Trace().Msg("Listing current generated files.")
+
+	g, err := ListStackGenFiles(stack)
+	if err != nil {
+		return nil, fmt.Errorf("checking for outdated code: %v", err)
+	}
+	currentFiles := newStringSet(g...)
+
 	stackpath := stack.AbsPath()
 	stackMeta := stack.Meta()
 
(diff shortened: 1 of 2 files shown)


generate/generate.go
```diff
@@ -228,26 +228,56 @@ func CheckStack(root string, stack stack.S) ([]string, error) {
 		return nil, fmt.Errorf("checking for outdated code: %v", err)
 	}
 
+	logger.Trace().Msg("Listing current generated files.")
+
+	g, err := ListStackGenFiles(stack)
+	if err != nil {
+		return nil, fmt.Errorf("checking for outdated code: %v", err)
+	}
+	currentFiles := newStringSet(g...)
```
katcipis, line 237: The overall idea is to:

1 - List all current generated code
2 - As we check for outdated code, we remove files from the set of files
3 - In the end, if all files exist and are OK, the set will be empty, if not...we have some extraneous/old generated files hanging around

generate/generate.go
```diff
@@ -745,14 +781,49 @@ func hasTerramateHeader(code []byte) bool {
 }
 
 func checkGeneratedFilesConflicts(genfiles []genfile) error {
-	observed := map[string]struct{}{}
+	observed := newStringSet()
```
katcipis, line 784: @thiesen now it is explicitely a set since I use it in different places =P

thiesen: :+1: :cool: makes sense!

thiesen (approved):
lgtm!

PR #209, opened 2022-02-09
feat: generate_hcl: fail if we have same block on multiple levels

katcipis:
Now this snippet is supposed to fail, since we have the same block defined on different levels:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
generate_hcl "duplicated.tf" {
    block {}
}
EOM

mkdir -p stacks/stack
terramate stacks init stacks/stack

generate_hcl "duplicated.tf" {
    block {}
}
EOM

terramate generate
```

Also updated the docs with new behavior.


kassianh (approved):
lgtm

thiesen (approved):
lgtm

PR #218, opened 2022-02-14
feat: allow terramate blocks to be merged

katcipis:
This PR allows multiple terramate / terramate.config blocks to be defined on the same file or multiple files, as far as they define different configurations. For example we can have one file with:

```hcl
terramate {
  config {
    git {}
  }
}
```

And another with:

```hcl
terramate {
  config {
    generate {}
  }
}
```

And it will work, by producing a final config that has both the git and the generate configs. The same is possible in the same file.


hcl/hcl.go
```diff
@@ -597,42 +597,38 @@ func parseRootConfig(cfg *RootConfig, block *hclsyntax.Block) error {
...
 			}
 		case "generate":
 			logger.Trace().Msg("Found block generate")
 
-			if foundGenerate {
+			if cfg.Generate != nil {
 				return errutil.Chain(
 					ErrMalformedTerramateConfig,
-					fmt.Errorf("multiple terramate.config.generate blocks"),
+					fmt.Errorf("multiple terramate.config.backend blocks"),
```
i4ki, line 628: we need to provide more details about which files have the duplicated entry, but we can do it in a separate PR.

i4ki (approved):
lgtm

hcl/hcl.go
katcipis: yeah, right now the files from which each block came from are not handy, would require some changes on the internal design (but indeed we need to improve that)

PR #223, opened 2022-02-17
chore: provide proper report on code generation

katcipis:
Now we get a proper report on code generation, for example:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}

cat > stacks/globals.tm.hcl <<- EOM
globals {
    tf_version = "1.1.3"
    gh_provider_version = "4.19.1"
}
EOM

cat > stacks/provider.tm.hcl <<- EOM
generate_hcl "provider.tf" {
    content {
        provider "github" {
          owner = global.gh_owner
        }

        terraform {
          required_providers {
            github = {
              source  = "integrations/github"
              version = global.gh_provider_version
            }
          }
        }

        terraform {
          required_version = global.tf_version
        }
    }
}
EOM

cat > stacks/main.tm.hcl <<- EOM
generate_hcl "main.tf" {
  content {
      module "repositories" {
        source = "../../../modules/terramate-github-repositories"

        repositories        = global.repositories
        repository_defaults = global.repository_defaults
        member_map          = global.member_map
      }
  }
}
EOM

# Now add some per stack globals

cat > stacks/stack-1/globals.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-1"
            visibility = "private"
        }
    ]
    repository_defaults = {
        config = "stack-1"
    }
    member_map  = {
        "first.lastname" = "powerpuffGirls69"
    }
}
EOM

cat > stacks/stack-2/globals.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-2"
            visibility = "public"
        }
    ]
    repository_defaults = {
        config = "stack-2"
    }
    member_map  = {
        "first.lastname" = "onePunchMan666"
    }
}
EOM

terramate generate
```

Will output this:

```
Code generation report

Successes:

- stack /stacks/stack-1
        [+] main.tf
        [+] provider.tf

- stack /stacks/stack-2
        [+] main.tf
        [+] provider.tf

Hint: '+', '~' and '-' means the file was created, changed and deleted, respectively.
```

Errors will also be shown proper per stack, so we wont have any massive single line errors with all errors from all stacks coalesced on them.


generate/generate.go
```diff
@@ -62,14 +62,16 @@ const (
 // The provided working dir must be an absolute path that is a child of the
 // provided root (or the same as root, indicating that working dir is the project root).
 //
-// It will return an error if it finds any invalid Terramate configuration files
-// or if it can't generate the files properly for some reason.
-func Do(root string, workingDir string) error {
-	errs := forEachStack(root, workingDir, func(
+// It will return an report including details of which stacks succeed and failed
```
i4ki, line 65:
```suggestion
// It will return a report including details of which stacks succeed and failed
```

generate/generate.go
```diff
@@ -79,49 +81,65 @@ func Do(root string, workingDir string) error {
...
-			return fmt.Errorf("removing old files to generate new ones: %v", err)
+		removedFiles, err := removeStackGeneratedFiles(stack)
+		failureReport := func(r stackReport, err error) stackReport {
+			r.err = err
+			for filename := range removedFiles {
+				r.addDeletedFile(filename)
+			}
+			return r
+		}
+		if err != nil {
```
i4ki, line 130: what do you think of moving the `err != nil` closer to its assignment?

```
        var removedFiles map[string]string
        var err error
        failureReport := func(r stackReport, err error) stackReport {
			r.err = err
			for filename := range removedFiles {
				r.addDeletedFile(filename)
			}
			return r
		}

        removedFiles, err = removeStackGeneratedFiles(stack)
        if err != nil {
            return failureReport(
				report,
				fmt.Errorf("removing old generated files: %v", err),
			)
		}

File: generate/generate_report.go [hunk 3]
```diff
@@ -0,0 +1,206 @@
...
+	Failures []FailureResult
+}
+
+func (r Report) HasFailures() bool {
+	return r.BootstrapErr != nil || len(r.Failures) > 0
+}
+
+func (r Report) String() string {
+	if r.empty() {
+		return "Nothing to do, code generation is updated"
```
i4ki, line 62: As tests rely on this message, why not put it in an exported const?

i4ki (dismissed):
lgtm but left improvement ideas. Feel free to ignore.

generate/generate_report.go
katcipis: The whole formatting of the report is also not exported properly (maybe if we used a template), so it seemed OK, if you change the formatting of the report only the formatting related tests will fail etc (which are not a lot). The main rationale being that I usually don't like exporting things that I dont see any usage beyond testing, like adding things to an API for test purposes only, and I could not imagine why someone would want to check programatically that the string representation is for an empty report. Like you have better ways to check an empty report, in this case exporting the empty method would seem better, then code can just call report.Empty().

generate/generate_report_test.go
```diff
@@ -0,0 +1,265 @@
...
+			name:   "empty report",
+			report: generate.Report{},
+			want:   "Nothing to do, code generation is updated",
+		},
+		{
+			name: "with bootstrap err",
+			report: generate.Report{
+				BootstrapErr: errors.New("such fail, much terrible"),
+			},
+			want: `Fatal failure preparing for code generation.
```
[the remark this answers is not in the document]
katcipis: commented here: https://github.com/mineiros-io/terramate/pull/223#discussion_r811856276
Not so sure if they are fragile, you are validating the string representation of a type, if the string representation of a type changes (it is something public/exported to the client) then the test fails, correctly, because something public changed on the API. It felt odd to solve this only for simple errors (exporting things on the API that I don't see any use...but maybe there is) but then for the actual report we would have the same problem anyway, if the formatting of the report changes tests that validate the report will break, because..well..the report changed (tests not interested on the string representation of the report test things using the Report type itself, ignoring the string representation, and the ones that are interested on that, if it changes they should fail).

i4ki (approved):

PR #224, opened 2022-02-17
feat: add multiple config files support to export_as_locals 

katcipis:
Now export_as_locals blocks can be loaded from multiple files.

### DIFF
--- a/exported_locals.go
+++ b/exported_locals.go
@@ -29,7 +29,10 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-const ErrExportedLocalRedefined errutil.Error = "export_as_locals attribute redefined"
+const (
+	ErrExportedLocalRedefined errutil.Error = "export_as_locals attribute redefined"
+	ErrExportedLocalsParsing  errutil.Error = "parsing export_as_locals"
+)
 
 // ExportedLocalValues represents information exported from Terramate
 // in a way that is suitable to be used for Terraform code generation.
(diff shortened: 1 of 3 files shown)


exported_locals.go
```diff
@@ -16,20 +16,21 @@ package terramate
...
 	"github.com/mineiros-io/terramate/hcl/eval"
 	"github.com/mineiros-io/terramate/stack"
 	"github.com/rs/zerolog/log"
 	"github.com/zclconf/go-cty/cty"
 )
 
-const ErrExportedLocalRedefined errutil.Error = "export_as_locals attribute redefined"
+const (
+	ErrExportedLocalRedefined errutil.Error = "export_as_locals attribute redefined"
+	ErrExportedLocalsParsing  errutil.Error = "parsing export_as_locals"
```
i4ki, line 32: this doesn't look like a error kind but an 'error context filling text'. By looking at the code that wraps this, it looks like this could be "HCL type errors" and "HCL syntax errors". The "type error" is like when the user didn't provide a valid `export_as_locals` block (eg.: it has labels or other child blocks, etc, valid HCL but invalid block type definition). Maybe the right thing to do is return those sentinels from the `hcl.ParseExportAsLocalsBlocks` function?

By making it export the sentinel, then here you just wrap:

```
return exportedLocalExprs{}, fmt.Errorf("parsing export_as_locals: %w", err)
```

i4ki (approved):
lgtm but I think the error handling could be improved. nothing serious.

exported_locals.go
katcipis: Yeah I think you are right, it felt off while I was doing this but I was on "minimal change to make things work" mode =P. Will address in another PR to avoid blocking the feature. We can improve this a lot by improving how config is loaded/parsed/etc, like we discussed multiple times before..but it would be a considerably bigger change.

PR #228, opened 2022-02-22
refactor: make it clear the distinction of local and remote branches.

i4ki:


test/sandbox/git.go
```diff
@@ -15,70 +15,87 @@
...
 
-// SetupRemote will do basic setup of a remote with the given branch.
-func (git Git) SetupRemote(remote, branch string) {
+func (git Git) ConfigureDefaultRemote() {
+	git.initRemoteRepo(defRemoteBranch)
+	git.RemoteAdd(defRemote, git.remoterepo)
+	// Pushes current branch onto defRemote and defBranch
+	git.PushOn(defRemote, defRemoteBranch, defLocalBranch)
+
+}
```
katcipis, line 70:
```suggestion
}
```

test/sandbox/git.go
```diff
@@ -15,70 +15,87 @@
...
+
+// SetupRemote creates a bare remote repository and setup the local repo with it
+// using remoteName and remoteBranch.
+func (git Git) SetupRemote(remoteName, remoteBranch, localBranch string) {
```
katcipis, line 74: I could be way off on this, but it seems that SetupRemote should not be called multiple times since it would change git.remoterepo and cause some interesting behavior on the git sandbox, if that is the case I would add a check that if it is calledwith al already setup remote it fails immediately, I had some "interesting" problems by making mistakes on my tests and then the errors that happened were really hard to track down since it fails in some distance place (not the setup) and correlating was hard/time consuming.

If it would work with multople remotes setup then just disregard.

katcipis: Taking a second look it seems that we could setup multiple remotes but then the field `git.remoterepo` is confusing, it seems that it is used as some sort of temporary variable between initialization + remoteAdd and nothing else, but in that case it being a field inside the git instance is misleading/confusing (at least to me), I thought it was integral part of the state of the git instance that was changing everytime, but it actually isn't...I think :thinking:

katcipis (approved):
LGTM + small newline change + some doubts around remote setup/internal state of sandbox git instance

test/sandbox/git.go
i4ki: I removed the Git.remoterepo, it wasn't actually needed (at least at this point).

katcipis: yeah if it was fundamental to internal state I would expect that we would not be able to have multiple remotes, or else we would need a slice of remotes instead of a single remote

katcipis (approved):

PR #229, opened 2022-02-22
refactor: add support for configuring sandbox git repo.

i4ki:


test/sandbox/git.go
```diff
@@ -23,24 +23,44 @@ import (
...
+// NewGit creates a new git wrapper using sandbox defaults.
+func NewGit(t *testing.T, repodir string) *Git {
+	cfg := defaultGitConfig()
+	cfg.repoDir = repodir
+
+	return &Git{
+		t:         t,
+		cfg:       cfg,
+		g:         test.NewGitWrapper(t, repodir, []string{}),
+		localrepo: repodir,
```
katcipis: Isn't localrepo now redundant with cfg.repoDir ? maybe we could always just reference cfg.repoDir ?

test/sandbox/sandbox.go
```diff
@@ -190,7 +213,11 @@ func (s S) BuildTree(layout []string) {
 
 // Git returns a git wrapper that is useful to run git commands safely inside
 // the test env repo.
-func (s S) Git() Git {
+func (s S) Git() *Git {
```
katcipis, line 216: This check + multiple new functions kinda give me a feeling that maybe we need two types, like a SandboxBare and Sandbox (with git), or the other way around with Sandbox being bare and having a GitSandox, so the types only have methods valid for them (and we could use composition to help avoid duplication accross them)

So it would be:

```go
type SandboxBare {
   // basic sandbox stuff with no git
}

type Sandbox {
  SandboxBare
  // extra git stuff
}
```

You can even create one from the other, like if you have a SandboxBare instance you can have a SetupGit method that returns a Sandbox, which now represents (also on the type system) a sandbox with git initialized on it.

But Im not sure, it is just that the multipe constructors for the same type and the checks if things are initalized or not seems like maybe something is off (overall I dislike types that can be unitialized and then fail at runtime, so maybe that is what is giving me uneasy feelings), but maybe here it is OK (I think most of the time we want git stuff...so maybe it is not worth to invest much on this).

katcipis (approved):
LGTM + some comments/design discussions but nothing critical, just brainstorming

PR #232, opened 2022-02-23
refactor: make globals a value instead of pointer

katcipis:
Since globals after loading are immutable and we just pass them around using a pointer is misleading (there should be no way to mutate them), and since the hold a map copying them around is not a full copy anyway. It will also make the code more consistent, we already use values for things that are not intended for mutation/side effects.

Also did a small improvement on reporting evaluation failures on globals.


i4ki (approved):
LGTM

PR #237, opened 2022-02-25
chore: improve test hclwrite

katcipis:
Here we:

* Simplify some things on the implementation
* Enforce that the document is generated the same way it was built (no implicit/automatic ordering)
* Allow creation of hcl document with only attributes on root (no blocks)
* Added missing test for genhcl (attributes on root)


i4ki (approved):
top!

PR #240, opened 2022-02-25
refactor: prepare for disable git safeguards.

i4ki:


cmd/terramate/cli/cli.go
```diff
@@ -422,7 +422,38 @@ func (c *cli) initStack(dirs []string) {
...
 }
 
-func (c *cli) listStacks(mgr *terramate.Manager, isChanged bool) ([]terramate.Entry, error) {
+func (c *cli) gitSafeguards(checks terramate.RepoChecks, shouldAbort bool) {
+	logger := log.With().
+		Str("action", "gitSafeguards()").
+		Logger()
+
+	if len(checks.UntrackedFiles) > 0 {
+		if shouldAbort {
```
i4ki, line 431: because of a limitation of `zerolog` we cannot store an instance of generic logger. Eg.:

```
    l := log.With().Logger()
    
    logger := l.Warn()
    if shouldAbort {
         logger = l.Fatal()
    }
    
    logger.Msg("test") // this prints a duplicated msg.
```
the example above is buggy because each logger.Warn()/Info()/.../etc creates a new zerolog event, so if called multiple times it duplicates messages. Have a look:
https://github.com/rs/zerolog/blob/master/log.go#L324

kassianh: @i4ki Couldn't find any solution to this, but maybe this way could work?
```
    l := log.With().Logger()
    
    msg := "test"
    if shouldAbort {
         l.Fatal().Msg(msg)
    }
    
    l.Warn().Msg(msg) 
```

kassianh (approved):
lgtm :)

cmd/terramate/cli/cli.go
i4ki: thanks! I'm gonna improve in a separate PR.

PR #245, opened 2022-03-15
chore: add support to --version

katcipis:
Now we have terramate --version  on top of terramate version, both should work. Code is a little odd because kong parsing fails if there is no subcommand (unless for --help...which I suppose it is handled internally by kong).

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -62,6 +62,7 @@ const (
 
 type cliSpec struct {
 	Version       struct{} `cmd:"" help:"Terramate version."`
+	VersionFlag   bool     `name:"version" help:"Terramate version."`
 	Chdir         string   `short:"C" optional:"true" help:"sets working directory."`
 	GitChangeBase string   `short:"B" optional:"true" help:"git base ref for computing changes."`
 	Changed       bool     `short:"c" optional:"true" help:"filter by changed infrastructure"`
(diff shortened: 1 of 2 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -186,6 +187,15 @@ func newCLI(args []string, stdin io.Reader, stdout, stderr io.Writer) *cli {
 		return &cli{exit: true}
 	}
 
+	// When we run terramate --version the kong parser just fails
```
i4ki, line 190: o.o

katcipis: yup o.O. To double check you can do this:

```
terramate --log-level="info"
2022-03-16T13:13:14+01:00 FTL failed to parse cli args: [--log-level=info] error="expected one of \"version\",  \"run\",  \"plan\",  \"stacks\",  \"generate\",  ..." action=newCli()
```

The error is "failed to parse cli args": https://github.com/mineiros-io/terramate/blob/2a152446ca7d7e0f38ef70c92798e95c713966d4/cmd/terramate/cli/cli.go#L192

Kong is just great :-)...I have no idea how we can make something that has no subcommands work, like just having flags...I tried checking the docs but it is way too big/annoying and I wasn't able to find anything so I just brute forced x_x.

i4ki (approved):

PR #246, opened 2022-03-15
feat: add tm_ prefix on funcs for globals/export_as_locals

katcipis:
Both globals and export_as_locals now only evaluated functions when prefixed with tm_.


i4ki (approved):
lgtm

PR #248, opened 2022-03-16
tests: add new globals tests

katcipis:
These help us map some funny behavior on the string interpolation on hcl eval, which then helps on more advanced features like partial eval.


i4ki (approved):
LGTM

PR #249, opened 2022-03-16
tests: some more tests involving interpolation

katcipis:
Based on discoveries done through Globals evaluation on this PR: https://github.com/mineiros-io/terramate/pull/248
Behavior for string interpolation is quite odd, so tried to map it here (not the error situations yet).
Im skipping the failing tests, so you can enable them and make them pass when it makes sense, without causing too much noise.


hcl/eval/partial.go
```diff
@@ -23,6 +23,32 @@ import (
 	"github.com/hashicorp/hcl/v2/hclwrite"
 )
 
+/*
+
+Here be dragons. Thou art forewarned
```
i4ki, line 28: :laughing: :heart: top!

i4ki (approved):
THANKS!!!

PR #252, opened 2022-03-17
chore: move terramate stacks list to terramate list

katcipis:


cmd/terramate/e2etests/general_test.go
```diff
@@ -425,7 +424,7 @@ terramate {
 }
 `)
 
-	assertRun(t, cli.run("stacks", "list", "--changed"))
+	assertRun(t, cli.listChangedStacks())
```
mariux, line 427: why not test the run command here? in case internal implementation changes this should be as close to the user as possible?

katcipis: we keep testing using cli.run, I just extracted the details of the cli args on a method so next time we need to change how to list stacks the change is more localized, you can check listStacks and listChangeStacks [here](https://github.com/mineiros-io/terramate/pull/252/files#diff-4dc6b8809155122f5578d8ef1258ff37ab2cbe0069046d06134502459a88c4a9R110-R117).

i4ki (approved):

PR #256, opened 2022-03-18
test: add simple attr test

katcipis:
Since we were discussing if something like:

```hcl
a = b
```

would work, at least on the parsing level it is.


i4ki (approved):
thanks

PR #265, opened 2022-03-22
chore: change release tar names

katcipis:
Copied the config from GoReleaser but using Linux instead of linux on the tar file is just odd :laughing:


soerenmartius (approved):
lgtm!

PR #274, opened 2022-03-25
chore: document the "wants" feature

i4ki:


docs/orchestration.md
```diff
@@ -1,16 +1,59 @@
...
+1. Change detection
+
+The [change detection](./change-detection.md) filter out stacks not changed.
+
+2. Current directory
+
+Terramate uses the current directory to filter out stacks, ie., limit the scope
+of the execution, so if you execute `terramate` from the project's root
+directory, all stacks will be selected, and changing to inner directories in the
+project structure will select only the child stacks.
```
katcipis, line 21:
```suggestion
Terramate uses the current directory it is being executed to filter out stacks, ie., limit the scope
of the execution, so if you execute `terramate` from the project's root
directory, all stacks will be selected, and changing to inner directories in the
project structure will select only stacks that are children of the current directory.
```

docs/orchestration.md
```diff
@@ -1,16 +1,59 @@
...
+
+3. Explicit `wants` relationship.
+
+The `wants` attribute of the stack block defines an explicit relationship
+between two stacks, that when provided it says that selecting this stack *also
+selects the provided list of stacks*. So instead of filtering this explicitly
+selects stacks that would have been filtered out by the other methods.
```
katcipis, line 28:
```suggestion
The `wants` attribute of the stack block defines an explicit relationship
between a stack and its list of wanted stacks, that when provided it says
that when a stack is selected all the stacks listed on its `wants` list will also be
selected, always, independent of any other selection criteria. 
```

docs/orchestration.md
```diff
@@ -1,16 +1,59 @@
...
+stack {
+    wants = [
+        "/other/stack1",
+        "/other/stack2"
+    ]
+}
+```
+
+These 3 selection methods could be used together, and the order which they are
+applied is: `change detection`, `current directory`, `wants`.
```
katcipis, line 42: Hmm don't we first get stacks on the current dir and then do change detection/wants only on the stacks that are inside the current dir ? :thinking: 

What I mean is that changes in a stack outside the current dir will be ignored, but saying that change detection comes first makes me think that it would not, since change detection comes first. Wouldn't it be:

1 - Stacks that are chidren of the current dir are selected.
2 - These stacks may be filtered if change detection is on.
3 - Given the filter result, wants for these stack is loaded.
4 - Any stacks on the wants list then are added, independent of anything else.

katcipis (commented):
Some considerations regarding the order that stack selection happens.

docs/orchestration.md
i4ki: this happens in this order because the ListChanged() works at the git root level, we ask git for all files changed and select which one is part of a stack, and so on.

katcipis: But isn't that an implementation detail ? The observed behavior is that changed stacks will be selected from stacks inside your cwd. I really think it is confusing to say that change detection comes first, it makes me think that first it detects all changed stacks and then also list only stacks on the cwd.

Like if I run:

```
terramate list
```

On a subdir, it will list only stacks on the subdir. But if I run:

```
terramate list --changed
```

Now it may list stacks outside the cwd, because change detection comes first. It could be interpreted as a ordered chain of filters, like first changed + filter by cwd + wants, but wants doesn't make much sense on this case when thinking on a filter pipeline, since it actually adds stacks that were not present :thinking: .

Anyway, I was not talking about details on how the code is implemented, but what is the expected observed behavior and when thinking on external terramate behavior it was odd to see change detection coming first.

katcipis: > this happens in this order because the ListChanged() works at the git root level, we ask git for all files changed and select which one is part of a stack, and so on.

Yeah I remember that, we also need to check if modules changed and then correlate to stacks that use them etc, I was just thinking more on the user/observed behavior side and I understood it as "change detection comes first" in a way that it superseds the cwd, so it seemed odd...but all good.

katcipis (approved):

docs/orchestration.md
i4ki: `terramate list --changed` has the same order precedence, it lists all stacks that changed, then filter by cwd. The thing is that cwd is always present, so it's always implicitly filtering, so maybe that's why you feel it's strange? but I think the implementation and the observed behavior are the same.

PR #285, opened 2022-03-30
feat: add file generation

kassianh:
https://app.shortcut.com/mineiros/story/5092/add-support-for-generate-file


generate/genfilehcl/genfile_test.go
```diff
@@ -0,0 +1,1327 @@
...
+
+				delete(got, res.name)
+			}
+
+			assert.EqualInts(t, 0, len(got), "got unexpected exported code: %v", got)
+		})
+	}
+}
+
+func TestPartialEval(t *testing.T) {
```
katcipis, line 249: I don't think we have partial evaluation on generate_file, we just fully evaluate contents and then save the evaluated result in a file

generate/genfilehcl/genfilehcl.go
```diff
@@ -0,0 +1,287 @@
...
+
+		val := ""
+
+		if !value.IsNull() {
+			val = value.AsString()
+		}
+
+		res.hcls[name] = HCL{
+			origin: loadedFileBlock.origin,
+			body:   hclwrite.Format([]byte(val)),
```
katcipis, line 140: Why are we formatting the value as HCL if the idea is to generate arbitrary data ?(YAML/JSON/etc)

generate/genfilehcl/genfile_test.go
```diff
@@ -0,0 +1,1327 @@
...
+			configs: []hclconfig{
+				{
+					path: "/stack",
+					add: generateHCL(
+						labels("test.json"),
+						expr("content", "\"something = tm_max(5, 12, 9)\""),
+					),
+				},
+			},
+			want: []result{
```
katcipis, line 123: Kinda odd that the tests "want" an "hcl", the idea of generate_file is to generate random files (JSON/YAML/some text/etc) :thinking: .

katcipis:
Implemented on #330. Thanks for the work on this @kassianh :rocket:

PR #288, opened 2022-04-01
chore: add project setup docs

katcipis:


README.md
```diff
@@ -100,4 +112,57 @@ terramate install-completions
...
+Terramate is as easy as just [installing Terramate](#installing).
+Terramate comes with sensible defaults so just using it inside a pre existent
+Git repository should not require any configurations.
+
+The exception being repositories that have a default remote branch
+other than `origin/main`, in that case to make change detection work you will
+need to set a customized [project configuration](docs/project-config.md).
+
+If you want to play around with Terramate from scratch locally you can also
+setup a local git repository:
```
i4ki, line 125: just wondering if it's not easier and less troublesome if we just say they could create and clone an empty GH repository?

i4ki (approved):
LGTM

README.md
katcipis: I wondered the same ...that is why I also started by saying that if you already have a project versioned on Git it would be trivial to set Terramate up. But I'm not sure about this either. WDYT @mariux @soerenmartius ?

i4ki (approved):

PR #290, opened 2022-04-04
chore: remove duplicated import

i4ki:


mariux (approved):
lgtm.. the naming seems odd as this is goversion not HCL?

i4ki:
@mariux I agree, any suggestion?

I called this `tfversion` in the first implementation because we are using the same version lib used in TF but this is also misleading because it's a separate repository likely used by other hashicorp products as well, that's why I kept @katcipis naming.

What about `semver`? Then the usages will be `semver.NewConstraint()` and `semver.NewSemver()`. But this go-version also don't follow `semver` specification strictly so...

katcipis (approved):

katcipis:
> @mariux I agree, any suggestion?
> 
> I called this `tfversion` in the first implementation because we are using the same version lib used in TF but this is also misleading because it's a separate repository likely used by other hashicorp products as well, that's why I kept @katcipis naming.
> 
> What about `semver`? Then the usages will be `semver.NewConstraint()` and `semver.NewSemver()`. But this go-version also don't follow `semver` specification strictly so...

semver kinda makes sense, but it is also a fact they don't follow strict semver...so though spot to do naming here x_x

PR #291, opened 2022-04-05
chore: refactor error handling to support file range whenever possible.

i4ki:


errors/error.go
```diff
@@ -0,0 +1,268 @@
...
+	Kind Kind
+
+	// Description of the error.
+	Description string
+
+	// FileRange holds the error source.
+	FileRange hcl.Range
+
+	// Stack which originated the error.
+	Stack Stack
```
katcipis, line 42: hmm why not just use the stack.S type ? then we could create an error related to a stack passing the stack itself, getting a string will allow different parts of the code to pass a different representation of the stack, like some code passing the name, other part passing the absolute path, etc.

errors/error.go
```diff
@@ -0,0 +1,268 @@
...
+		case string:
+			e.Description = arg
+		case error:
+			e.Err = arg
+		default:
+			panic(fmt.Errorf("called with unknown type %T", arg))
+		}
+	}
+	if e.isEmpty() {
+		panic(fmt.Errorf("empty error"))
```
katcipis, line 133:
```suggestion
		panic(errors.New("empty error"))
```

errors/error.go
```diff
@@ -0,0 +1,268 @@
...
+		if e.Err != nil {
+			return Is(e.Err, target)
+		}
+		return false
+	}
+
+	// both are *Error
+	if IsKind(e, t.Kind) {
+		return true
+	}
```
katcipis, line 263: hmm so errors of the same kind but for different stacks would be true when checking errors.Is() ? :thinking: 
we could ignore fields that are not present on the target error, so it would be a partial match, but if both errors have kind + stack and stack differs it seems odd to detect as true, or maybe Im missing something here.

errors/error_test.go
```diff
@@ -0,0 +1,185 @@
...
+			e := errors.E(tc.args...)
+			assert.EqualStrings(t, tc.want, e.Error())
+		})
+	}
+}
+
+func TestErrorIs(t *testing.T) {
+	type testcase struct {
+		name    string
+		args    []interface{}
```
katcipis, line 130:
```suggestion
		err    E
```

Not so sure what we gained with defining test cases like this:

```go
{
			name:   "nil target is not equal",
			args:   []interface{}{"any error"},
			target: nil,
},
```

Instead of (which IMHO reads better):

```go
{
			name:   "nil target is not equal",
			args:   E("any error"),
			target: nil,
},
```

Unless we test when the constructor panics or something, but doesn't seem to be the case here.

errors/error.go
i4ki: Because `errors` must be imported by every other package, then importing any other TM package here will make cycles...

We can add a `filepath.IsAbs()` check so it helps catch this.

errors/error.go
i4ki: I thought about this but the same thing could be said about the other metadata/info (filerange, description, etc). Currently (I mean what we have in `main` now) the stack info goes in the error message so we also don't check this. 

We can add this improvement in a separate PR together with the tests changes to make use of it, but I think everything available in the `target` should be used for match. Just don't think make sense adding this now because I'll have to work in the tests to use them, otherwise it's just dead code.

errors/error_test.go
i4ki: yep, this is leftover from the previous iteration.

errors/error.go
katcipis: As discussed out of the PR, we can also go with an interface to break the cycle, which seems nice IMO, so we can still pass a stack directly when creating errors for it + no need to check things for invariants since the stack itself already did that.

cmd/terramate/cli/cli.go
```diff
@@ -498,7 +497,7 @@ func (c *cli) printStacks() {
 
 	for _, entry := range report.Stacks {
 		stack := entry.Stack
-		stackRepr, ok := c.friendlyFmtDir(stack.PrjAbsPath())
+		stackRepr, ok := c.friendlyFmtDir(stack.Path())
```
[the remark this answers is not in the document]
i4ki: those coverage warnings are not correct... these lines are tested with e2e test...

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+	// Description of the error.
+	Description string
+
+	// FileRange holds the error source.
+	FileRange hcl.Range
+
+	// Stack which originated the error.
+	Stack StackMeta
+
+	// Err represents the underlying error.
```
katcipis, line 44:
```suggestion
	// Err holds the underlying error.
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+	Err error
+}
+
+type (
+	// Kind defines the kind of error.
```
katcipis, line 49:
```suggestion
	// Kind defines the kind of an error.
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+	Kind string
+	// StackMeta has the metadata of the stack which originated the error.
+	// Same interface as stack.Metadata.
```
katcipis, line 52:
```suggestion
	// StackMeta provides metadata about the stack that originated the error.
```

thinking if it makes sense to explicit document that the interface is copied :thinking: 

Also thinking if StackMetadata would not be more clear (could be just me more used to always referencing it as metadata not only as meta).

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+// The type of each argument determines its meaning. If more than one argument
+// of a given type is presented, only the last one is recorded.
+//
+// The types are:
+// 	errors.Kind
+//		The kind of error (eg.: HCLSyntax, TerramateSchema, etc).
+//	hcl.Range
+//		The file range where the error originated.
+//  errors.Stack
+//		The stack which the error originated.
```
katcipis, line 73:
```suggestion
//		errors.StackMeta
//		The stack that originated the error.
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+//	error
+//		The underlying error that triggered this one.
+//	hcl.Diagnostics
+//		The underlying hcl error that triggered this one.
+//		If this error's FileRange is not set, the diagnostic subject range is
```
katcipis, line 78:
```suggestion
//		If hcl.Range is not set, the diagnostic subject range is
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+//		pulled.
+//		If this error's Description is not set, the diagnostic detail field is
```
katcipis, line 80:
```suggestion
//		If the string Description is not set, the diagnostic detail field is
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+//		pulled.
+//	string
+//		The error description. It supports formatting using the Go's fmt verbs
+//      as long as the arguments are not of the defined types.
```
katcipis, line 84:
```suggestion
//	string
//		The error description. It supports formatting using the Go's fmt verbs
//		as long as the arguments are not one of the defined types.
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+//
+// If the error is printed, only those items that have been
+// set to non-zero values will appear in the result. For the `hcl.Range` type,
+// the `range.Empty()` method is used.
+//
+// Error promotions:
+//
+// If Kind is not specified or Any, we set it to the Kind of
```
katcipis, line 92: Do we have an  "Any" kind ?

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+// the underlying error.
+// If FileRange is not specified, we set it to the FileRange of the underlying
+// error (if set).
```
katcipis, line 95:
```suggestion
// The following fields are promoted from underlying errors when absent:
//
// - errors.Kind
// - errors.StackMeta
// - hcl.Range
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+//
+// Minimization:
+//
+// In order to avoid duplicated messages, we erase the fields present in the
+// underlying error if already set with same value in this error.
```
katcipis, line 100:
```suggestion
// In order to avoid duplicated messages, when wrapping an errors.E inside another errors.E 
// any fields present on the current error overrides the fields defined on the wrapped error.
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+		case Kind:
+			e.Kind = arg
+		case hcl.Range:
+			e.FileRange = arg
+		case hcl.Diagnostics:
+			diags = arg
+			if diags.HasErrors() {
+				diag := diags[0]
+				if diag.Subject != nil {
+					e.FileRange = *diag.Subject
```
katcipis, line 125: The range...on the diagnostics...is called SUBJECT ?

terror

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// Is tells if err (or any of its underlying errors) matches target.
+// It works with any error but if comparing *Error type it uses the Kind field,
+// otherwise it fallback to standard errors.Is().
+func Is(err, target error) bool {
+	if (err == nil) != (target == nil) {
+		return false
+	}
+	e, ok := err.(*Error)
```
katcipis, line 303: when (err == nil) && (target == nil) this will panic, do we want it to panic ? Go's errors.Is returns true if both errors are nil: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/errors/wrap.go;l=41

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+	}
+	if e.Kind != "" && e.Kind == k {
+		return true
+	}
+	return IsKind(e.Err, k)
+}
+
+// Is tells if err (or any of its underlying errors) matches target.
+// It works with any error but if comparing *Error type it uses the Kind field,
+// otherwise it fallback to standard errors.Is().
```
katcipis, line 298:
```suggestion
// It works with any error but when comparing with an errors.Error type it will try to match the following fields:
// - Kind
// - Description
// - Stack
// - FileRange
// - Err
// Any fields absent (empty) on the target error are ignored even if they exist on err (partial match).
// If the target error is no an error.Error it fallbacks to standard errors.Is().
```

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+		return true
+	}
+	if e.Err != nil {
+		return hasDesc(e.Err, desc)
+	}
+	return false
+}
+
+// hasSameStack recursively check if err or any of its underlying errors has the
+// provided stack set.
```
katcipis, line 358:
```suggestion
// hasSameStack recursively check if err or any of its underlying errors has
// the same stack metadata.
```

katcipis (commented):
Some more comments but still haven't finished it :-)

errors/error.go
katcipis: Since on the context of the errors.E constructor we are usually talking about the types of the variadic args passed, not the field on the struct they will be stored.

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+	emptyRange := hcl.Range{}
+	if e.FileRange == emptyRange {
+		e.FileRange = prev.FileRange
+	}
+	if prev.FileRange == e.FileRange {
+		prev.FileRange = hcl.Range{}
+	}
+	if prev.Stack == e.Stack {
+		prev.Stack = nil
+	}
```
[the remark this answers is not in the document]
katcipis: Also, feel free to disregard this because maybe Im being pedantic, but comparing using interfaces here seems risky if there is any other field on the underlying object that implements the interface that may differ and we are not interested in. For our purposes we are interested on then having the same name/path/description and that is it. Also, even though it is a stretch, if there is any type implementing this interface that is not comparable this would panic.

When doing the error matching we are already comparing by calling the interface methods: https://github.com/mineiros-io/terramate/pull/291/files#diff-d128886dfda28afbfb957c032a184b95f625ca12ff25c8656247783f3e9bf112R364

So maybe we could do it here too.

errors/error_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"testing"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/madlambda/spells/assert"
+	"github.com/mineiros-io/terramate/errors"
+)
+
+var E = errors.E
+
+var (
```
katcipis, line 30:
```suggestion
const (
```

errors/error_test.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+func TestNoArgs(t *testing.T) {
+	defer func() {
+		err := recover()
+		if err == nil {
+			t.Fatal("E() did not panic")
+		}
+	}()
+	_ = E()
```
katcipis, line 43: Maybe we could have something like a assert.Panic(func() {}) that executes the given function and checks it panics.

errors/error_test.go
```diff
@@ -0,0 +1,217 @@
...
+		{
+			name: "nested errors with last one with range",
+			err: E("1", E("2", E("3", hcl.Range{
+				Filename: "test.tm",
+				Start:    hcl.Pos{Line: 1, Column: 5, Byte: 3},
+				End:      hcl.Pos{Line: 1, Column: 10, Byte: 13},
+			})),
+			),
+			want: "test.tm:1,5-10: 1: 2: 3",
+		},
```
katcipis, line 141: I think we are missing the behavior when building errors using hcl.Diagnostics here.

errors/error.go
```diff
@@ -0,0 +1,373 @@
...
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/hcl/v2"
+)
+
+// Error is the default Terramate error type.
+// It may contains some fields unset, but not all.
```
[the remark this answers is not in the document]
i4ki: I know "partial" means incompletion but saying it could be partially set imply at least one must be set?

What about: At least one of the error fields must be set.
?

errors/error_test.go
```diff
@@ -0,0 +1,217 @@
...
+
+func TestErrorIs(t *testing.T) {
+	type testcase struct {
+		name    string
+		err     error
+		target  error
+		areSame bool
+	}
+
+	for _, tc := range []testcase{
```
katcipis, line 157: Maybe we could add a test to define when both are nil here.

errors/testing.go
```diff
@@ -0,0 +1,62 @@
...
+	t.Helper()
+	if (got == nil) != (want == nil) {
+		t.Fatalf("got error[%v] differs from want[%v]", got, want)
+	}
+	if want == nil {
+		return
+	}
+	e1, ok := got.(*Error)
+	if !ok {
+		t.Fatal("got is not an *errors.Error")
```
katcipis, line 30:
```suggestion
		t.Fatalf("got %v is not an *errors.Error", got)
```

errors/testing.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	e2, ok := want.(*Error)
+	if !ok {
+		t.Fatal("want is not an *errors.Error")
```
katcipis, line 35:
```suggestion
		t.Fatalf("want %v is not an *errors.Error", want)
```

errors/testing.go
```diff
@@ -0,0 +1,62 @@
...
+// AssertIsKind asserts err is of kind k.
+func AssertIsKind(t *testing.T, err error, k Kind) {
+	t.Helper()
+	if !IsKind(err, k) {
+		t.Fatalf("error[%v] is not of kind %q", err, k)
+	}
+}
+
+// Assert err is (contains, wraps, etc) target.
+func Assert(t *testing.T, err, target error) {
```
katcipis, line 50:
```suggestion
// AssertIs err is (contains, wraps, etc) target.
func AssertIs(t *testing.T, err, target error) {
```

I think just assert is too vague, specially since this is a specialized assertion built on top of "is" behavior (not some simple boolean check etc).

errors/error.go
i4ki: I think it's important to add this hint because changes to this type must be aligned with the definition of stack metadata in the stack package.
Now thinking about this I also would say that we should add a similar hint in the stack package like:

```
// Metadata has all metadata loaded per stack
// Hint: changes to this interface must be replicated in errors.StackMeta.
```

What do you think?

errors/testing.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2022 Mineiros GmbH
```
katcipis, line 1: The idea of exporting some test facilities is really cool, Im just not so sure if we should do it in the same package/namespace of the actual code, like when you see the package docs there will also be a bunch of assertion functions there...maybe it is OK and it is just some subjective bias because I never saw it before, but could make it more annoying for people who are just looking to use the errors API and not actually write tests (although they should write tests too :laughing: ).

Maybe we could go with "test/errors" like we have for other testing facilities on the project.

katcipis (commented):
Not finished yet but GH interface is dumb and doesn't allow me to answer any discussions while Im reviewing code....because....

errors/error.go
katcipis: Hmm I disagree. Any incompatibilties on the interfaces will be caught at build time already, IMO it only clobbers the docs (imagine a third interface somewhere else, now we have to add comments on all places about all other places). But if you want to do it Im OK, not a strong opinion or anything.

errors/error.go
i4ki: outdated docs, thanks!

errors/error.go
katcipis: I thought intuitively that stacks were going to be promoted, but I think I was wrong on that so disregard this

generate/generate.go
```diff
@@ -73,20 +72,19 @@ func Do(root string, workingDir string) Report {
 		stack stack.S,
 		globals terramate.Globals,
 	) stackReport {
-		stackpath := stack.AbsPath()
+		stackpath := stack.HostPath()
```
katcipis, line 75: Noice

errors/error.go
i4ki: overrides is not the correct term because if the underlying error has different field values, we keep them for the error context. We just clear the fields in the underlying error *when* they have the same value in the current error.
I also don't know if saying "wrapping" is a good term because this concept in Go already means wrapping errors with `%w` and this is not supported here.

globals_test.go
```diff
@@ -1067,21 +1068,28 @@ func TestLoadGlobals(t *testing.T) {
...
 			wantGlobals := tcase.want
 
-			stacks := s.LoadStacks()
-			for _, stack := range stacks {
-				stackMeta := stack.Meta()
-				got, err := terramate.LoadStackGlobals(s.RootDir(), stackMeta)
+			stackEntries, err := terramate.ListStacks(s.RootDir())
+			if errors.IsKind(tcase.wantErr, hcl.ErrTerramateSchema) {
+				errors.AssertKind(t, err, tcase.wantErr)
+			}
```
katcipis, line 1074: When tcase.wantErr != hcl.ErrTerramateSchema (like nil) is it safe/best to go on with the test if listing stacks failed ? I imagine the stackEntries will be an empty slice and the test failure will be harder to understand because there will be no indication that stack listing failed (the error will indicate that something wrong happened loading the globals) :thinking:

errors/error.go
i4ki: it kinda makes sense for me... because just Range says nothing, FileRange says a bit more but still what range? The word *Subject* at least points that this field contains the offending part, the topic of the diagnostics, and so on. They also provide a *Context* field with a broader range so you can print/render this part of the code in an error in order to easily spot the problem because sometimes the `Subject` is a tiny expression inside an expression and so on.
But Context can be nil, which means the Context == Subject, but they don't provide any helper method for that. For the extension the tighter range is enough but I think we can provide better error messages using the Context (if we get rid of logging for showing errors in text mode).

katcipis: was unaware they had two ranges, if there is just one it is simple to me that it is the file range of the source that caused the problem/diagnostic and calling subject is just confusing, but if they have multiple ranges then it may make sense.

errors/error.go
i4ki: > Any incompatibilties on the interfaces will be caught at build time already
how? errors.E() receives interface{}...

katcipis: ah true.. still it seems odd to keep documenting this and hoping it will be enough to avoid issues, for a scenario like that it would be safer to just define a single interface/type and use it on all places then... but that would introduce some sort of types package.

errors/error.go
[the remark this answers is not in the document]
i4ki: this comparison is there since errors.Stack was a string. It needs to be fixed. Thanks for catching it.

hcl/hcl.go
```diff
@@ -92,22 +97,108 @@ type Stack struct {
 // Blocks maps a filename to a slice of HCL blocks associated with it
 type Blocks map[string][]*hclsyntax.Block
 
-// Errors returned when parsing Terramate configuration.
-const (
-	ErrHCLSyntax                errutil.Error = "HCL syntax error"
-	ErrMalformedTerramateConfig errutil.Error = "malformed terramate config"
-	ErrMalformedTerraform       errutil.Error = "malformed terraform"
-)
+// TerramateParser is an HCL parser tailored for Terramate configurion schema.
```
katcipis, line 100:
```suggestion
// TerramateParser is an HCL parser tailored for Terramate configuration schema.
```

errors/error.go
i4ki: this message would be accurate if both params were errors and we were looking if they have the same metadata.

katcipis: hmm but what would be a "stack set" here ? I just don't get it :thinking:

i4ki: ... has the provided stack set to its Stack field. ?

katcipis: hmm for some reason I was always thinking about "sets"...like a set of stacks :laughing: . Maybe Im just extra stupid..but I was trying to improve this because it is confusing at least to me. But it is probably an isolated dumbness on my part, disregard.

hcl/hcl.go
```diff
@@ -173,45 +264,36 @@ func ParseModules(path string) ([]Module, error) {
...
 		if block.Type != "module" {
 			continue
 		}
 
 		if len(block.Labels) != 1 {
-			return nil, errutil.Chain(
-				ErrMalformedTerraform,
-				fmt.Errorf("module block must have 1 label"),
-			)
+			return nil, errors.E(ErrTerraformSchema, block.OpenBraceRange,
```
katcipis, line 281: nice :rocket:

hcl/hcl_test.go
```diff
@@ -132,30 +134,38 @@ module "bleh" {
...
+			configdir := t.TempDir()
+			tfpath := test.WriteFile(t, configdir, "main.tf", tc.input)
 
-			modules, err := hcl.ParseModules(path)
-			assert.IsError(t, err, tc.want.err)
+			if tc.want.err != nil {
+				if e, ok := tc.want.err.(*errors.Error); ok {
+					if e.FileRange.Filename != "" {
+						e.FileRange.Filename = tfpath
+					}
```
katcipis, line 163: hmm if the wanted error defines a filename in the range we overwrite it, if none is defined (empty string) we leave it empty. Is this the behavior we want here ? (usually we set automatically something when none is defined, like a default)

hcl/hcl_test.go
```diff
@@ -1009,6 +1044,24 @@ func testParser(t *testing.T, tc testcase) {
 	}
 }
 
+// some helpers to easy build file ranges.
```
katcipis, line 1047: noice :rocket:

init.go
```diff
@@ -39,47 +40,49 @@ func Init(root, dir string) error {
...
+		return errors.E(errInit, err)
 	}
 
 	if found {
-		return fmt.Errorf("directory %q is inside stack %q but nested stacks are disallowed",
-			dir, parentStack.PrjAbsPath())
+		return errors.E(errInit,
+			"directory %q is inside stack %q but nested stacks are disallowed",
+			dir, parentStack.Path(),
+		)
```
katcipis, line 78:
```suggestion
		return errors.E(errInit, parentStack,
			"directory %q is inside stack but nested stacks are disallowed",
			dir,
		)
```

list.go
```diff
@@ -50,18 +49,14 @@ func ListStacks(root string) ([]Entry, error) {
 				return filepath.SkipDir
 			}
 
-			logger.Trace().
-				Str("stack", path).
-				Msg("Try load stack.")
+			logger.Trace().Str("stack", path).Msg("Try load stack.")
```
katcipis, line 52: noice :100:

katcipis (approved):
LGTM + bunch of comments where Im possibly just wrong/being dumb so feel free to disregard.

errors/error_test.go
i4ki: cool

errors/error.go
i4ki: fixed.

errors/error_test.go
i4ki: fixed

errors/error_test.go
i4ki: fixed

errors/testing.go
i4ki: yeah I thought the same thing but I didn't know it was worth the effort. Will do.

hcl/hcl_test.go
i4ki: fixed

globals_test.go
i4ki: fixed

errors/testing.go
i4ki: the context is in the package name: `errors.Assert(err1, err2)`.. it assert that errors match.

katcipis: hmm not so sure that is clear enough, assertions are usually about something being true (predicates) or comparing things (equality). The matching logic implemented by errors.Is (Go's and ours) is different from most classical assertions. One example is that this assertion function is not symmetrical, `errors.Assert(t, err1, err2)` != `errors.Assert(t, err2, err1)`, for a function called just Assert, even though inside the errors package, that is a little odd IMHO. But not a huge deal, just sharing my perception on this.

katcipis (approved):
YEAHHH :rocket:

PR #292, opened 2022-04-05
fix: after/before selecting stacks outside cwd/-C

katcipis:
When running:

```
terramate -C <dir>
```

Stacks that are outside the subdir tree/-C `dir` are being selected on `terramate run`.

The idea of the PR is to ensure stacks returned by the DAG building process exists on the selected stacks passed to the order function, so it is responsibility of the run.Sort function caller to pass a properly filtered set of stacks, which is what is done today already unless I missed something.


i4ki (approved):
LGTM

Thanks bro.

PR #300, opened 2022-04-11
fix: semantic versioning does not care about -dev so lets use -alpha

mariux:
fixes https://github.com/mineiros-io/terramate-example-code-generation/runs/5979406505?check_suite_focus=true#step:7:7


katcipis:
The go-version lib doesn't implement strict semver, specially the constraint logic since according to them is not specified on the spec. You can check in details here:

https://github.com/hashicorp/go-version/pull/35#issuecomment-373078159

"This lib doesn't strictly follow semver so it wouldn't be unprecedented to merge, just curious"

a version that is `2.0.0-alpha` will only be accepted on a constraint when it is compared to 2.0.0-<something>, any other version comparison will fail the constraint and that is expected when using this lib + constraints. So changing the semver pre-release segment to -alpha will not fix anything AFAIK.

I suppose this is also the terramate behavior, since it is a hashi lib.

mariux:
@katcipis whats your take on fixing this then?

katcipis (approved):

katcipis:
Terraform itself uses -dev: https://github.com/hashicorp/terraform/blob/8040dfec3467d9dca7561d944ff9642bd0699ff0/version/version.go#L19

Didn't find anything on the go-version lib that handles alpha specially, if that works on terraform it would be some extra logic on terraform itself :thinking: ...or maybe I missed something. @i4ki and dug throught eh go-version lib, WDYT ?

i4ki:
I didn't find any special handling in the terraform version check code as well.
:thinking:

katcipis:
Some local testing that @i4ki did:

```
$ terramate list
2022-04-12T14:15:09+01:00 FTL  error="version check error: version constraint \"~> 0.0.11\" not satisfied by terramate version \"0.0.13-alpha\"" action=cli.checkVersion() root=/home/i4k/src/mineiros/tm-example2
```

```
$ cat terramate.tm
terramate {
	config {
	}
	required_version = "~> 0.0.11"
}
```

Just to confirm, this is the scenario that is supposed to work by using -alpha instead of -dev ? I doesn't seem like it is going to work :thinking: (terraform also uses -dev on main: https://github.com/hashicorp/terraform/blob/8040dfec3467d9dca7561d944ff9642bd0699ff0/version/version.go#L19)

PR #305, opened 2022-04-16
chore: update setup go version +add newer macos

katcipis:


soerenmartius (approved):
lgtm!

i4ki (approved):
lgtm but why update to v3?

.github/workflows/ci.yml
```diff
@@ -43,16 +43,17 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      fail-fast: false
```
i4ki, line 46: is this really required? by reading the docs it seems it's mostly used to get quicker feedback when changing the workflow, so you can iterate faster on the CI changes, but when it's ready (meaning the CI config is ok) it should be "on", right?

.github/workflows/ci.yml
```diff
@@ -43,16 +43,17 @@ jobs:
...
       matrix:
-        os: ["ubuntu-20.04", "macos-10.15"] # maybe in the future: windows-2019
+        os: ["ubuntu-20.04", "macos-10.15", "macos-11"] # maybe in the future: windows-2019, windows-2022
```
i4ki, line 48: :+1:

.github/workflows/ci.yml
katcipis: When this is ON if we have a failure only on macos + go version X it will stop/abort all other parallel builds on the build matrix. So we get a faster failure but we lose the information if a specific change is going to work on other combinations of OS + Go version. The main drawback is only taking longer, IMO it is more desireable to in the face of a failure be sure if the failure is on all supported OS/versions than being faster.

katcipis:
> lgtm but why update to v3?

I was checking some things on the plugin and I saw they mentioning that v3 has some bug fixes, so it seemed like a good idea to update to it.

.github/workflows/ci.yml
katcipis: + as explored with @i4ki GH doesn't provide a proper way to run just an aborted step, so there is no good way to check if an aborted step will succeed or fail (it always runs all steps or all failed steps, but that keeps the aborted step being aborted).

PR #308, opened 2022-04-20
fix: disallow generating files outside of the stack

katcipis:
Disallows having any "/" on the filename that will be used  to save generated code (will work for any code generation strategy).


generate/generate.go
```diff
@@ -543,6 +553,16 @@ func checkGeneratedFilesConflicts(genfiles []genfile) error {
 	return nil
 }
 
+func checkGeneratedFilesPaths(genfiles []genfile) error {
+	for _, gen := range genfiles {
+		if strings.Contains(gen.name, "/") {
```
i4ki, line 558: I think we should check for `filepath.Separator` otherwise on windows `dir\test.tf` will be valid, or we convert everything to ForwardSlash before calling the generation.

generate/generate_report_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	tmerrors "github.com/mineiros-io/terramate/errors"
```
i4ki, line 22: let's use `errors` by default and `stderrors` if we ever need the standard one?

generate/generate_report_test.go
```diff
@@ -250,10 +251,33 @@ func assertReportHasError(t *testing.T, report generate.Report, err error) {
...
+
+	if diff := cmp.Diff(got, want, cmp.Comparer(compareErrors)); diff != "" {
 		t.Errorf("got %s", got)
 		t.Errorf("want %s", want)
 		t.Errorf("got(-) want(+)")
 		t.Fatal(diff)
 	}
 }
+
+func compareErrors(err, err2 error) bool {
```
i4ki, line 263: not nice to have this duplication just for the diff.
I think it'd be better to call cmp.Diff() on the individual fields of the generate.Report but call errtest.Assert() for the errors.

i4ki (changes requested):
lgtm but I think the slash check is not portable (we don't support windows yet but let's not introduce even more issues).
If it's not the case, sorry, feel free to disregard it.

generate/generate_report_test.go
katcipis: I was in the fence between that and just providing a cmp function, it got more involved than what I initially thought it would be.

katcipis: I changed it to use the individual fields comparison, seems to be working well. The tradeoff is that for a complex structure like the report it is not extremely simple + as the report evolves it would be easy to end up not comparing the report properly, forgetting fields, etc. Just to ellaborate to why I was on the fence + for me it is not as clear that the field comparison will be better and the alternative was "not nice"...neither are very nice IMHO :laughing:

generate/generate.go
katcipis: Will do + add tests for this

katcipis: hmm actually filepath.ToSlash introduces the same problem, when using `\` on linux it will not replace it with slash:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/path/filepath/path.go;l=167

katcipis: OK gave it a shot by checking for both separators with Contains. Funny enough, using filepath.Separator would not work at all since on Windows `/` is also a separator:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/os/path_windows.go;l=13;bpv=1;bpt=0

So we would allow `/` on windows but that would be interpreted as a valid separator/dir for Windows (apparently, Im quite outdated on my windows skills :laughing: )

i4ki (approved):
lgtm

i4ki (approved):
LGTM :rocket:

generate/generate.go
```diff
@@ -543,6 +553,17 @@ func checkGeneratedFilesConflicts(genfiles []genfile) error {
 	return nil
 }
 
+func checkGeneratedFilesPaths(genfiles []genfile) error {
+	for _, gen := range genfiles {
+		filename := filepath.ToSlash(gen.name)
+		if strings.Contains(filename, `/`) {
```
katcipis, line 559:
```suggestion
		if strings.Contains(filename, "/") {
```

generate/generate_report_test.go
```diff
@@ -250,10 +252,27 @@ func assertReportHasError(t *testing.T, report generate.Report, err error) {
 
 func assertEqualReports(t *testing.T, got, want generate.Report) {
 	t.Helper()
-	if diff := cmp.Diff(got, want); diff != "" {
-		t.Errorf("got %s", got)
-		t.Errorf("want %s", want)
-		t.Errorf("got(-) want(+)")
+
+	// WHY: we can't just use cmp.Diff since the errors included on the Report
```
katcipis, line 256: @kassianh this should solve the problem you were having testing generate file :smiley_cat:

PR #310, opened 2022-04-20
draft

katcipis:


katcipis:
We can't have parallel tests when using the hashicorp HCL library because the race detector finds a race condition on the shared nil token used here: 

* https://github.com/hashicorp/hcl/blob/2ef09d129d968e8c6e7df29f9af103025eafd61a/hclwrite/format.go#L11
* https://github.com/hashicorp/hcl/blob/2ef09d129d968e8c6e7df29f9af103025eafd61a/hclwrite/format.go#L122

This may also affect our ability to parallelize things on Terramate itself, if the code path depends on format (building an hcl doc using the lib ends up using the format.go, in this case at least).

PR #316, opened 2022-04-22
refactor: use Go's unwrap/Is protocol

katcipis:
So we can focus on implementing just the Is method specific for each error type, no need to rollout our own Is implementation/recursion/unwrapping logic. Will make it easier for us to add the error list type.

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -326,21 +326,6 @@ func As(err error, target interface{}) bool {
 	return errors.As(err, target)
 }
 
-func hasDesc(err error, desc string) bool {
-	e, ok := err.(*Error)
-	if !ok {
-		return false
-	}
-
-	if e.Description == desc {
-		return true
-	}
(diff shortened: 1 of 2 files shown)


errors/error.go
```diff
@@ -261,112 +261,58 @@ func (e *Error) Detailed() string {
...
-	}
-	if equalStack(e.Stack, stack) {
-		return true
-	}
-	if e.Err != nil {
-		return hasSameStack(e.Err, stack)
-	}
-	return false
+// Is is just an alias to Go stdlib errors.Is
+func Is(err, target error) bool {
```
katcipis, line 309: these could be useful to avoid importing two errors package, but not 100% sure they are a good idea. WDYT @i4ki ?

i4ki: yeah It makes sense!

i4ki (approved):
lgtm

PR #318, opened 2022-04-22
chore: add errors.List

katcipis:
Add new errors.List type + constructor errors.L. Should be enough to represent errors list without giving up using the error interface. Usage will be:

```go
errs := errors.L()
errs.Append(something.Do())
return errs.AsError()
```

It will automatically ignore nil error (sucesses) and if no errors are present errs.AsError() returns nil. Using errors.Is will also work by matching with all errors inside the list, any match is enough for success.

To inspect all errors contained on the LSP server there is a method List.Errors that returns all errors that are of the type Error or have an Error wrapped inside them, so we can use it like this:

```go
for _, err := range errs.Errors() {
    // err type is errors.Error, so we can just acccess information there
}
```

Not in love with any of the names I chose, so feel free to propose different names for anything :laughing:

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -45,6 +45,15 @@ type Error struct {
 	Err error
 }
 
+// Errors represents a list of error instances that also
+// implements Go's error interface.
+// Errors implements Go's errors.Is protocol matching the
+// target error with all the errors inside it, returning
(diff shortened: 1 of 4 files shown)


i4ki (approved):
LGTM

PR #319, opened 2022-04-23
fix: ignore nl and comments in the needsEval check.

i4ki:


hcl/eval/partial.go
```diff
@@ -1328,6 +1330,16 @@ func trimIgnored(tokens hclwrite.Tokens) hclwrite.Tokens {
 	return tokens[:ignorePos]
 }
 
+func ignorenlc(tokens hclwrite.Tokens) hclwrite.Tokens {
```
katcipis, line 1333: I have a vague memory that we had some sort of filter function that would accept a list of tokens that it would filter them out...but I'm probably wrong :laughing:

katcipis (approved):
:rocket:

PR #321, opened 2022-04-25
feat: add multiple errs when parsing hcl.Modules

katcipis:
# Reason for This Change

So we can provide better error reporting when supporting LSP integration.

## Description of Changes

Parsing modules now return an error list instead of a single error, and won't stop as soon as it finds an error.

### DIFF
--- a/errors/error_test.go
+++ b/errors/error_test.go
@@ -302,9 +302,14 @@ func TestErrorIs(t *testing.T) {
 		},
 		{
 			name:    "same file range",
-			err:     E("error", hcl.Range{Filename: "test.hcl"}),
-			target:  E("error", hcl.Range{Filename: "test.hcl"}),
+			err:     E("error", filerange),
+			target:  E("error", filerange),
 			areSame: true,
+		},
+		{
+			name:   "different file ranges",
+			err:    E("error", filerange),
+			target: E("error", otherFileRange),
 		},
 		{
 			name:    "error match wrapped on stderr",
--- a/generate/generate_report_test.go
+++ b/generate/generate_report_test.go
@@ -21,7 +21,6 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/madlambda/spells/assert"
 	"github.com/mineiros-io/terramate/generate"
-	errtest "github.com/mineiros-io/terramate/test/errors"
 )
 
 func TestReportRepresentation(t *testing.T) {
@@ -256,7 +255,7 @@ func assertEqualReports(t *testing.T, got, want generate.Report) {
 	// WHY: we can't just use cmp.Diff since the errors included on the Report
 	// are not comparable and may contain unexported fields (depending on how errors are built)
 
-	errtest.Assert(t, got.BootstrapErr, want.BootstrapErr)
+	assert.IsError(t, got.BootstrapErr, want.BootstrapErr)
 
 	if diff := cmp.Diff(got.Successes, want.Successes); diff != "" {
 		t.Errorf("success results differs: got(-) want(+)")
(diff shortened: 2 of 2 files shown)


errors/list.go
```diff
@@ -77,7 +77,13 @@ func (l *List) Detailed() string {
 	}
 	details := []string{"error list:"}
 	for _, err := range l.errs {
-		details = append(details, "\t-"+err.Error())
+		var errmsg string
+		if e, ok := err.(interface{ Detailed() string }); ok {
```
katcipis, line 81: quite important to reverse engineer the byte positions on the ranges :laughing:

errors/error.go
```diff
@@ -252,7 +252,13 @@ func (e *Error) error(verbose bool) string {
 			}
 		case error:
 			if v != nil {
-				errParts = append(errParts, v.Error())
+				errmsg := ""
+				if e, ok := v.(interface{ Detailed() string }); ok && verbose {
```
katcipis, line 256: Important to debug an error.Error that has a list inside it, so we can see the list in detail too

hcl/hcl_test.go
```diff
@@ -152,20 +153,79 @@ module "test" {
...
-				}
-			}
+			addFilenameToErrorsFileRanges(wantErrs, tfpath)
 
 			modules, err := hcl.ParseModules(tfpath)
-			errtest.Assert(t, err, tc.want.err)
-			assert.EqualInts(t, len(tc.want.modules), len(modules), "modules len mismatch")
+			errtest.AssertIsErrors(t, err, wantErrs)
+			if err != nil {
+				errtest.AssertAsErrorsList(t, err)
```
katcipis, line 223: Since it is important that we export an error list on parsing errors I thought it would be a good idea to always ensure that errors returned by hcl.ParseModules will work when calling errors.As() with an errors.List, so it should always work when using on the LSP server (and the same would be done for other parse functions). But checking this is not strictly necessary since we always check errors with errors.Is and that will also iterate errors inside the lists.

errors/list_test.go
```diff
@@ -150,6 +150,6 @@ func TestErrorListStringDetailedPresentation(t *testing.T) {
 	errs := errors.L(E("one"))
 	assert.EqualStrings(t, "error list:\n\t-one", errs.Detailed())
 
-	errs.Append(E("two"))
+	errs.Append(stderrors.New("two"))
```
katcipis, line 153: improving coverage for printing detailed VS normal errors

mariux:
can we use this to also to annotate issues in PRs in GitHub? should be possible? maybe? ;) lgtm

katcipis:
> can we use this to also to annotate issues in PRs in GitHub? should be possible? maybe? ;) lgtm

hmm I don't think I get the idea :thinking: . You mean the output of terramate itself would be parsed and then use that information to publish things on a PR ? (like a comment with nice error information or something ?)

Sorry if I'm completely off :laughing:

hcl/hcl_test.go
```diff
@@ -50,6 +50,7 @@ func TestHCLParserModules(t *testing.T) {
 	type want struct {
 		modules []hcl.Module
 		err     error
```
i4ki, line 52: I think this should be removed and always treat as returning multiple errors. No?

i4ki (approved):
lgtm but left a minor comment in tests. We can improve later if makes sense.

hcl/hcl_test.go
katcipis: yeah but I didn't wanted to change all the tests also on this PR so we could focus more on the actual changes, will do in a follow up PR.

PR #322, opened 2022-04-26
chore: flatten diagnostics + error list when appending on an error list

katcipis:
So it is easier to build error list from operations that produce a diagnostic from HCL (which is also a list) or even our own list. It seems that the most desireable behavior is flattening, but I could be wong.

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -131,6 +131,11 @@ func E(args ...interface{}) *Error {
 			if diag.Subject != nil {
 				e.FileRange = *diag.Subject
 			}
+		case *hcl.Diagnostic:
+			diag = arg
+			if diag.Subject != nil {
+				e.FileRange = *diag.Subject
+			}
 		case StackMeta:
 			e.Stack = arg
 		case error:
--- a/errors/error.go
+++ b/errors/error.go
@@ -75,9 +75,12 @@ const separator = ": "
 //		The underlying error that triggered this one.
 //	hcl.Diagnostics
 //		The underlying hcl error that triggered this one.
+//		Only the first hcl.Diagnostic will be used.
 //		If hcl.Range is not set, the diagnostic subject range is pulled.
 //		If the string Description is not set, the diagnostic detail field is
 //		pulled.
+//	hcl.Diagnostic
+//		Same behavior as hcl.Diagnostics but for a single diagnostic.
 //	string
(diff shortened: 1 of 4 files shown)


errors/error.go
```diff
@@ -116,13 +119,23 @@ func E(args ...interface{}) error {
 		case hcl.Range:
 			e.FileRange = arg
 		case hcl.Diagnostics:
-			diags = arg
+			diags := arg
 			if diags.HasErrors() {
```
i4ki, line 123: I just realized that `HasErrors()` looks for error severity, then if there are also Warnings just getting the diags[0] will get the wrong error. Not related to your change but have to be fixed.

errors/error.go
```diff
@@ -96,13 +99,13 @@ const separator = ": "
 //
 // In order to avoid duplicated messages, we erase the fields present in the
 // underlying error if already set with same value in this error.
-func E(args ...interface{}) error {
+func E(args ...interface{}) *Error {
```
i4ki, line 102: curious why this change is needed.

errors/list.go
```diff
@@ -84,11 +92,30 @@ func (l *List) Detailed() string {
...
+//
+// Any error of type errors.List will be flattened inside
+// the error list.
 func (l *List) Append(err error) {
 	if err == nil {
 		return
 	}
-	l.errs = append(l.errs, err)
+	switch e := err.(type) {
+	case hcl.Diagnostics:
```
i4ki, line 106: :thinking: 
strange this also here.
Why not use like `errs.Append(errors.E(diags))` ?

i4ki (approved):
LGTM but left some questions.
The issue with `HasErrors()` and accessing `diags[0]` can be fixed in a separate PR, but it's your call.

errors/error.go
katcipis: I was comparing with the result of List.Errors and wanted to build Error instances, I could also have used the type directly instead of errors.E(), but then the other question would be, why keep it as error ? When a type implements an interface it is rare to return the interface type itself on the constructor, you usually just return the concrete type and that will work out fine anyway when it is used as the interface. So I changed more because I don't see strong reasons for it to be error, it seems to enforce a constraint that is not necessary.

katcipis:
I will merge this so we can see it in action on a PR Im already working on, then we can keep refining it and fix the HasErrors thing.

PR #324, opened 2022-04-27
docs: add genfile initial docs/spec

katcipis:
Initial spec for the new generate_file block.

### DIFF
--- a/docs/codegen/generate-file.md
+++ b/docs/codegen/generate-file.md
@@ -48,7 +48,7 @@ whatever text format you want, here is a global reference:
 
 a = ${global.reference}
 
-and now a metadata reference:
+and a metadata reference:
 
 b = ${terramate.path}
 EOT
--- a/docs/codegen/generate-file.md
+++ b/docs/codegen/generate-file.md
@@ -63,4 +63,4 @@ or even the project root, which then generates the file for multiple stacks.
 
 There is no overriding or merging behavior for `generate_file` blocks.
 Blocks defined at different levels with the same label aren't allowed, resulting
-in failure for the overall code generation process.
+in failure of the overall code generation process.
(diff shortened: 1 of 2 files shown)


docs/codegen/generate-file.md
```diff
@@ -0,0 +1,66 @@
...
+* Terramate Metadata references
+* Terramate function calls
+* Expressions using string interpolation
+
+The final evaluated **content** attribute **must** be a string.
+
+Multiple `generate_file` blocks with the same label/filename will
+result in an error.
+
+Here is an example on how a JSON file can be created:
```
soerenmartius, line 25:
```suggestion
# File Generation

Terramate supports the generation of arbitrary text files referencing 
[Terramate defined data](../sharing-data.md).

File generation is done using `generate_file`
blocks in [Terramate configuration files](../config-overview.md).

The block **must** have a single label, that will be used to determine the
name of the generated file. Inside the block, the **content** attributes define
the string that will be written on the file.

The definition of the **content** attribute may include:

* Terramate Global references
* Terramate Metadata references
* Terramate function calls
* Expressions using string interpolation

The final evaluated **content** attribute **must** be a string.

Multiple `generate_file` blocks with the same label/filename will
result in an error.

Here is an example of how a JSON file can be created:
```

docs/codegen/overview.md
```diff
@@ -4,5 +4,7 @@ Code generation is the main way you can glue [Terramate data](../sharing-data.md
 to other tools, like Terraform.
 
 Different code generation strategies will be provided in the future to support
-each integration scenario the best way possible. Currently we support
-[HCL generation](./generate-hcl.md).
+each integration scenario the best way possible. Currently we support:
```
soerenmartius, line 7:
```suggestion
each integration scenario in the best way possible. Currently, we support:
```

soerenmartius (changes requested):

soerenmartius (approved):
lgtm! Good job!

i4ki (approved):
LGTM :rocket:

mariux (approved):
lgtm

PR #325, opened 2022-04-27
refactor: move globals to stack pkg

katcipis:
We already had issues with cyclic dependencies multiple times because globals are on the main terramate pkg. I think it makes sense for it to be on the stack, so we have stack.Metadata + stack.Globals, seems to map the domain well. We can improve things more on the future, I focused only on moving it to the package.


i4ki (approved):
lgtm

PR #332, opened 2022-04-29
feat: add generate_file

katcipis:
Adds generate_file to Terramate by integrating it to the overall code generation. This is more complex/bigger than expected because we can't handle detection/headers for generic file generation the same way we did for HCL generation. This is definitely not the better/final design, the current generate approach was not designed with this in mind,we were thinking on generate_hcl + generate_terraform at the time, both would be HCL, so I hacked through the current design.

We will need to address outdated detection soon and that may change up the design anyway, so I just cutted some corners on this for now.

Outdated detection for generate_hcl remains the same, but for generate_file it will not work the same since we can't track files for the lack of headers. We will still detect changes when possible, but when changing/removing files it won't detect.

Protection against overwriting user files also doesn't work for generate_file, since there is no header to check we can't detect if it is a user file or a terramate file, so generate_file for now will overwrite any file that matches the label.

We do have some alternatives on how to improve that, discussed with @i4ki, but we are going to leave this for a future PR since we still need to refine it and decide on possible options (prefixing files, maybe using git diffs, etc).

Internal task: https://app.shortcut.com/mineiros/story/5092/add-support-for-generate-file


generate/generate_file_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/mineiros-io/terramate/errors"
+	"github.com/mineiros-io/terramate/generate"
+)
+
+func TestGenerateFile(t *testing.T) {
+	checkGenFiles := func(t *testing.T, got string, want string) {
```
i4ki, line 27: as `t` is a pointer, passing it as a parameter doesn't help much, then using it as a closure variable saves a few characters when calling checkGenFiles()... it's a minor change then up to you.

i4ki (dismissed):
LGTM but found some odd refactor changes. I think the `hcldoc()` alike helpers (`str()`, `boolean()`, etc), were very nice but because of a clash with a `type str string` you renamed all the attribute ones to `strAttr()`, `boolAttr()` and so on, leading to a bigger change (at least in the test abstraction).
I think renaming the stringer `str` to another thing would be a smaller change and keep the gen abstraction nicer (as it was).
But it's not a big problem.
Another thing is that now the `generate` package still knows a lot of `hcl` stuff, but we can improve this later. Just commenting here now so we can remember.

katcipis:
> LGTM but found some odd refactor changes. I think the `hcldoc()` alike helpers (`str()`, `boolean()`, etc), were very nice but because of a clash with a `type str string` you renamed all the attribute ones to `strAttr()`, `boolAttr()` and so on, leading to a bigger change (at least in the test abstraction). I think renaming the stringer `str` to another thing would be a smaller change and keep the gen abstraction nicer (as it was). But it's not a big problem.

Was on the fence on this change, we can definitely change it back, no strong opinions.

> Another thing is that now the `generate` package still knows a lot of `hcl` stuff, but we can improve this later. Just commenting here now so we can remember.

Yeah extracting HCL specific knowledge from generate has been done incrementally and we are not done yet for sure x_x, in the beginning everything was quite all together, like the backend config stuff...but I started to realize that it would make more sense to isolate etc etc.

generate/generate.go
```diff
@@ -257,19 +246,98 @@ func CheckStack(root string, st stack.S) ([]string, error) {
...
+		} else {
+			logger.Trace().Msg("generate_file code is updated")
+			outdatedFiles.remove(filename)
+		}
+	}
+
+	return nil
 }
 
 func updateGenHCLOutdatedFiles(
```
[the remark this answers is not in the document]
katcipis: I think it does, the whole movement to extracting things better has been quite incremental and is definitely not done yet. Do you prefer to see it done right away or in a separete PR ?

generate/generate_file_test.go
katcipis: Can't do that in this case, subtests are created and if a subtest calls Fail on a parent test *testing.T Go panics since that is not allowed (basically I need the subtests to pass their own testing instance pointers).

generate/generate.go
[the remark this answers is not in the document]
i4ki: it could be done separated. not a big deal.

generate/generate_file_test.go
i4ki: ah they're not the same pointer, each subtest has its own *testing.T. gotcha.

i4ki (approved):

PR #333, opened 2022-05-03
fix: avoid unnecessary git checks

katcipis:
Currently on prj.SetDefaults we are doing git operations that makes terramate fail even for commands that we don't need any git functionality, like terramate generate/list. The idea is to do git configuration + checks closer to the commands that actually need it.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -810,7 +810,7 @@ func (c *cli) runOnStacks() {
 		log.Fatal().
 			Str("action", "runOnStacks()").
 			Err(err).
-			Msgf("configuring git")
+			Msg("configuring git")
 	}
 
 	if err := c.prj.checkLocalDefaultIsUpdated(); err != nil {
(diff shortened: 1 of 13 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -319,27 +319,6 @@ func (c *cli) run() {
 
 	c.checkVersion()
 
-	if c.parsedArgs.Changed {
```
katcipis, line 322: Here we moved the logic for checking git closer to the actual features, although this was not necessarily wrong (change detection needs git)

cmd/terramate/cli/cli.go
```diff
@@ -814,6 +806,19 @@ func (c *cli) runOnStacks() {
 		Str("workingDir", c.wd()).
 		Logger()
 
+	if err := c.prj.configureGit(c.parsedArgs); err != nil {
+		log.Fatal().
+			Str("action", "runOnStacks()").
+			Err(err).
+			Msgf("configuring git")
```
katcipis, line 813:
```suggestion
			Msg("configuring git")
```

PR #334, opened 2022-05-03
fix: check local default is updated

katcipis:
The check for local default is updated is not working properly, this PR fixes that.


i4ki (approved):
lgtm

PR #348, opened 2022-05-13
fix: code generation does't ignore dotfiles

katcipis:
Currently code generation will read dotfiles, but those should be ignored.


i4ki:
sorry: we need to get rid of the codegen/hcl duplication ASAP.

katcipis:
> sorry: we need to get rid of the codegen/hcl duplication ASAP.

Yeah lets introduce some improvements on this soon :rocket:

PR #349, opened 2022-05-16
Add logos for dark and light mode

soerenmartius:
This should in theory render different logos for dark and light mode as described in https://stackoverflow.com/questions/65413712/changing-readme-md-image-display-conditional-to-github-light-mode-dark-mode

PR #350, opened 2022-05-16
chore: add dark mode logo

katcipis:
Just playing around to help on #349 

It would be better to work with SVG since that enables rescaling the page without losing quality on the logos. But I haven't found a way to use markdown syntax + enforce a size on the SVG. Using the `` HTML tag doesn't work for the light/dark mode logic on GH.


katcipis:
Since I was unable to find a way to keep the SVGs + use the GH dark mode thing Im closing this. We can re-assess in the future.

More info can be found here: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to

GH only support markdown images with dark/light theme annotations, we can't use ``, then we can't use SVGs and determina a size very easily :-(

katcipis:
Re opened since it may be useul in the future.

PR #351, opened 2022-05-16
feat: add terramate fmt

katcipis:
Adds the `terramate fmt` command. It will by default format all files inside the current working dir + all subdirs recursively (the -recursive flag on terraform fmt). We can also run `terramate fmt --check`, in that case it only checks if there are any files to be formatted and if there is any exits with 1, 0 otherwise.

Both on `terramate fmt` and `terramate fmt --check` the list of the files changed will be sent to stdout.

Also changed all references to hclwrite.Format to use our own hcl.Format function, with some further changes on testing code since our hcl.Format only formats valid code, not anything =P

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -79,11 +79,12 @@ type cliSpec struct {
 	} `cmd:"" help:"List stacks"`
 
 	Run struct {
-		DisableCheckGenCode bool     `optional:"true" default:"false" help:"Disable outdated generated code check"`
-		ContinueOnError     bool     `default:"false" help:"Continue executing in other stacks in case of error"`
-		DryRun              bool     `default:"false" help:"Plan the execution but do not execute it"`
-		Reverse             bool     `default:"false" help:"Reverse the order of execution"`
-		Command             []string `arg:"" name:"cmd" passthrough:"" help:"Command to execute"`
+		DisableCheckGenCode   bool     `optional:"true" default:"false" help:"Disable outdated generated code check"`
+		DisableCheckGitRemote bool     `optional:"true" default:"false" help:"Disable checking if local default branch is updated with remote"`
+		ContinueOnError       bool     `default:"false" help:"Continue executing in other stacks in case of error"`
+		DryRun                bool     `default:"false" help:"Plan the execution but do not execute it"`
+		Reverse               bool     `default:"false" help:"Reverse the order of execution"`
+		Command               []string `arg:"" name:"cmd" passthrough:"" help:"Command to execute"`
 	} `cmd:"" help:"Run command in the stacks"`
 
 	Generate struct{} `cmd:"" help:"Generate terraform code for stacks"`
@@ -266,8 +267,7 @@ func newCLI(args []string, stdin io.Reader, stdout, stderr io.Writer) *cli {
 			Msg("EvalSymlinks() failed")
 	}
 
-	logger.Trace().
-		Msg("Running in directory")
+	logger.Trace().Msg("Running in directory")
 
 	prj, foundRoot, err := lookupProject(wd)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


hcl/fmt_test.go
```diff
@@ -0,0 +1,224 @@
...
+		name     string
+		input    string
+		want     string
+		wantErrs []error
+	}
+
+	tcases := []testcase{
+		{
+			name: "attributes alignment",
+			input: `
```
katcipis, line 41: when we start adding extra formatting features the idea is to add more test cases here

hcl/eval/partial_fuzz_test.go
```diff
@@ -92,13 +90,9 @@ func FuzzPartialEval(f *testing.F) {
 
 		const testattr = "attr"
 
-		cfg := hcldoc(
```
katcipis, line 95: Now hclwriter always produces valid HCL, so if you use it to build invalid HCL files it panics, seemed like the way to go since it is what we want on 99% of the tests, we use it to build some valid HCL file for testing. But here we generate possibly invalid HCLs and that would cause panics, so it seemed easier to just create the HCL with a string since it is a pretty straithforward attr = val.

test/hclwrite/hclwrite.go
```diff
@@ -222,7 +222,11 @@ func NumberInt(name string, val int64) BlockBuilder {
 
 // Format formats the given HCL code.
 func Format(code string) string {
-	return strings.Trim(string(hclwrite.Format([]byte(code))), "\n ")
+	formatted, err := hcl.Format(code, "gen.hcl")
+	if err != nil {
```
katcipis, line 226: One alternative to assuming that only valid HCL is allowed when using the hclwrite builders would be to just skip formatting if the code is not valid HCL (or maybe falling back to hclwrite.Format that formats anything). But the alternative seems clumsy, overall I think we want to build valid HCL using the builders, so if because a bug it is not valid HCL it is desireable for the tests to fail fast.

cmd/terramate/e2etests/fmt_test.go
```diff
@@ -0,0 +1,104 @@
...
+		sprintf("f:another-stacks/stack-1/globals.tm.hcl:%s", unformattedHCL),
+		sprintf("f:another-stacks/stack-2/globals.tm.hcl:%s", unformattedHCL),
+		sprintf("f:stacks/globals.tm:%s", unformattedHCL),
+		sprintf("f:stacks/stack-1/globals.tm:%s", unformattedHCL),
+		sprintf("f:stacks/stack-2/globals.tm:%s", unformattedHCL),
+	})
+
+	wantedFiles := []string{
+		"globals.tm",
+		"another-stacks/globals.tm.hcl",
```
i4ki, line 62: maybe add a test with `--chdir` into the generated tree to check if the file paths are expected?

hcl/fmt.go
```diff
@@ -0,0 +1,153 @@
...
+	return f.path
+}
+
+// Formatted is the contents of the original file after formatting.
+func (f FormatResult) Formatted() string {
+	return f.formatted
+}
+
+const (
+	errFormatTree errors.Kind = "formatting tree"
```
i4ki, line 152: we were placing this at the top in all other packages. Just curious why in the bottom in this one.

i4ki (approved):
lgtm but left minor ideas/questions.

cmd/terramate/cli/cli.go
```diff
@@ -465,6 +471,52 @@ func (c *cli) listStacks(mgr *terramate.Manager, isChanged bool) (*terramate.Sta
...
+
+	logger.Trace().Msg("formatting all files recursively")
+	results, err := hcl.FormatTree(c.wd())
+	if err != nil {
+		logger.Fatal().Err(err).Msg("formatting files")
+	}
+
+	logger.Trace().Msg("listing formatted files")
+	for _, res := range results {
+		path := strings.TrimPrefix(res.Path(), c.wd()+"/")
```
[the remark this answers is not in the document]
katcipis: Using filepath.Separator. Maybe we could use `prj.FriendlyFmtDir()` but I'm not so sure :thinking:

hcl/fmt.go
katcipis: By default I never placed any private symbols/constants at the top =P. I'm just used to first listing all public things and then later all private things. But can also put on the top for consistency (not a strong opinion).

katcipis: We are adding on the top all exported error sentinels/kinds etc. But this is just my default mode of organizing things, very personal, but I do make a distinction between a private vs public error sentinel, the private one can't even be used by caller for any checking, it is just to avoid duplication internally.

i4ki: ah ok, all the others are public indeed.

PR #352, opened 2022-05-18
fix: errors.AssertIsErrors should check if no err is wanted

katcipis:
Currently when we call `errors.AssertIsErrors(err, nil)` if the err is not nil it will pass, which is wrong since passing nil/empty target error list indicates that you want no errors at all (just as passing nil on errors.Is() also fails unless err is also nil).


i4ki (approved):
LGTM

PR #357, opened 2022-05-24
feat: add support for multiple errors per file.

i4ki:
This PR adds support for multiple errors per configuration file.

### DIFF
--- a/errors/error.go
+++ b/errors/error.go
@@ -73,7 +73,7 @@ const separator = ": "
 //		The stack that originated the error.
 //  *List
 //		The underlying error list. By wrapping an error list, you also wrap all
-//		of its error items.//
+//		of its error items.
 //	error
 //		The underlying error that triggered this one.
 //	hcl.Diagnostics
--- a/errors/error.go
+++ b/errors/error.go
@@ -175,16 +175,30 @@ func E(args ...interface{}) *Error {
 
 	errs, ok := e.Err.(*List)
 	if ok {
-		newargs := []interface{}{}
+		// if the underlying error is a *List we wrap all of its elements so they
+		// carry all the context needed to print them individually.
+		// Eg.:
+		//   errs := errors.L()
+		//   obj, err := something.Do()
+		//   if err != nil {
+		// 	 	errs.Append(errors.E(ErrSomethingBadHappened, err))
+		//   }
+		//
+		//   if `err` is an *errors.List, the code above means that all of their
+		//   error items have the kind `ErrSomethingBadHappened`.
+		//
+
+		// code below captures all arguments but the *List
+		wrappingArgs := []interface{}{}
 		for _, arg := range args {
 			_, ok := arg.(*List)
 			if !ok {
-				newargs = append(newargs, arg)
+				wrappingArgs = append(wrappingArgs, arg)
 			}
 		}
 
 		for i, el := range errs.errs {
-			errs.errs[i] = E(append(newargs, el)...)
+			errs.errs[i] = E(append(wrappingArgs, el)...)
 		}
 
 		return e
--- a/errors/error.go
+++ b/errors/error.go
@@ -325,8 +325,9 @@ func (e *Error) Detailed() string {
 // If it's underlying error is a *List, then it just returns it because
 // they're already explicitly wrapped.
 func (e *Error) AsList() *List {
-	if errs, ok := e.Err.(*List); ok {
-		return errs
+	var el *List
+	if errors.As(e, &el) {
+		return el
 	}
 
 	return L(e)
--- a/errors/error.go
+++ b/errors/error.go
@@ -72,8 +72,8 @@ const separator = ": "
 //	errors.StackMeta
 //		The stack that originated the error.
 //  *List
-//		The underlying error list. By wrapping an error list, you also wrap all
-//		of its error items.
+//		The underlying error list. In this case we wrap all of its individual
+//		errors so they carry all the context to print them individually.
 //	error
 //		The underlying error that triggered this one.
 //	hcl.Diagnostics
--- a/errors/error.go
+++ b/errors/error.go
@@ -69,8 +69,11 @@ const separator = ": "
 //		The kind of error (eg.: HCLSyntax, TerramateSchema, etc).
 //	hcl.Range
 //		The file range where the error originated.
-//		errors.StackMeta
+//	errors.StackMeta
 //		The stack that originated the error.
+//  *List
+//		The underlying error list. By wrapping an error list, you also wrap all
+//		of its error items.//
 //	error
 //		The underlying error that triggered this one.
 //	hcl.Diagnostics
--- a/errors/error.go
+++ b/errors/error.go
@@ -305,7 +305,7 @@ func (e *Error) Detailed() string {
 
 // AsList returns the error as a list.
 // If it's underlying error is a *List, then it just returns it because
-// they're already explictly wrapped.
+// they're already explicitly wrapped.
 func (e *Error) AsList() *List {
 	if errs, ok := e.Err.(*List); ok {
 		return errs
--- a/errors/error.go
+++ b/errors/error.go
@@ -206,15 +206,9 @@ func (e *Error) isEmpty() bool {
 	return e.FileRange == hcl.Range{} && e.Kind == "" && e.Description == "" && e.Stack == nil
 }
 
-func (e *Error) error(verbose bool) string {
+func (e *Error) error(members []interface{}, verbose bool) string {
 	var errParts []string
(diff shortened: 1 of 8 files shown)


errors/list.go
```diff
@@ -62,14 +62,21 @@ func (l *List) Error() string {
...
+// Errors returns all errors contained on the list.
+// It flattens out the wrapped error lists inside *error.Error.
+func (l *List) Errors() []error {
+	var errs []error
 	for _, err := range l.errs {
-		var e *Error
-		if errors.As(err, &e) {
+		switch e := err.(type) {
+		case *Error:
+			if el, ok := e.Err.(*List); ok {
```
katcipis, line 72: I'm curious with removing errors.As and using the access directly to the wrapped error instead. We want to always force that the depth must be 1 ? I'm not getting why errors.As was dropped here and in other places 🤔. It will unwrap and get the list when it is wrapped on the error itself + it will recursively search for the list deeper if necessary through unwrapping, what do we gain by not allowing further unwrapping ?

errors/list_test.go
```diff
@@ -33,21 +33,26 @@ func TestEmptyErrorListReturnsEmptyErrors(t *testing.T) {
 }
 
 func TestErrorListReturnsAllErrors(t *testing.T) {
+	// This test was updated to reflect the changed API of errors.List.Errors().
+	// Now it doesn't ignore errors anymore.
+
```
katcipis, line 38:
```suggestion
```

errors/list_test.go
```diff
@@ -33,21 +33,26 @@ func TestEmptyErrorListReturnsEmptyErrors(t *testing.T) {
...
 	e := errors.L()
 
 	assert.EqualInts(t, 0, len(e.Errors()))
 
+	notIgnored := stderrors.New("ignored")
```
katcipis, line 43:
```suggestion
	notIgnored := stderrors.New("not ignored")
```

errors/list.go
```diff
@@ -104,23 +111,38 @@ func (l *List) Detailed() string {
 //
 // Any error of type errors.List will be flattened inside
 // the error list.
-func (l *List) Append(err error) {
-	if err == nil {
+func (l *List) Append(errs ...error) {
```
katcipis, line 114: Kinda curious when we append multiple errors, we usually call something that returns an error instance and then already call Append on it 🤔

hcl/hcl.go
```diff
@@ -567,36 +598,37 @@ func assignSet(name string, target *[]string, val cty.Value) error {
...
-		Msg("Iterate over values.")
+	errs := errors.L()
 	values := map[string]struct{}{}
 	iterator := val.ElementIterator()
 	for iterator.Next() {
 		_, elem := iterator.Element()
 
-		logger.Trace().
-			Msg("Check element is of correct type.")
+		logger.Trace().Msg("Check element is of correct type.")
```
katcipis, line 609: noice

errors/error.go
```diff
@@ -69,8 +69,11 @@ const separator = ": "
 //		The kind of error (eg.: HCLSyntax, TerramateSchema, etc).
 //	hcl.Range
 //		The file range where the error originated.
-//		errors.StackMeta
+//	errors.StackMeta
 //		The stack that originated the error.
+//  *List
+//		The underlying error list. By wrapping an error list, you also wrap all
```
katcipis, line 75: Maybe add some of the info from the internal comment to here, the docs here are a little confusing on what exactly will be the behavior, like what does "you also wrap all of its error items" means ? That doesn't make it clear to me that all args passed to E() will be propagated to each of the errors on the list.

errors/error.go
```diff
@@ -206,15 +242,9 @@ func (e *Error) isEmpty() bool {
 	return e.FileRange == hcl.Range{} && e.Kind == "" && e.Description == "" && e.Stack == nil
 }
 
-func (e *Error) error(verbose bool) string {
+func (e *Error) error(members []interface{}, verbose bool) string {
```
katcipis, line 245: If we go with defaultErrorFields, could also be defaultErrorMembers:

```suggestion
func (e *Error) error(fields[]interface{}, verbose bool) string {
```

katcipis (approved):
LGTM

My main source of confusion was dropping errors.As usage, but I imagine I'm missing something and it is not critical to the PR.

Also we are missing some tests involving multiple errors, but we can improve that later too.

errors/list_test.go
```diff
@@ -95,27 +100,21 @@ func TestErrorListFlattensAllDiagnostics(t *testing.T) {
...
-	wantErrs := []*errors.Error{
-		{
-			Description: detail1,
-			FileRange:   *range1,
-		},
-		{
-			Description: detail2,
-			FileRange:   *range2,
-		},
+	wantErrs := []error{
```
katcipis, line 115: noice 🚀

errors/error.go
```diff
@@ -270,14 +301,47 @@ func (e *Error) error(verbose bool) string {
...
 func (e *Error) Detailed() string {
-	return e.error(true)
+	return e.error(e.defaultErrorParts(), true)
+}
+
+// AsList returns the error as a list.
+// If it's underlying error is a *List, then it just returns it because
+// they're already explicitly wrapped.
+func (e *Error) AsList() *List {
+	if errs, ok := e.Err.(*List); ok {
```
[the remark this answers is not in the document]
i4ki: you're right.
I created this method to easily always handle errors as it it's a list, even if it's just one error, to avoid type switches or multiple `errors.As()` calls in downstream packages.

errors/error.go
```diff
@@ -270,14 +301,47 @@ func (e *Error) error(verbose bool) string {
...
+		return errs
+	}
+
+	return L(e)
+}
+
+// Message returns the error message without some metadata.
+// This method is suitable for editor extensions that needs to handle the
+// metadata themselves.
+func (e *Error) Message() string {
```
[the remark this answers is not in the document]
i4ki: yep I feel the same. The `Error()` method I kept with the filerange because this is probably the best default, like if you only print the error then it have everything. But it's also not so easy to customize the "printer" of the error...
if you have an idea let me know.

errors/list.go
i4ki: This is not actually needed anymore but I think it makes sense to support this.

1. it's intuitive, in the same way we can do `errs = append(errs, err1, err2, ...)`
2. when wrapping multiple lists of errors with different kinds.

```
    el1 := errors.L()
    for ... {
        err = ...
        el1.Append(err)
        ...
    }
    
    el2 := errors.L()
    for ... {
        err = ...
        el2.Append(err)
    }
    
    allerrs := errors.L()
    allerrs.Append(errors.E(SuchFail, el1), errors.E(SuchFail, el2))
    return allerrs.AsError()
```

but current code doesn't depend on this anymore so I can rollback to single parameter if you do prefer.

errors/list.go
i4ki: this was a brain fart =) I did this change for cosmetic purposes.. to avoid the if-elses and handle them in a switch but forgot that `errors.As()` also unwraps the error. I reverted this, thanks for catching.

i4ki:
@katcipis I added tests for multiple errors in the terramate-lsp PR because I thought it'd be a huge change in an already big PR, but actually it doesn't seem like it's going to be that bad.
I will at least add here the same tests I did there.

errors/error.go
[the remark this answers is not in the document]
katcipis: Hmm you already have a reasonably nice internal API to personalize which information will be present on the final message and in which order, but not sure we need that exposed :thinking:. All good for now, don't have better ideas yet =P.

errors/list.go
katcipis: Hmm it doesn't change the usage of the method so I suppose it is OK to leave with vargs since it can be useful on some scenarios, was just curious when that happens because I usually just keep appending errors as soon as they happen etc.

katcipis:
> @katcipis I added tests for multiple errors in the terramate-lsp PR because I thought it'd be a huge change in an already big PR, but actually it doesn't seem like it's going to be that bad. I will at least add here the same tests I did there.

Always like smaller PRs, thanks for doing it :heart:

hcl/hcl_test.go
```diff
@@ -1097,6 +1145,9 @@ func testParser(t *testing.T, tc testcase) {
 
 // some helpers to easy build file ranges.
 func mkrange(start, end hhcl.Pos) hhcl.Range {
+	if start.Byte == end.Byte {
```
katcipis, line 1148: Nice, I had some troubles with that but I missed the opportunity to improve this :heart:

katcipis (approved):
:rocket:

PR #365, opened 2022-05-27
chore: improve errors.E and errors.L relationship.

i4ki:


errors/list.go
```diff
@@ -122,18 +122,28 @@ func (l *List) Append(errs ...error) {
 
 		switch e := err.(type) {
 		case hcl.Diagnostics:
-			for _, diag := range e {
-				l.errs = append(l.errs, E(diag))
+			if e != nil {
```
i4ki, line 125: now that `err` is an interface, we need to explicitly check if the underlying value is nil.

katcipis: Instead of checking if the err is nil on all cases wouldn;t it be easier to just filter out all nil errors from errs and then iterate the result ?

errors/error.go
```diff
@@ -61,88 +61,93 @@ const separator = ": "
 
 // E builds an error value from its arguments.
 // There must be at least one argument or E panics.
-// The type of each argument determines its meaning. If more than one argument
-// of a given type is presented, only the last one is recorded.
+// The type of each argument determines its meaning.
+// Multiple underlying errors can be provided and in such case E() builds a *List
+// of errors as its underlying error.
+// If multiple arguments of same type is presented (and it's not an
```
katcipis, line 67: What is an "underlying error type" ?

errors/error.go
```diff
@@ -61,88 +61,93 @@ const separator = ": "
...
+//  The underlying error types are:
+//
 //  *List
-//		The underlying error list. In this case we wrap all of its individual
-//		errors so they carry all the context to print them individually.
-//	error
-//		The underlying error that triggered this one.
+//		The underlying error list wrapped by this one.
+// 		This error wraps all of its individual errors so they carry all the
+//      context to print them individually.
```
katcipis, line 87:
```suggestion
//		The underlying error list wrapped by this one.
// 		This error wraps all of its individual errors so they carry all the
//		context to print them individually.
```

errors/list.go
```diff
@@ -102,14 +102,13 @@ func (l *List) Detailed() string {
...
+// Append appends the provided errs on the error list, ignoring nil values.
 //
-// Any error of type hcl.Diagnostics will be flattened inside
-// the error list, each hcl.Diagnostic will become an error.Error.
+// Any error of type hcl.Diagnostics will have its hcl.Diagnostic elements added
+// to the error list.
 //
-// Any error of type errors.List will be flattened inside
-// the error list.
+// Any error of type *List will have its error elements added to the
```
katcipis, line 110: hmm for me it was quite clear that if I append a list on another list I would end up having a list of lists ( [ [err1, err2] ] ) and to avoid that and have a single list of errors ([ err1, err2]) what need to happen is flattening, so you always have a single list of errors. So not entirely sure how the new phrasing is better 🤔

katcipis (dismissed):

errors/list.go
i4ki: ok, it's because for me "flatten" means something different depending on the case... I just wanted to be explicit about what was happening here but probably the wording was good enough. will rollback.

katcipis: hmm I could be wrong on my take on this, what I get by flattening is this:

```
a = [1,2,3]
a.append([4,5])
```

a normal append behavior would produce:

```
[1,2,3, [4,5]]
```

An append that does flattening will produce this:

```
[1,2,3,4,5]
```

But maybe I'm using the term incorrectly here.

errors/list.go
katcipis: Ah OK I missed the casting and was thinking about just the errs slice, sorry for the noise

katcipis (approved):

errors/list.go
i4ki: yes, I got this. I just wasn't sure if it was clear enough but probably is.

i4ki: I reverted the change

PR #366, opened 2022-05-27
chore: add multi-file errors.

i4ki:


hcl/hcl_test.go
```diff
@@ -932,6 +1014,115 @@ func TestHCLParserStack(t *testing.T) {
...
+					filename: "stack1.tm",
+					body:     "stack {}",
+				},
+				{
+					filename: "stack2.tm",
+					body:     "stack {}",
+				},
+			},
+			want: want{
+				errs: []error{
```
katcipis, line 1080: Conflicting git config produces two errors but here only stack2.tm is reported on the error list ? Is this expected ?

hcl/hcl_test.go
```diff
@@ -1100,14 +1291,23 @@ func testParser(t *testing.T, tc testcase) {
...
+			test.WriteFile(t, configsDir, filename, inputConfigFile.body)
 		}
+		fixupFiledirOnErrorsFileRanges(configsDir, tc.want.errs)
 		got, err := hcl.ParseDir(configsDir)
 		errtest.AssertErrorList(t, err, tc.want.errs)
 
+		var gotErrs *errors.List
+		if errors.As(err, &gotErrs) {
+			if len(gotErrs.Errors()) != len(tc.want.errs) {
+				t.Logf("got errors: %s", gotErrs.Detailed())
```
katcipis, line 1305: Logging wanted errors would help debugging when this fails (instead of just got).

hcl/hcl_test.go
```diff
@@ -1100,14 +1291,23 @@ func testParser(t *testing.T, tc testcase) {
...
-			addFilenameToErrorsFileRanges(tc.want.errs, cfgfile)
+			test.WriteFile(t, configsDir, filename, inputConfigFile.body)
 		}
+		fixupFiledirOnErrorsFileRanges(configsDir, tc.want.errs)
 		got, err := hcl.ParseDir(configsDir)
 		errtest.AssertErrorList(t, err, tc.want.errs)
 
+		var gotErrs *errors.List
+		if errors.As(err, &gotErrs) {
+			if len(gotErrs.Errors()) != len(tc.want.errs) {
```
katcipis, line 1304: hmm maybe we want that on AssertErrorList ? The current behavior is partial matching but not sure this is what we should have as default behavior 🤔

katcipis (approved):

hcl/hcl_test.go
```diff
@@ -53,28 +53,39 @@ func TestHCLParserModules(t *testing.T) {
...
 		want  want
 	}
 
 	for _, tc := range []testcase{
 		{
-			name:  "module must have 1 label",
-			input: `module {}`,
+			name: "module must have 1 label",
+			input: cfgfile{
+				filename: "main.tf",
```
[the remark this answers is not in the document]
i4ki: it worked for some time, but now having multiple errors coming from different files it was very hard to maintain it.

katcipis (approved):

hcl/hcl_test.go
i4ki: the test called "conflicting terramate git config and other errors" has an additional error in it (the unknown `test` block) that's why it reports two errors.
Maybe I could separate this in another test case, but I added this to make sure conflicting errors still continue parsing the same file and generating other errors.

katcipis: Ah OK I misunderstood what was going on. All good :-)

hcl/hcl_test.go
i4ki: yeah I think makes sense to check this in the AssertErrorList. The problem is that I got a case where tests were passing but I wasn't checking all errors coming from the parser.
This is quite useful sometimes but in this case it made me anxious haha

katcipis (approved):

PR #367, opened 2022-05-31
chore: add specs for conditional code generation

katcipis:
Initial specs for conditional code generation on generate_file and generate_hcl.


docs/codegen/generate-file.md
```diff
@@ -64,3 +64,24 @@ or even the project root, which then generates the file for multiple stacks.
...
 in failure of the overall code generation process.
+
+
+## Conditional Code Generation
+
+Conditional code generation is achieved by the use of the `condition` attribute.
+The `condition` attribute should always evaluate to a boolean. If `condition` 
+is true, the code will be generated for the given stack. If `condition` is false,
+the code will **not** be generated for the given stack.
+
```
i4ki, line 75: maybe makes sense to state that if `condition` is not provided, then the code is always generated.. but not sure, it seems clear enough.

i4ki (dismissed):
lgtm

docs/codegen/generate-hcl.md
```diff
@@ -125,6 +125,31 @@ Blocks defined at different levels with the same label aren't allowed, resulting
...
+  condition = tm_length(global.list) > 0
+  content {
+    locals {
+      list = global.list
+    }
+  }
+}
+```
+
+Will only generate code for stacks that define the global `generate_file` to true.
```
mariux, line 150: i guess this is a mismatch of whatthe condition actually says?

mariux (changes requested):

docs/codegen/generate-hcl.md
i4ki: I think this is leftover of a change I suggested. @katcipis need to fix here.

PR #371, opened 2022-05-31
feat: add support to conditional code generation on genhcl/genfile

katcipis:
Adds support to the condition attribute when loading generate_hcl and generate_file on the  genhcl and genfile packages (specified on #367). The Loaded condition information is not integrated on the generate package yet, so the feature is not available to users on this PR. The follow up PR will integrate these changes on the generate package, which will use this information to not save files on disk (also used on outdated code detection, etc).


generate/genfile/genfile.go
```diff
@@ -28,12 +28,19 @@ import (
 
 const (
 	// ErrInvalidContentType indicates the content attribute on
-	// generate_file has a invalid type.
+	// has a invalid type.
```
i4ki, line 31:
```suggestion
	// has an invalid type.
```

generate/genfile/genfile.go
```diff
@@ -28,12 +28,19 @@ import (
...
 	ErrInvalidContentType errors.Kind = "invalid content type"
 
+	// ErrInvalidConditionType indicates the condition attribute on
+	// has a invalid type.
```
i4ki, line 35:
```suggestion
	// has an invalid type.
```

generate/genfile/genfile.go
```diff
@@ -28,12 +28,19 @@ import (
 
 const (
 	// ErrInvalidContentType indicates the content attribute on
```
i4ki, line 30:
```suggestion
	// ErrInvalidContentType indicates the content attribute
```

generate/genfile/genfile.go
```diff
@@ -28,12 +28,19 @@ import (
...
-	// generate_file has a invalid type.
+	// has a invalid type.
 	ErrInvalidContentType errors.Kind = "invalid content type"
 
+	// ErrInvalidConditionType indicates the condition attribute on
```
i4ki, line 34:
```suggestion
	// ErrInvalidConditionType indicates the condition attribute
```

generate/genfile/genfile.go
```diff
@@ -127,6 +141,23 @@ func Load(rootdir string, sm stack.Metadata, globals stack.Globals) (StackFiles,
...
+				return StackFiles{}, errors.E(ErrConditionEval, err)
+			}
+			if value.Type() != cty.Bool {
+				return StackFiles{}, errors.E(
+					ErrInvalidConditionType,
+					"condition has type %s but must be boolean",
+					value.Type().FriendlyName(),
+				)
+			}
+			condition = value.True()
```
i4ki, line 158: :rofl: :rofl: :rofl:

generate/genhcl/genhcl.go
```diff
@@ -51,16 +53,22 @@ const (
...
 
-	// ErrEval indicates the failure to evaluate the generate_hcl block.
-	ErrEval errors.Kind = "evaluating generate_hcl block"
+	// ErrContentEval indicates the failure to evaluate the content block.
+	ErrContentEval errors.Kind = "evaluating content block"
+
+	// ErrConditionEval indicates the failure to evaluate the condition attribute.
+	ErrConditionEval errors.Kind = "evaluating condition attribute"
+
+	// ErrInvalidConditionType indicates the condition attribute on
```
i4ki, line 69:
```suggestion
	// ErrInvalidConditionType indicates the condition attribute
```

generate/genhcl/genhcl.go
```diff
@@ -51,16 +53,22 @@ const (
...
+	// has a invalid type.
```
i4ki, line 70:
```suggestion
	// has an invalid type.
```

generate/genhcl/genhcl.go
```diff
@@ -74,24 +82,30 @@ func (s StackHCLs) GeneratedHCLs() map[string]HCL {
...
 // Origin returns the path, relative to the project root,
 // of the configuration that originated the code.
-func (b HCL) Origin() string {
-	return b.origin
+func (h HCL) Origin() string {
+	return h.origin
+}
+
+// Condition returns the result of the evaluation of the
+// condition attribute for the generated code.
```
i4ki, line 106:
```suggestion
// Condition returns the evaluated condition attribute for the generated code.
```

generate/genfile/genfile.go
```diff
@@ -127,6 +141,23 @@ func Load(rootdir string, sm stack.Metadata, globals stack.Globals) (StackFiles,
 			return StackFiles{}, errors.E(ErrContentEval, err)
 		}
 
+		condition := true
```
i4ki, line 144: we should not evaluate the content block before evaluating the `condition`.
If the condition is false, then the content must not be evaluated.
The condition can check for the existence of a variable that will be used in the content block, so with current approach some valid use cases would not be possible.

generate/genfile/genfile_test.go
```diff
@@ -145,14 +152,123 @@ func TestLoadGenerateFiles(t *testing.T) {
...
+					),
+				},
+			},
+			want: []result{
+				{
+					name:      "test",
+					condition: false,
+					file: genFile{
+						origin: "/stack/test.tm",
+						body:   "data",
```
i4ki, line 183: the body should not be evaluated in those cases IMHO

i4ki (dismissed):
LGTM but in my opinion we should not evaluate the content block if the condition is false.

In the `content` expression of `generate_file` or in the `content` block of the `generate_hcl` the user can have an expression that depend on the `condition` being true for being valid. Example:

```
condition = tm_try(global.bleh, false)
content = global.bleh
```

The `content` block also can contain reference for files (using `tm_file(), etc` that the condition should be used as guard if the file exist or not. So if the condition is true, it's safe to evaluate and generate the file.
I know we didnt talked about this in the spec but thinking now I guess it makes sense to change it.

i4ki (approved):
LGTM :rocket:

PR #375, opened 2022-06-02
refactor: isolate evaluation context better

katcipis:
Thinking on ways to isolate further the evaluation context so we can easily change the metadata definition in a single place and be confident that there will be no inconsistencies (like some part of the code manipulating the lower level eval ctx directly). The idea is to do it all through the stack.EvalCtx everywhere, since evaluation so far is always done regarding a stack anyway. The only one using more low level hcl eval contexts will be the stack.EvalCtx itself.

Also worked on isolating as much as possible decisions like "what is the name of the namespaces". Im not entirely sure this is a good idea, prone to just dropping it if it is stupid or if we have a better idea :-).

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -25,7 +25,6 @@ import (
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/mineiros-io/terramate/errors"
-	"github.com/mineiros-io/terramate/hcl/eval"
 	"github.com/rs/zerolog/log"
 	"github.com/zclconf/go-cty/cty"
 )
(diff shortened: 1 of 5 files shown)


hcl/hcl.go
```diff
@@ -109,6 +108,9 @@ type GenFileBlock struct {
 	Content *hclsyntax.Attribute
 }
 
+// PartialEvaluator represents an HCL partial evaluator
+type PartialEvaluator func(hclsyntax.Expression) (hclwrite.Tokens, error)
```
katcipis, line 112: Introduced this indirection to avoid dependency cycles. It would be odd for something general purpose like hcl to depend on a stack.EvalCtx anyway + I really wanted to avoid allowing the inner evaluation context on stack.EvalCtx to leak publicly, so it seemed like the only way.

stack/eval.go
```diff
@@ -15,37 +15,69 @@
...
-	if err := evalctx.SetNamespace("global", globals.Attributes()); err != nil {
-		return nil, errors.E(sm, err, "setting global namespace on eval context")
+	if err := evalctx.SetGlobals(globals); err != nil {
+		return nil, err
 	}
 
 	return evalctx, nil
 }
 
+// SetGlobals sets the given globals on the stack evaluation context.
```
katcipis, line 56: Not a huge fan of these really small layers, but wanted to avoid leaking the inner evalctx that allows overall manipulation of namespaces/etc.

i4ki (approved):
:heart:

PR #377, opened 2022-06-03
chore: remove duplicated import

i4ki:


hcl/eval/eval.go
```diff
@@ -98,7 +97,7 @@ func (c *Context) PartialEval(expr hclsyntax.Expression) (hclwrite.Tokens, error
 
 	exprRange := expr.Range()
 	exprBytes := filedata[exprRange.Start.Byte:exprRange.End.Byte]
-	tokens, diags := hclsyntax.LexExpression(exprBytes, exprFname, hcl.Pos{})
+	tokens, diags := hclsyntax.LexExpression(exprBytes, exprFname, hhcl.Pos{})
```
katcipis, line 100: hmm can't we use just "hcl" since it seems the name is available ? :thinking: (I thought hhcl was used because we had a clash)

i4ki: we have an `hcl` package... this one is just to make it clear it's hashicorp's hcl..

katcipis: when we import our own "errors" package we don't call it something else just because there is also a Go std "errors" package...unless there is a clash :thinking:

katcipis (approved):

hcl/eval/eval.go
i4ki: for me it makes sense for having naming consistency, like `hcl` is own hcl and `hhcl` is hashicorp's one throughout the code base.. but we didn't talk about this yet. In some places we do need to rename anyway (like the stack/globals.go and hcl_test.go).

I don't like to depend on the editor to figure this out because sometimes we need to know it in a review snippet.

But no strong feelings, we can change the naming later, I'm just fixing the duplicated import here.

PR #379, opened 2022-06-03
chore: reproduce issue

i4ki:


generate/genhcl/genhcl_test.go
```diff
@@ -2449,6 +2450,52 @@ func TestPartialEval(t *testing.T) {
 				str("string", `/stack test`),
 			),
 		},
+		{
+			name:      "reproduce issue #376 - byte slice corruption",
+			stackpath: "live/production/us-east-1/infrastructure/route53-association",
```
katcipis, line 2455: hmm for things that depends on where stacks are located (or amount of tests) we could use the generate_hcl tests at the generate package (requiring stackpath here seems like a considerable outlier).

hcl/eval/partial.go
```diff
@@ -1340,6 +1347,24 @@ func ignorenlc(tokens hclwrite.Tokens) hclwrite.Tokens {
...
 
+func copytokens(tokens hclwrite.Tokens) hclwrite.Tokens {
+	var newtokens hclwrite.Tokens
+	for _, tok := range tokens {
+		newtokens = append(newtokens, copytoken(tok))
+	}
+	return newtokens
+}
+
+func copytoken(tok *hclwrite.Token) *hclwrite.Token {
```
katcipis, line 1358: Copy for the win :rocket:

katcipis (dismissed):
LGTM + some small considerations if we should test stack path related things on the genhcl package, but not sure about it, feel free to disregard.

generate/genhcl/genhcl_test.go
i4ki: I placed here because in the `generate_hcl_test.go` the files are actually generated but this is not needed here. 

As the problem is not really related to stack but interpolated string having a decent length, maybe I will remove this and focus on having big globals only.
wdyt?

katcipis: If we can reproduce with big globals too (makes sense given the problem) I think it would make things simpler :100:

katcipis (approved):

PR #380, opened 2022-06-03
feat: move stack metadata related to terramate.stack

katcipis:
Moving stack metadata to "terramate.stack". The old metadata is being kept for now to avoid breaking anything. New metadata:

* terramate.name -> terramate.stack.name
* terramate.description -> terramate.stack.description
* terramate.path -> terramate.stack.path.absolute

New metadata will be added on a future PR.

### DIFF
--- a/hcl/eval/eval.go
+++ b/hcl/eval/eval.go
@@ -60,9 +60,9 @@ func (c *Context) SetNamespace(name string, vals map[string]cty.Value) error {
 		Str("action", "SetNamespace()").
 		Logger()
 
-	logger.Trace().
-		Msg("Convert from map to object.")
-	obj, err := fromMapToObject(vals)
+	logger.Trace().Msg("Convert from map to object")
+
+	obj, err := FromMapToObject(vals)
 	if err != nil {
 		return fmt.Errorf("setting namespace %q:%v", name, err)
 	}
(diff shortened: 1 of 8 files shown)


hcl/eval/eval.go
```diff
@@ -107,39 +107,40 @@ func (c *Context) PartialEval(expr hclsyntax.Expression) (hclwrite.Tokens, error
...
 		ctyTypes[key] = value.Type()
 	}
 
-	logger.Trace().
-		Msg("Convert type and value to object.")
+	logger.Trace().Msg("Convert type and value to object")
+
 	ctyObject := cty.Object(ctyTypes)
-	ctyVal, err := gocty.ToCtyValue(m, ctyObject)
+	ctyVal, err := gocty.ToCtyValue(values, ctyObject)
```
katcipis, line 126: I'm torn if we shouldn't just panic here. The conversion always uses terramate evaluated things (or built by ourselves some other way), so feels like a bug/critical thing going on that could be panicked (would simplify a lot the error handling around this). WDYT @i4ki ?

i4ki: I think makes sense, as this a controlled namespace.

katcipis: Will improve this on a follow up PR.

PR #387, opened 2022-06-07
refactor: avoid unnecessary err on namespace building

katcipis:
As discussed previously [here](https://github.com/mineiros-io/terramate/pull/380#discussion_r890984111)


i4ki (approved):
top!

PR #392, opened 2022-06-09
chore: add initial spec for terramate.config.run.env

katcipis:
Initial spec for the `export_env` feature. We could use different names, maybe `run_env` or `exec_env` would be more explicit ? Go crazy on name suggestions =P

### DIFF
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -282,7 +282,7 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
 The environment in which stacks are going to be executed can be configured
 using the `terramate.config.run.env` block. This block has no labels and
 accepts arbitrary attributes where each attribute represents an environment
-variable that will be exported when executing the stack.
+variable that will be evaluated and exported when executing the stack.
 The attributes must always evaluate to strings.
 
 Example:
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -280,40 +280,52 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
 ## Execution Environment
 
 The environment in which stacks are going to be executed can be configured
-using the `export_env` block. This block has no labels and accepts arbitrary
-attributes where each attribute represents an environment variable that will
-be exported when executing the stack. The attributes must always evaluate to
-strings.
+using the `terramate.config.run.env` block. This block has no labels and
+accepts arbitrary attributes where each attribute represents an environment
+variable that will be exported when executing the stack.
+The attributes must always evaluate to strings.
 
 Example:
 
 ```hcl
-export_env {
-  TF_PLUGIN_CACHE_DIR = "/some/path/etc"
+terramate {
+  config {
+    run {
+      env {
+        TF_PLUGIN_CACHE_DIR = "/some/path/etc"
+      }
+    }
+  }
 }
 ```
 
 Will export the environment variable `TF_PLUGIN_CACHE_DIR` on the stack
 execution environment with the value `/some/path/etc`.
 
-Inside the `export_env` block the `env` namespace will be available including
-environment variables exported from the host:
+Inside the `terramate.config.run.env` block the `env` namespace will be
+available including environment variables exported from the host:
 
 ```hcl
-export_env {
-  TF_PLUGIN_CACHE_DIR = "${env.HOME}/.terraform-cache-dir"
+terramate {
+  config {
+    run {
+      env {
+        TF_PLUGIN_CACHE_DIR = "${env.HOME}/.terraform-cache-dir"
+      }
+    }
+  }
 }
 ```
 
 The `env` namespace is read-only and is only available when evaluating
-`export_env` blocks.
+`terramate.config.run.env` blocks.
 
-The `export_env` blocks have the same hierarchical behavior as other features
-on Terramate, so an `export_env` defined on the project root is inherited by
-all stacks in the project.
+The `terramate.config.run.env` blocks have the same hierarchical behavior
+as other features on Terramate, so an `terramate.config.run.env` defined on
+the project root is inherited by all stacks in the project.
 
-All `export_env` blocks are merged, so different variables defined on different
-blocks will all be exported on the stack execution environment.
+All `terramate.config.run.env` blocks are merged, so different variables
+defined on different blocks will all be exported on the stack execution environment.
 
 Regarding variables redefinitions:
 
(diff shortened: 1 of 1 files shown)


docs/orchestration.md
```diff
@@ -278,6 +277,46 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
...
+
+The `export_env` blocks have the same hierarchical behavior as other features
+on Terramate, so an `export_env` defined on the project root is inherited by
+all stacks in the project.
+
+Regarding redefinitions:
+
+* Redefinition on same level is disallowed
+* Redefinition on different levels: more specific definition (closer to stack) is used
+
```
mariux, line 319: for each variable! not for the block... we can have multi blocks but different vars on same level.. ?

katcipis: I was thinking for each variable, but was not explicit on the text, let me improve the spec

katcipis:
We could handle this as an overall Terramate configuration for the run environment and then configure it inside the `terramate` block, like this:

```hcl
terramate {
  config {
    run {
      env {
          ENVIRONMENT_VAR="etc"
      }
    }
  }
}
```

Analogous to what we already have for git. WDYT @mariux @i4ki @soerenmartius @waxb ?

docs/orchestration.md
```diff
@@ -278,6 +277,61 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
...
+    run {
+      env {
+        TF_PLUGIN_CACHE_DIR = "${env.HOME}/.terraform-cache-dir"
+      }
+    }
+  }
+}
+```
+
+The `env` namespace is read-only and is only available when evaluating
```
i4ki, line 320: it's strange to say that the `env` namespace is read-only if this config is for updating it. Do you mean, the `env` namespace is read-only inside the `.run.env` config? if so, there's no way to update a variable from an expression (AFAIK).

or maybe you mean that you cannot create variables in the `env` and reference the newly created variable inside this block? Like the config below, is it supported by current spec?

```
terramate {
  config {
    run {
      env {
        A = "test"
        B = "other ${A}"
      }
    }
  }
}
```

docs/orchestration.md
```diff
@@ -278,6 +277,61 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
 **runtime** order.
 
 
+## Execution Environment
+
+The environment in which stacks are going to be executed can be configured
+using the `terramate.config.run.env` block. This block has no labels and
+accepts arbitrary attributes where each attribute represents an environment
+variable that will be exported when executing the stack.
```
i4ki, line 285:
```suggestion
variable that will be evaluated and exported when executing the stack.
```

so users know they can also create environment variables referencing `terramate.stack.*` variables.

i4ki: one example would be:

```
terramate {
  config {
    run {
      env {
        PATH = "${terramate.stack.path.absolute}/bin}:${env.PATH}"
      }
    }
  }
}
```

i4ki (approved):
lgtm but left some question and improvement ideas.

The name is good IMO.

docs/orchestration.md
katcipis: I meant that the attributes defined inside the env block do not affect the `env` namespace (they only affect the actual environment when running commands), the namespace only exports the current environment from the host (hence read only). I'll try to make this more clear, I had a more elaborate explanation before but I remove it because I thought it complicated things...apparently I was wrong :laughing:

katcipis: Tried to improve the docs around this + added another example

docs/orchestration.md
```diff
@@ -278,6 +277,78 @@ Even though **stack-c** defined that it needs to be run after **stack-b**, since
...
+```
+
+Globals and Metadata can also be referenced:
+
+```hcl
+terramate {
+  config {
+    run {
+      env {
+        TF_PLUGIN_CACHE_DIR = "${terramate.stack.path.absolute}/${global.cache_dir}"
```
waxb, line 327: idea: what if `terramate` caches the stuff by default to a path maybe specific for the terramate setup and if this environment variable set then it overrides it, and to disable it there could be something like `TF_PLUGIN_CACHE=false`?

waxb:
> We could handle this as an overall Terramate configuration for the run environment and then configure it inside the `terramate` block, like this:
> 
> ```terraform
> terramate {
>   config {
>     run {
>       env {
>           ENVIRONMENT_VAR="etc"
>       }
>     }
>   }
> }
> ```
> 
> Analogous to what we already have for git. WDYT @mariux @i4ki @soerenmartius @waxb ?

maybe the question is coming from my limited knowledge but is the `run` block used for anything else besides `env {}`? Or is it planned to use it for something in the future? For me in this context `config` and `run` seems a bit of interchangeable, of course i can imagine use cases where they are differ

katcipis:
> > We could handle this as an overall Terramate configuration for the run environment and then configure it inside the `terramate` block, like this:
> > ```terraform
> > terramate {
> >   config {
> >     run {
> >       env {
> >           ENVIRONMENT_VAR="etc"
> >       }
> >     }
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Analogous to what we already have for git. WDYT @mariux @i4ki @soerenmartius @waxb ?
> 
> maybe the question is coming from my limited knowledge but is the `run` block used for anything else besides `env {}`? Or is it planned to use it for something in the future? For me in this context `config` and `run` seems a bit of interchangeable, of course i can imagine use cases where they are differ

Good question :100: 

We already have different use cases for config, for example git related config, so it would look like this:

```hcl
terramate {
  config {
    git {
      // Git related configs
    }
    run {
      env {...}
    }
  }
}
```

So the idea is to have something run specific inside the overall config, since the config block is a general purpose configuration block. The `run` block itself would be used only for the env, for now, but we could have some run related options in the future, lets say we have support for parallel execution. On top of being configured with a flag we could also do it on the config:

```hcl
terramate {
  config {
    git {
      // Git related configs
    }
    run {
      parallel = true // Or something like this, etc
      env {...}
    }
  }
}
```

So it seems to be to be worthwhile having a dedicated `run` block inside the general purpose `config` one.

docs/orchestration.md
i4ki: I think it makes sense to have a default if this is what most people need but I don't have much experience to judge the benefit of it..

i4ki (approved):

PR #393, opened 2022-06-10
fix: conflict detection only for generated files

### LINKED ISSUE #390: [BUG] Label conflict in conditional code generation when different conditions is used
**Describe the bug**
When we conditionally generate code we should allow same names when only one condition is valid

**To Reproduce**

Try to run terraform generate with next code snippet

```

generate_hcl "providers.gen.tf" {
  condition = global.aws_role_arn != ""

  content {
    provider "aws" {
      region = "${global.region}"
      assume_role {
        role_arn = global.aws_role_arn
      }
    }
  }
}

generate_hcl "providers.gen.tf" {
  condition = global.aws_role_arn == ""

  content {
    provider "aws" {
      region = "${global.region}"
    }
  }
}
```

**Expected behavior**
Code should be generated without errors because only one condition will be valid at the same time

**Log Output**
Add logs from Terramate to help debug your problem.

```
- stack /some-private-stack
        error: label conflict detected: loading generate_hcl: /global.tm.hcl has blocks with same label "providers.gen.tf"

Hint: '+', '~' and '-' means the file was created, changed and deleted, respectively.
```

 - OS: [e.g. Linux] MacOS (MBP M1)
 - OS Version [e.g. Ubuntu 20.04]
 - Git Version [e.g. 2.35.1] (git --version) git version 2.32.0 (Apple Git-132)
 - Terramate Version [e.g. 0.1.0] (terramate --version) 0.1.4

**Additional context**

i4ki:
Closes #390

### DIFF
--- a/generate/generate_test.go
+++ b/generate/generate_test.go
@@ -63,6 +63,7 @@ func testCodeGeneration(t *testing.T, checkGenFile genFileChecker, tcases []test
 
 	for _, tcase := range tcases {
 		t.Run(tcase.name, func(t *testing.T) {
+			t.Helper()
 			s := sandbox.New(t)
 			s.BuildTree(tcase.layout)
 
--- a/generate/genfile/genfile.go
+++ b/generate/genfile/genfile.go
@@ -15,7 +15,9 @@
 package genfile
 
 import (
+	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/mineiros-io/terramate/errors"
@@ -46,17 +48,17 @@ const (
 	ErrLabelConflict errors.Kind = "label conflict detected"
 )
 
-// StackFiles represents all generated files for a stack,
-// mapping the generated file path to the actual file body.
-type StackFiles struct {
-	files map[string]File
-}
-
 // File represents generated file from a single generate_file block.
 type File struct {
+	name      string
 	origin    string
 	body      string
 	condition bool
+}
+
+// Name of the file.
+func (f File) Name() string {
+	return f.name
 }
 
 // Body returns the file body.
@@ -82,14 +84,9 @@ func (f File) Header() string {
 	return ""
 }
 
-// GeneratedFiles returns all generated files, mapping the file path to
-// the file description. The path is absolute relative to the project root.
-func (s StackFiles) GeneratedFiles() map[string]File {
-	cp := map[string]File{}
-	for k, v := range s.files {
-		cp[k] = v
-	}
-	return cp
+func (f File) String() string {
+	return fmt.Sprintf("Generating file %q (condition %t) (body %q) (origin %q)",
+		f.Name(), f.Condition(), f.Body(), f.Origin())
 }
 
 // Load loads and parses from the file system all generate_file blocks for
@@ -104,7 +101,7 @@ func (s StackFiles) GeneratedFiles() map[string]File {
 // generate_file blocks.
 //
 // The rootdir MUST be an absolute path.
-func Load(rootdir string, sm stack.Metadata, globals stack.Globals) (StackFiles, error) {
+func Load(rootdir string, sm stack.Metadata, globals stack.Globals) ([]File, error) {
 	stackpath := filepath.Join(rootdir, sm.Path())
 	logger := log.With().
 		Str("action", "genfile.Load()").
@@ -115,23 +112,25 @@ func Load(rootdir string, sm stack.Metadata, globals stack.Globals) (StackFiles,
 
 	genFileBlocks, err := loadGenFileBlocks(rootdir, stackpath)
 	if err != nil {
-		return StackFiles{}, errors.E("loading generate_file", err)
+		return nil, errors.E("loading generate_file", err)
 	}
 
 	evalctx, err := stack.NewEvalCtx(stackpath, sm, globals)
 	if err != nil {
-		return StackFiles{}, errors.E(err, "creating eval context")
+		return nil, errors.E(err, "creating eval context")
 	}
 
 	logger.Trace().Msg("generating files")
 
-	res := StackFiles{
-		files: map[string]File{},
-	}
+	var files []File
 
-	for name, genFileBlock := range genFileBlocks {
+	for _, genFileBlock := range genFileBlocks {
+		name := genFileBlock.label
+		origin := genFileBlock.origin
+
 		logger := logger.With().
 			Str("block", name).
+			Str("origin", origin).
 			Logger()
 
 		logger.Trace().Msg("evaluating condition")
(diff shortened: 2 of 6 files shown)


generate/generate_test.go
```diff
@@ -58,11 +59,96 @@ func (s stringer) String() string {
...
+			wantReport: generate.Report{
+				Successes: []generate.Result{
+					{
+						StackPath: "/stacks/stack",
+						Created:   []string{"repeated"},
+					},
+				},
+			},
+		},
+	})
```
katcipis, line 143: Not fundamental, but maybe add a test where we have generate_hcl + generate_file on the same level (like the stack) but it works because of the condition (checking it works when it is unrelated to hierarchy).

generate/genfile/genfile.go
```diff
@@ -243,38 +247,8 @@ func loadGenFileBlocks(rootdir string, cfgdir string) (map[string]genFileBlock,
...
 
-	if err := merge(res, parentRes); err != nil {
-		return nil, err
-	}
+	res = append(res, parentRes...)
 
 	logger.Trace().Msg("loaded generate_file blocks with success.")
 	return res, nil
 }
-
```
katcipis, line 253: noice :100:

generate/genfile/genfile_test.go
```diff
@@ -536,7 +536,63 @@ stack_description=
...
+						labels("test"),
+						str("content", "test"),
+						boolean("condition", false),
+					),
+				},
+			},
+			want: []result{
+				{
+					name:      "test",
+					condition: false,
```
katcipis, line 582: hmmm a  change on the string representation of the file could change the ordering according to condition and break the test I think :thinking:.

I'm kinda torn on the sorting using String() + also torn if we should have ordering. We already re-order things on generate, on genfile/genhcl I dont think we have a strong constraint on ordering other than making the tests easier, but maybe we could write the tests in a way that is order independent. Or at least being more explicit on the ordering instead of using the string representation :thinking:

generate/genhcl/genhcl.go
```diff
@@ -107,6 +94,11 @@ func (h HCL) Condition() bool {
 	return h.condition
 }
 
+func (h HCL) String() string {
+	return fmt.Sprintf("Generating file %q (condition %t) (body %q) (origin %q)",
```
katcipis, line 98:
```suggestion
	return fmt.Sprintf("generate_hcl %q (condition %t) (body %q) (origin %q)",
```

We are not generating the file here, just a representation of a loaded generate_hcl block.

generate/genhcl/genhcl.go
```diff
@@ -254,43 +251,9 @@ func loadGenHCLBlocks(rootdir string, cfgdir string) (map[string]loadedHCL, erro
...
-		return nil, errors.E(ErrLabelConflict, err)
-	}
+
+	res = append(res, parentRes...)
 
 	logger.Trace().Msg("loaded generate_hcl blocks with success.")
 	return res, nil
 }
-
-func join(target, src map[string]loadedHCL) error {
```
katcipis, line 265: Noice :100: . Also funny why I called join here but merge on generate file, Im stupid :laughing:

generate/genhcl/genhcl_test.go
```diff
@@ -1007,7 +1007,7 @@ func TestLoadGeneratedHCL(t *testing.T) {
 			},
 		},
 		{
-			name:  "stack with block with same label as parent is an error",
+			name:  "stack with block with same label as parent is not an error",
```
katcipis, line 1010:
```suggestion
			name:  "stack with block with same label as parent is allowed",
```

generate/genhcl/genhcl_test.go
```diff
@@ -1207,10 +1249,31 @@ func TestLoadGeneratedHCL(t *testing.T) {
 					),
 				},
 			},
-			wantErr: errors.E(genhcl.ErrLabelConflict),
+			want: []result{
```
katcipis, line 1252: The test name was not updated [here](https://github.com/mineiros-io/terramate/pull/393/files#diff-74888afee21574583ccb4e0af6e3ab123ccf5cc5dfb2715c9cd156539a987ce6R1227), it still says that blocks with same label fails.

generate/genhcl/genhcl_test.go
```diff
@@ -1207,10 +1249,31 @@ func TestLoadGeneratedHCL(t *testing.T) {
...
+						body: terraform(
+							str("data2", "some literal data2"),
+						),
+					},
+				},
+			},
 		},
 		{
-			name:  "blocks with same label on multiple config files fails",
+			name:  "blocks with same label on multiple config files do not fail",
```
katcipis, line 1276:
```suggestion
			name:  "blocks with same label on multiple config files are allowed",
```

generate/genhcl/genhcl_test.go
```diff
@@ -1410,16 +1501,18 @@ func TestLoadGeneratedHCL(t *testing.T) {
 				wantcode := res.hcl.body.String()
 
 				assertHCLEquals(t, gotcode, wantcode)
+				assert.EqualStrings(t,
+					res.name,
+					gothcl.Name(),
+					"wrong origin config path for generated code",
```
katcipis, line 1507:
```suggestion
					"wrong name for generated code",
```

katcipis (dismissed):
LGTM + some small possible improvements. Regarding sorting it can also be ignored or assessed later in a different PR. Thanks for fixing this :rocket:

generate/generate_hcl_test.go
```diff
@@ -308,6 +308,189 @@ func TestGenerateHCL(t *testing.T) {
...
+				},
+				{
+					path: "/stacks/stack",
+					add: generateHCL(
+						labels("repeated"),
+						content(
+							block("block",
+								str("data", "stack data"),
+							),
+						),
```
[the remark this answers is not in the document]
i4ki: the intention was to make sure it also follow the same rules if condition is missing (using the default value).

generate/generate_test.go
i4ki: test added. Thanks

generate/genhcl/genhcl_test.go
i4ki: fixed, thanks

generate/genfile/genfile_test.go
i4ki: you're right, there's no constraint other than making the tests easier (at this point) but also I don't have anything against making the order of generated files deterministically. We definitely need to talk about this, I did the simpler approach...

generate/generate_hcl_test.go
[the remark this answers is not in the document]
katcipis: Ah OK, I think some other tests didn't had the missing condition... or I just missed them and gave a false positive on inconsistency,sorry for the noise :+1:

generate/genfile/genfile_test.go
katcipis: Sure thing, I'm not a 100% sure on either approach (sorting or not, and if we sort which criteria), lets discuss in the future :rocket:

katcipis: Ah...actually I forgot we have some other maps involved in the parsing, so we would lose deterministic results. Since we replaced maps with slices my first instinct was that now we have deterministic output, just unordered, but that is not true here, so I would be pro the ordering because deterministic output is cool =P. Still wondering if the sorting implementation is the best, but that can be assesed/changed later.

PR #397, opened 2022-06-14
chore: improve type checks in the stack attributes.

i4ki:


hcl/hcl.go
```diff
@@ -639,28 +639,43 @@ func assignSet(name string, target *[]string, val cty.Value) error {
...
 	for iterator.Next() {
+		index++
 		_, elem := iterator.Element()
 
 		logger.Trace().Msg("Check element is of correct type.")
 
 		if elem.Type() != cty.String {
-			errs.Append(errors.E("field %q is a set(string) but contains %q",
-				name, elem.Type().FriendlyName()))
+			errs.Append(errors.E("field %q must be a set(string) but element %d "+
```
katcipis, line 666:
```suggestion
			errs.Append(errors.E("field %q must be a string but element %d "+
```

katcipis (commented):
The error message references set(string) but I think it is supposed to be just string

hcl/hcl.go
i4ki: this generates the error: `field "wants" must be a set(string) but element 1 has type "number"`.
I'm using the terraform type declaration here.. not sure if ok, but if then it says the `wants` must be a `set` (unique values) of `string` values.

i4ki:
@soerenmartius can I have a quick review here?

katcipis (approved):

PR #403, opened 2022-06-16
feat: add terramate.config.run.env

katcipis:
This introduces the basics for the terramate.config.run.env block parsing. But still doens't expose the feature + there is no hierarchical behavior/merging. This will be addressed on follow up PRs.


hcl/attribute.go
```diff
@@ -0,0 +1,58 @@
...
+import "github.com/hashicorp/hcl/v2/hclsyntax"
+
+// Attribute represents a parsed attribute.
+type Attribute struct {
+	origin string
+	val    *hclsyntax.Attribute
+}
+
+// Attributes represents multiple parsed attributes.
+// The attributes can be sorted by lexicographically by their name.
```
katcipis, line 26:
```suggestion
// The attributes can be sorted lexicographically by their name.
```

hcl/attribute.go
```diff
@@ -0,0 +1,58 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hcl
+
+import "github.com/hashicorp/hcl/v2/hclsyntax"
+
+// Attribute represents a parsed attribute.
+type Attribute struct {
+	origin string
```
katcipis, line 21: Origin info will be good when we eval, so we can provide accurate data about where the original attribute came from.

hcl/attribute.go
```diff
@@ -0,0 +1,58 @@
...
+	val    *hclsyntax.Attribute
+}
+
+// Attributes represents multiple parsed attributes.
+// The attributes can be sorted lexicographically by their name.
+type Attributes []Attribute
```
katcipis, line 27: Since we augment the attribute to have the origin I already added an Attributes type that is sorteable (so we can always have deterministic output etc).

cmd/terramate/cli/project.go
```diff
@@ -39,7 +39,7 @@ type project struct {
 }
 
 func (p project) gitcfg() *hcl.GitConfig {
-	return p.rootcfg.Terramate.RootConfig.Git
+	return p.rootcfg.Terramate.Config.Git
```
katcipis, line 42: since these parsed structure doens't necessarily represents a root config, but any config block on any part of the hierarchy, I renamed it. I think we can also rename the types to reflect that, but will do that in a separate PR because the changeset is quite big.

i4ki: what about the issue we were talking earlier today? we talked about failing if the user has `terramate.config` in a non-root directory.. but the `run.env` feature kinda requires that.

katcipis: `terramate.config` is not a root configuration. The root configurations are:

* terramate.required_version
* terramate.config.git

At least as documented today. If everything inside `terramate.config` is supposed to be root only then we need to come up with a different place to put the run/env configurations.

katcipis: Maybe we should have a project_config block to make it explicit that it has project wide configuations that can only be defined on the top level/root dir ? :thinking:

hcl/hcl.go
```diff
@@ -1154,13 +1202,40 @@ func (p *TerramateParser) parseTerramateSchema() (Config, error) {
...
+		config.Terramate.Config.Run.Env == nil {
+		return nil
+	}
+
+	errs := errors.L()
+	attrs := map[string]Attribute{}
+
+	for _, attr := range config.Terramate.Config.Run.Env.Attributes {
+		name := attr.Value().Name
+		if _, ok := attrs[name]; ok {
```
i4ki, line 1227: isn't this ensured by the hcl language spec and also validated by hcl lib?

got this in the playground:
```
playground:13,4-5: Attribute redefined; The argument "a" was already set at playground:12,4-5. Each argument may be set only once.
```

hcl/hcl.go
```diff
@@ -1302,3 +1377,21 @@ func listTerramateDirs(dir string) ([]string, error) {
...
 }
+
+func sortAttributes(attrs hclsyntax.Attributes) []*hclsyntax.Attribute {
+	names := make([]string, 0, len(attrs))
+
+	for name := range attrs {
+		names = append(names, name)
+	}
+
+	log.Trace().Str("action", "sortedAttributes()").Msg("Sort attributes.")
```
i4ki, line 1388:
```suggestion
	log.Trace().Str("action", "sortAttributes()").Msg("Sort attributes.")
```

i4ki (approved):
lgtm but left some comments to discuss.
Also, I know there's some discussion around the spec going on in other channels.

hcl/hcl.go
katcipis: Not ensured if they come from different blocks or different files. Like this is allowed:

```
block {
  a = 666
}

block {
  a = 666
}
```

Here we detect conflicts accross multiple blocks/files, so I think we need this (you can check the tests for these errors too).

hcl/hcl_cfg_run_test.go
```diff
@@ -0,0 +1,505 @@
...
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/madlambda/spells/assert"
+	"github.com/mineiros-io/terramate/errors"
+	"github.com/mineiros-io/terramate/hcl"
+)
+
+func TestHCLParserConfigRun(t *testing.T) {
+
+	runEnvCfg := func(hcldoc string) hcl.Config {
+		// Comparing attributes/expressions with hcl/hclsyntax is impossible
```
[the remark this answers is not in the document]
katcipis: I would love something better for this...let me know if you have any ideas :-(

hcl/hcl.go
i4ki: gotcha

cmd/terramate/cli/project.go
katcipis: we discussed the following options with the team:

* Only allow run/env on root and update specs (but if we want hierarchical in the future the problem comes back)
* Handle as project root terramate.required_version OR terramate.config.git instead of overall existence of terramate.config
* Move terramate.config.run.env to some other place not inside terramate.config and keep it as is
* Move `terramate.config` to something like terramate.project to make it more explicit that it is a project wide/root only config
* Make terramate.config a non-root config that is hierarchical + change project root detection

We ended up deciding for "Only allow run/env on root and update specs" for now. So `terramate.config.run.env` will be a root only configuration for now, no hierarchical/merge behavior.

i4ki: awesome.
:rocket:

i4ki (approved):

PR #407, opened 2022-06-17
chore: redefine spec for run.env

katcipis:


i4ki (approved):
lgtm but maybe makes sense to keep some words in the `orchestration.md` mentioning that `terramate.config.run.env` can be used to change the env and provide a link?

PR #411, opened 2022-06-20
chore: add docs for 'import' block

i4ki:
I'm not sure about the naming yet, feel free to suggest changes.

In my mind, `import` works better if we support namespace in the future. Eg.:

```
import {
    source = "/other/common"
    as = "common"
}

globals {
    test = "my ${common.global.test}"
}
```


soerenmartius (approved):
lgtm

docs/config-overview.md
```diff
@@ -15,13 +13,41 @@ Terramate files can be found in any non-hidden directory of a Terramate project
 and all non-hidden files in a single directory will be handled as the 
 concatenation of all of them in a single file, forming a single **configuration**.
 
-A Terramate project is essentially a collection of Terraform code
-organized into stacks.
+The configuration blocks can be defined multiple times and its values are merged
```
soerenmartius, line 16:
```suggestion
The configuration blocks can be defined multiple times and their values are merged
```

docs/config-overview.md
```diff
@@ -15,13 +13,41 @@ Terramate files can be found in any non-hidden directory of a Terramate project
...
+```
+# globals.tm
+
+import {
+    source = "/more/globals.tm"
+}
+```
+
+The `source` must reference a file using a relative path or an absolute path
+relative to project's root.
```
soerenmartius, line 31:
```suggestion
relative to the project's root.
```

docs/config-overview.md
```diff
@@ -30,6 +56,45 @@ In general, a Terramate project looks like this:
...
 * Configuration may be present on any directory.
 
+# Config merging
+
+The configuration defined in a directory is merged into a single configuration
+where multiple blocks of same type can be defined if their contents do not
+conflict. In other words, the definition of a block can be split into multiple
+blocks where each define a part of the whole definition. The only exception being
+the [generate](https://github.com/mineiros-io/terramate/blob/main/docs/codegen/overview.md) blocks and the `import` block. 
+The [globals](https://github.com/mineiros-io/terramate/blob/main/docs/sharing-data.md) block extend the merging to the hierarchy of globals.
```
soerenmartius, line 66:
```suggestion
The [globals](https://github.com/mineiros-io/terramate/blob/main/docs/sharing-data.md) block extends the merging to the hierarchy of globals.
```

docs/config-overview.md
```diff
@@ -30,6 +56,45 @@ In general, a Terramate project looks like this:
 * Stacks are organized as different directories.
 * Configuration may be present on any directory.
 
+# Config merging
+
+The configuration defined in a directory is merged into a single configuration
+where multiple blocks of same type can be defined if their contents do not
+conflict. In other words, the definition of a block can be split into multiple
+blocks where each define a part of the whole definition. The only exception being
```
soerenmartius, line 64:
```suggestion
blocks where each defines a part of the whole definition. The only exception being
```

soerenmartius, line 64:
```suggestion
blocks where each define a part of the whole definition. The only exception is
```

docs/config-overview.md
```diff
@@ -30,6 +56,45 @@ In general, a Terramate project looks like this:
...
+terramate {
+    config {
+        git {
+            default_branch = "main"
+        }
+    }
+}
+```
+
+and the blocks can also be defined in different files.
```
soerenmartius, line 84:
```suggestion
And the blocks can also be defined in different files.
```

soerenmartius (changes requested):

docs/config-overview.md
```diff
@@ -70,7 +137,7 @@ The `terramate.config.git` block has no labels and has the following schema:
 
 # stack block schema
 
-The `stack` block has no labels and has the following schema:
+The `stack` block has no labels, supports merging and has the following schema:
```
katcipis, line 140:
```suggestion
The `stack` block has no labels, supports [merging](#config-merging) and has the following schema:
```

katcipis (approved):
LGTM + small link addition for consistency :rocket:

katcipis (dismissed):

katcipis:
> didn't we align on include as we act as if the file content would be replaced as-is?

I understood from [here](https://github.com/mineiros-io/terramate/pull/411/files#diff-934147e7402d67ff2192bde8aaf423ff184b8225654846b4fc7c972020f5623dR33)

"The *imported* file is handled as if it's in the directory of the *importing* file"

That it is essentially the same as replacing the file content, maybe the phrasing is different but it should be the same behavior as acting like the imported file is now part of the configuration that imported it. :thinking:

soerenmartius (approved):
lgtm

i4ki:
@mariux I don't think we want a pure macro include (ala C/C++ preprocessor) implementation in the long term, so maybe we can avoid the compromise of adding this to the spec. 

The spec doesn't talk about the implementation but the outcome of importing files.
At this moment, implementing a preprocessor like include is more work than semantic importing, and both have the same high-level outcome but the include one needs more work to be user-friendly (good error messages and so on).
So I avoided talking about the implementation now and just do what users need, we can change the implementation later.

If you do think we should explicitly say (and implement) it's just "content replace" (like an `m4` preprocessor), then I'd like to enumerate some drawbacks of doing it (just for the sake of transparency).

If we solve the issues below, then it's not pure "file replace" anymore.

1. it's hard to provide meaningful error messages.

As the preprocessing happens before parsing, the Hashicorp library will report errors in the wrong line/column, which also breaks the editor extensions...
We can try some hacks to fixup the line/column but then it's a lot of more unneeded hacky code. (remember that we need to replace the import block being aware of the newlines and comments present, so we can calculate the right offsets).

2. We will need to add constraints to the preprocessing or change the file content explicitly

In order to have friendly error messages, we should not do a pure preprocessing because otherwise the following can happen:

```
# imported file
block {
```

```
# importing file

import {
    source = "/path/to/file"
}

attr = 1
```

The error message will be in the last line of the importing file and the only way of figuring out where is the problem is only by checking each imported file syntax. This problem will exist even if we fix up the line/column of HCL diagnostics...

So the imported file must be valid HCL, then it needs parsing before replacing the importing file content, which makes it a semantic importing anyway, with the drawback of having to parse the content twice.

3. We likely need to add more constraints or change the imported file content explicitly

Like enforcing newline at the end of files, etc, in order to make the final file a valid HCL.

There's not much benefit to implementing this as a preprocessor thing (IMO), when people can have this today already using a mature generic preprocessor (like m4) and some makefile/script for file generation.

So the spec here talks about a feature of `import` that just works, have good error messages and less surprises.

WDYT?

mariux (approved):
makes total sense ;) thanks for clarifying..

PR #413, opened 2022-06-20
chore: build terramate statically whenever possible

katcipis:
Not all targets targets support static binaries (like macos), but for linux now we have a static build, both on make + release + e2e testing.

To check the release binaries you can run:

```
make release/dry-run
```

And check inside the dist dir:

```
file dist/terramate_*/terramate
dist/terramate_darwin_amd64/terramate: Mach-O 64-bit x86_64 executable
dist/terramate_darwin_arm64/terramate: Mach-O 64-bit arm64 executable, flags:<|DYLDLINK|PIE>
dist/terramate_linux_386/terramate:    ELF 32-bit LSB executable, Intel 80386, version 1 (SYSV), statically linked, Go BuildID=Z0uQDQNinOYf7RfT23LA/6FMEyC8F2_39TyAJx8rp/Hzi3ovxvism8APu_csVO/V_GfsOkddGoQSay72BYt, not stripped
dist/terramate_linux_amd64/terramate:  ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=p_lac1sQB1eDcJloaogQ/o7V0Kf1ihJLTDY5NDJw7/CUhlwVzU0Q5pX9y00NhB/86hIfr1jW26fBhsYYmSL, not stripped
dist/terramate_linux_arm64/terramate:  ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=CguNnxFygOUHY-eAEDtO/PFFdtAYvH19ZYHSnhu16/ZcAT_rYOwzJgL4Ys44b3/2UtxG4gNwESxaoxYt3aR, not stripped
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,10 +37,6 @@ jobs:
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v2
-        with:
-          distribution: goreleaser
-          version: v1.2.5
-          args: release --rm-dist
+        run: make release
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SHELL := /bin/bash -o pipefail -o errexit -o nounset
 COVERAGE_REPORT ?= coverage.txt
 
 addlicense=go run github.com/google/addlicense@v1.0.0 -ignore **/*.yml
+buildflags=--ldflags '-extldflags "-static"'
 
 .PHONY: default
 default: help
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -37,10 +37,6 @@ jobs:
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v2
-        with:
-          distribution: goreleaser
-          version: v1.2.5
-          args: release --rm-dist
```
i4ki, line 44: :thinking:

Makefile
```diff
@@ -90,6 +92,16 @@ release/tag:
 	git tag -a $(VERSION) -m "Release $(VERSION)"
 	git push origin $(VERSION)
 
+## executes a dry run of the release process
+.PHONY: release/dry-run
+release/dry-run: 
+	go run github.com/goreleaser/goreleaser@$(GO_RELEASER_VERSION) release --snapshot --rm-dist
```
i4ki, line 98: in the PR description you said it was possible to check the dist directory but here you provide the `--rm-dist` option :thinking: 

```
--rm-dist                      Removes the dist folder
```

.github/workflows/release.yml
katcipis: I moved the go releaser call to make so we don't duplicate the go releaser version on the two CI files + we can call the dry run easily locally. WDYT ?

Makefile
katcipis: hmm true, that is odd.. I double checked and even with that option the dist file persists :thinking: 
when you run the dry run, do you also get the dist folder after ?

PR #414, opened 2022-06-20
fix: handle interruption signals when on run

### LINKED ISSUE #412: [BUG] Terramate do not wait to its backgroung processes when a stop signal is sent
**Describe the bug**

Terraform is still running after Terramate has stopped completely due to a stop signal (control-c).

**To Reproduce**

1. Given a very [simple project](https://github.com/alvsanand/playing-with-terramate) with several stacks. 
2. Run command 'terramate run terraform apply' and send a stop signal with control-c while Terraform is aplying changes.
3. See error

**Expected behavior**

Terramate exit and all its processes.

**Log Output**

Plan: 2 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

^C
Interrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...

Stopping operation...

~/workspace/playing-with-terramate$ ^C
~/workspace/playing-with-terramate$ ╷
│ Error: execution halted
│ 
│ 

 - OS: Ubuntu
 - OS Version: Ubuntu 20.04
 - Git Version: 2.25.1
 - Terramate Version: 0.1.7

**Additional context**

error_terrmate_exit_but_terraform_still_working

katcipis:
Currently Terramate will exit as soon as it gets an interruption signal, not waiting for the child process to exit. This PR introduces waiting for the child process to finish before exiting Terramate + It will force kill the child process after receiving 3 interruption signals.

Closes #412

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -936,8 +936,7 @@ func (c *cli) runOnStacks() {
 
 	err = run.Exec(
 		orderedStacks,
-		c.parsedArgs.Run.Command[0],
-		c.parsedArgs.Run.Command[1:],
+		c.parsedArgs.Run.Command,
 		c.stdin,
 		c.stdout,
 		c.stderr,
@@ -945,9 +944,17 @@ func (c *cli) runOnStacks() {
 	)
 
 	if err != nil {
-		logger.Warn().
-			Err(err).
-			Msg("failed to execute commands")
+
+		logger.Warn().Msg("one or more commands failed")
+
+		var errs *errors.List
+		if errors.As(err, &errs) {
+			for _, err := range errs.Errors() {
+				logger.Warn().Err(err)
+			}
+		} else {
+			logger.Warn().Err(err)
+		}
 
 		os.Exit(1)
 	}
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -950,10 +950,10 @@ func (c *cli) runOnStacks() {
 		var errs *errors.List
 		if errors.As(err, &errs) {
 			for _, err := range errs.Errors() {
-				logger.Warn().Err(err)
+				logger.Warn().Err(err).Send()
 			}
 		} else {
-			logger.Warn().Err(err)
+			logger.Warn().Err(err).Send()
 		}
 
 		os.Exit(1)
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -935,40 +934,28 @@ func (c *cli) runOnStacks() {
...
-		cmd.Env = os.Environ()
-		cmd.Stdin = c.stdin
-		cmd.Stdout = c.stdout
-		cmd.Stderr = c.stderr
-
-		logger := log.With().
-			Str("cmd", strings.Join(c.parsedArgs.Run.Command, " ")).
-			Stringer("stack", stack).
-			Logger()
+	err = run.Exec(
```
katcipis, line 937: Since run is more involved now with the signal handling I extracted it from the cli. It creates goroutines and changes signal handlers etc.

cmd/terramate/cli/cli.go
```diff
@@ -935,40 +934,28 @@ func (c *cli) runOnStacks() {
...
-					Err(err).
-					Msg("failed to execute command")
-			} else {
-				logger.Fatal().
-					Err(err).
-					Msg("failed to execute command")
+		var errs *errors.List
+		if errors.As(err, &errs) {
+			for _, err := range errs.Errors() {
+				logger.Warn().Err(err).Send()
```
katcipis, line 953: here there is a slight change in behavior, before when there was no continue on error the message would be fatal. Now the message is always a warning and then we exit with an error status code. Since exiting with the err status code is the most important part it seems this change is OK, given that it simplifies the handling of the errors (doesn't matter if it is one or N errors).

run/exec.go
```diff
@@ -0,0 +1,162 @@
...
+
+		logger := logger.With().
+			Stringer("stack", stack).
+			Logger()
+
+		logger.Info().Msg("Running")
+
+		// The child process should not get signals directly since
+		// we want to handle first interrupt differently.
+		cmd.SysProcAttr = &syscall.SysProcAttr{
```
katcipis, line 91: Without this I can't implement the behavior of first interruption not being sent to the child process. But after that the child process will not receive any other signal than interruptions sent from terramate, at first glance this seems OK but I could be missing something and some other signal may need to be relayed to the child process immediately :thinking:

i4ki: I think now the child process cannot read from the TTY anymore. 

Can you confirm that?
```
$ terramate run -- terraform console
```
here it hangs without reading/writing to the terminal.

katcipis: Hmm yeah there is no more TTY x_x. Thanks for pointing out.

run/exec.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		cmds <- cmd
+		interruptions := 0
+		cmdIsRunning := true
+
+		for cmdIsRunning {
+			select {
+			case sig := <-signals:
+				logger.Trace().
```
i4ki, line 97: What do you think of logging them as `info` here? Because this is a kinda useful feedback, so users know that terramate is aware of the SIGINT sent.

katcipis: hmm for the kill I added as Info but on this case it seemed that maybe we would generate too much noise...but I was on the fence. Will change to Info for now :+1:

i4ki (approved):
LGTM

PR #415, opened 2022-06-21
feat: customize interrupt signals

katcipis:


katcipis:
Customizing the signal handling behavior by detaching the child process from the parent group is more involved than I expected. Relaying signals is not enough since changing the group (to a non-foreground group) changes how access to terminal I/O happen (the child process loses access, which is bad for interactive processes).

Here is some more detailed explanation:

https://en.wikipedia.org/wiki/Process_group

The distribution of signals to process groups forms the basis of [job control](https://en.wikipedia.org/wiki/Job_control_(Unix)) employed by [shell programs](https://en.wikipedia.org/wiki/Shell_(computing)). The [TTY device](https://en.wikipedia.org/wiki/Computer_terminal) driver incorporates a notion of a foreground process group, to which it sends signals generated by [keyboard interrupts](https://en.wikipedia.org/wiki/Keyboard_controller_(computing)), notably [SIGINT](https://en.wikipedia.org/wiki/SIGINT_(POSIX)) ("interrupt", Control+C), [SIGTSTP](https://en.wikipedia.org/wiki/SIGTSTP) ("terminal stop", Control+Z), and [SIGQUIT](https://en.wikipedia.org/wiki/SIGQUIT) ("quit", Control+\). It also sends the [SIGTTIN](https://en.wikipedia.org/wiki/SIGTTIN) and [SIGTTOU](https://en.wikipedia.org/wiki/SIGTTOU) signals to any processes that attempt to read from or write to the terminal and that are not in the foreground process group. The shell, in turn, partitions the command [pipelines](https://en.wikipedia.org/wiki/Pipeline_(Unix)) that it creates into process groups, and controls what process group is the foreground process group of its [controlling terminal](https://en.wikipedia.org/wiki/Controlling_terminal), thus determining what processes (and thus what command pipelines) may perform I/O to and from the terminal at any given time.

PR #420, opened 2022-06-22
chore: add run.Env that loads a stack run environment

katcipis:
This will be used on the next PR to implement the run with env feature. When we are iterating the ordered stacks this function will be used to load the stack env so we can pass it to the created command.

### DIFF
--- a/run/env.go
+++ b/run/env.go
@@ -27,7 +27,7 @@ import (
 
 const (
 	// ErrParsingCfg indicates that an error happened while parsing configuration.
-	ErrParsingCfg errors.Kind = "parsing run env cfg"
+	ErrParsingCfg errors.Kind = "parsing terramate.config.run.env configuration"
 
 	// ErrLoadingGlobals indicates that an error happened while loading globals.
 	ErrLoadingGlobals errors.Kind = "loading globals to evaluate run env cfg"
--- a/run/env.go
+++ b/run/env.go
@@ -30,7 +30,7 @@ const (
 	ErrParsingCfg errors.Kind = "parsing terramate.config.run.env configuration"
 
 	// ErrLoadingGlobals indicates that an error happened while loading globals.
-	ErrLoadingGlobals errors.Kind = "loading globals to evaluate run env cfg"
+	ErrLoadingGlobals errors.Kind = "loading globals to evaluate terramate.config.run.env configuration"
 
 	// ErrEval indicates that an error happened while evaluating one of the
 	// terramate.config.run.env attributes.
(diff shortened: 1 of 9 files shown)


run/env.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// EnvVars represents a set of environment variables to be used
+// when running commands. Each string follows the same format used
+// on os.Environ and can be used to set env on exec.Cmd.
+type EnvVars []string
+
+// Env will load environment variables to be exported when running any command
+// inside the given stack. The order of the env vars is guaranteed to be the same
+// and is ordered lexicographically.
```
katcipis, line 51: This is not a hard requirement for setting the env on the command, but it is useful for testing + Im planning on have an experimental command to show the environment vars for each stack, that will make the output nice/deterministic there too.

run/env.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/mineiros-io/terramate/errors"
+	"github.com/mineiros-io/terramate/hcl"
+	"github.com/mineiros-io/terramate/stack"
+	"github.com/rs/zerolog/log"
+	"github.com/zclconf/go-cty/cty"
+)
+
+const (
+	// ErrParsingCfg indicates that an error happened while parsing configuration.
+	ErrParsingCfg errors.Kind = "parsing run env cfg"
```
i4ki, line 30:
```suggestion
	ErrParsingCfg errors.Kind = "parsing terramate.config.run.env configuration"
```
wdyt?

stack/eval.go
```diff
@@ -44,6 +46,17 @@ func (e *EvalCtx) SetMetadata(sm Metadata) {
 	e.evalctx.SetNamespace("terramate", metaToCtyMap(sm))
 }
 
+// SetEval sets the given environment on the env namespace of the evaluation context.
+// environ must be on the same format as os.Environ().
+func (e *EvalCtx) SetEval(environ []string) {
```
i4ki, line 51: SetEval? should not it be `SetEnviron()` or something?

i4ki (dismissed):
LGTM but left some suggestions and questions.

stack/eval.go
katcipis: The name of the namespace is "eval", so it seemed more straithforward :thinking: (maybe call SetEvalNamespace, but the other namespace setters are just SetNamespaceName).

katcipis: Oh my crap, it should be SetEnv :laughing: ....

i4ki (approved):

PR #422, opened 2022-06-23
feat: add env support on terramate run

katcipis:
Introduces support to manipulating the environment variables when running commands with `terramate run` as specified [here](https://github.com/mineiros-io/terramate/blob/main/docs/project-config.md#terramateconfigrunenv). 

It also introduces a `terramate experimental run-env` that should help debugging env definitions for stacks (only shows env manipulated directly, not like `env`).

Example:

```sh
#!/bin/bash

set -o errexit
set -o pipefail
set -o nounset

workdir=$(mktemp -d)
cd "${workdir}"

echo "workdir: ${workdir}"

cat >terramate.tm <<EOL
terramate {
    config {
      run {
        env {
          NEW_ENV_VAR="test"
          HOME="newhome"
        }
      }
    }
}
EOL

mkdir stack
terramate experimental init-stack stack
echo
terramate experimental run-env
echo
terramate run -- env
```

### DIFF
--- a/run/exec.go
+++ b/run/exec.go
@@ -36,6 +36,7 @@ import (
 // If continue on error is false it will return as soon as it finds an error,
 // returning a list with a single error inside.
 func Exec(
+	rootdir string,
 	stacks []stack.S,
 	cmd []string,
 	stdin io.Reader,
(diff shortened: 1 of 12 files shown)


run/exec.go
```diff
@@ -59,12 +77,10 @@ func Exec(
 
 	results := startCmdRunner(cmds)
 
-	errs := errors.L()
-
 	for _, stack := range stacks {
 		cmd := exec.Command(cmd[0], cmd[1:]...)
 		cmd.Dir = stack.HostPath()
-		cmd.Env = os.Environ()
+		cmd.Env = append(os.Environ(), stackEnvs[stack.Path()]...)
```
i4ki, line 83: I think we should handle the de-duplication ourselves (maybe in the future) and log the overridden entries, otherwise, it could be very confusing to understand some scenarios later when we introduce `config.run` per-directory (hierarchy, etc).
Even today, if the user us unaware of all envs set in the project and try to just export something in the shell and run `terramate run`, his value will be ignored and no log will show this.
But I think it's minor at this point.

katcipis: Agreed, needs improvement :100:

i4ki: I also dug into the Go's deduplication and it changed to resemble shells in Go1.9, before that it used the first entry... we already depend on go1.17 so it's ok. Just sharing this in case you maybe knows anything else that could be a problem in this case. Feel free to ignore.

i4ki: https://github.com/golang/go/issues/19877

katcipis: I was not aware of that, just tested and checked that the current behavior is to use the last one (tests also cover that, if behavior changes it would fail). Thanks for checking this :100:

PR #428, opened 2022-06-27
fix: use terramate.root.path.absolute

katcipis:
Using terramate.root.path.absolute instead of just terramate.root :-)


mariux (approved):
lgtm

PR #430, opened 2022-06-28
feat: add terramate.root.path.fs.basename/absolute

katcipis:
Renames terramate root path metadata + adds basename. In the end we have the following metadata available:

- `terramate.root.path.fs.absolute` : The absolute path of the root of the project.
- `terramate.root.path.fs.basename` : The basename of the root project path, which is the project name.


i4ki (approved):
lgtm

PR #432, opened 2022-06-28
fix: inconsistent evaluation of tm_try()

katcipis:


stack/globals_test.go
```diff
@@ -653,7 +653,28 @@ func TestLoadGlobals(t *testing.T) {
...
+						expr("b", `tm_try(global.undefined, global.a)`),
+						expr("c", `tm_try(global.a, global.undefined)`),
+					),
+				},
+			},
+			want: map[string]*hclwrite.Block{
+				"/stack": globals(
+					str("a", "value"),
+					str("b", "value"),
+					str("c", "value"),
```
mariux, line 672: :heart:

stack/globals_test.go
```diff
@@ -830,6 +851,19 @@ func TestLoadGlobals(t *testing.T) {
...
 			wantErr: errors.E(stack.ErrGlobalEval),
 		},
+		{
+			name:   "global interpolating undefined reference fails",
+			layout: []string{"s:stack"},
+			configs: []hclconfig{
+				{
+					path: "/stack",
+					add: globals(
+						str("a_interpolated", "${global.undefined}-something"),
```
mariux, line 861: :+1:

mariux (approved):
lgtm

this should not be a breaking change afaiu?

it's a change in behavior but stuff that worked before is still the same result?

i4ki (approved):
:rocket:

katcipis:
> lgtm
> 
> this should not be a breaking change afaiu?
> 
> it's a change in behavior but stuff that worked before is still the same result?

It should not be a breaking change, the only change in behavior is that code that used to fail before will now work, if you are pedantic it could be considered a breaking change, because maybe some people depended on the breaking/failing behavior, but I don't think we need to be that pedantic at this point =P

PR #437, opened 2022-06-29
chore: refactor hcl parser

i4ki:
Reuse the code for merging of attrs and blocks in the hcl APIs, so it makes easy to handle `imports` in a single place.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -1217,13 +1217,7 @@ func parseUnmergedBlocks(dir, blocktype string, validate blockValidator) (Blocks
 		return nil, errors.E("adding files to parser", err)
 	}
 
-	err = parser.parseSyntax()
-	if err != nil {
-		return nil, err
-	}
-
-	// TODO(i4k): improve API
-	err = parser.mergeConfig()
+	err = parser.MinimalParse()
 	if err != nil {
 		return nil, err
 	}
@@ -1231,7 +1225,6 @@ func parseUnmergedBlocks(dir, blocktype string, validate blockValidator) (Blocks
 	logger.Trace().Msg("Validating and filtering blocks")
 
 	blocks := filterBlocksByType(blocktype, parser.Blocks)
-
 	for _, block := range blocks {
 		if err := validate(block); err != nil {
 			return nil, errors.E(err, "validation failed")
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -20,7 +20,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/hashicorp/hcl/v2"
+	hhcl "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
@@ -105,19 +105,10 @@ type Stack struct {
 	Wants []string
 }
 
-// Blocks maps a filename to a slice of HCL blocks associated with it
-type Blocks map[string][]*hclsyntax.Block
-
-// GenFileBlocks maps filenames to slices of parsed generated_file blocks
-// found on the file.
-type GenFileBlocks map[string][]GenFileBlock
-
-// GenHCLBlocks maps filenames to slices of parsed generated_hcl blocks
-// found on the file.
-type GenHCLBlocks map[string][]GenHCLBlock
-
 // GenHCLBlock represents a parsed generate_hcl block.
 type GenHCLBlock struct {
+	// Origin is the filename where this block is defined.
+	Origin string
 	// Label of the block.
 	Label string
 	// Content block.
@@ -128,6 +119,8 @@ type GenHCLBlock struct {
 
 // GenFileBlock represents a parsed generate_file block
 type GenFileBlock struct {
+	// Origin is the filename where this block is defined.
+	Origin string
 	// Label of the block
 	Label string
 	// Content attribute of the block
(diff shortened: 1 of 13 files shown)


i4ki:
tests are failing because the ordering of errors matters -- working on this.

hcl/hcl.go
```diff
@@ -208,13 +212,92 @@ func (p *TerramateParser) AddFile(name string, data []byte) error {
...
+
+	type mergeHandler func(origin string, block *hclsyntax.Block)
+	mergeHandlers := map[string]mergeHandler{
+		"terramate": merged,
+		"globals":   merged,
+
+		"stack":         unmerged,
+		"generate_file": unmerged,
+		"generate_hcl":  unmerged,
+	}
```
katcipis, line 261:
```suggestion
	mergeHandlers := map[string]mergeHandler{
		"terramate": merged,
		"globals":   merged,
		"stack":         unmerged,
		"generate_file": unmerged,
		"generate_hcl":  unmerged,
	}
```

hcl/hcl.go
```diff
@@ -208,13 +212,92 @@ func (p *TerramateParser) AddFile(name string, data []byte) error {
...
+
+	fileMap := p.ParsedFiles()
+	for _, origin := range p.sortedFilenames() {
+		body := fileMap[origin]
```
katcipis, line 265:
```suggestion
	bodiesMap := p.ParsedFiles()
	for _, origin := range p.sortedFilenames() {
		body := bodiesMap[origin]
```

Because of fileMap I thought body was the contents of a file, not a parsed body.

hcl/hcl.go
```diff
@@ -208,13 +212,92 @@ func (p *TerramateParser) AddFile(name string, data []byte) error {
...
+	errs := errors.L()
+	for _, attrVal := range sortAttributes(other) {
+		if _, ok := p.MergedAttributes[attrVal.Name]; ok {
+			errs.Append(errors.E(ErrHCLSyntax,
+				attrVal.NameRange,
+				"attribute %q redeclared", attrVal.Name))
+			continue
+		}
+
+		p.MergedAttributes[attrVal.Name] = NewAttribute(origin, attrVal)
```
katcipis, line 296:
```suggestion
	for _, attr := range sortAttributes(other) {
		if _, ok := p.MergedAttributes[attr.Name]; ok {
			errs.Append(errors.E(ErrHCLSyntax,
				attrVal.NameRange,
				"attribute %q redeclared", attr.Name))
			continue
		}

		p.MergedAttributes[attr.Name] = NewAttribute(origin, attr)
```

hcl/hcl.go
```diff
@@ -977,239 +991,181 @@ func filterBlocksByType(blocktype string, blocks []*hclsyntax.Block) []*hclsynta
...
+
+	logger.Trace().Msg("Parse terramate sub blocks")
+
+	configBlock, ok := block.SubBlocks["config"]
+	if ok {
+		logger.Trace().Msg("Found config block.")
+
+		if tm.Config == nil {
+			tm.Config = &RootConfig{}
+		}
```
katcipis, line 1151:
```suggestion
		tm.Config = &RootConfig{}
```

If I'm not mistaken now that the blocks are already merged there is no need to keep checking if the field is nil (it won't be set multiple times anyway). Ignore if I missed something here.

katcipis (approved):
LGTM + some comments that could be addressed later (or plain wrong =P)

hcl/hcl_test.go
```diff
@@ -1348,7 +1348,7 @@ func TestHCLParserTerramateBlocksMerging(t *testing.T) {
 			},
 			want: want{
 				errs: []error{
-					errors.E(hcl.ErrTerramateSchema),
+					errors.E(hcl.ErrHCLSyntax),
```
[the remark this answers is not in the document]
i4ki: yep, this was a mistake. Thanks

stack/globals.go
```diff
@@ -223,36 +223,33 @@ func loadStackGlobalsExprs(rootdir string, cfgdir string) (*globalsExpr, error)
...
-		for _, fileblock := range fileblocks {
-			for name, attr := range fileblock.Body.Attributes {
-				if globals.has(name) {
-					return nil, errors.E(
-						ErrGlobalRedefined,
-						"%q redefined in %q", name, filename,
-					)
-				}
+	globalsBlock, ok := p.MergedBlocks["globals"]
+	if ok {
```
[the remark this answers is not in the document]
i4ki: we can't because we still need to check the parent directories.

hcl/hcl_test.go
```diff
@@ -46,227 +45,6 @@ type (
 	}
 )
 
-func TestHCLParserModules(t *testing.T) {
```
i4ki, line 49: moved to `tf` package.

hcl/hcl.go
```diff
@@ -33,15 +34,8 @@ import (
 const (
 	ErrHCLSyntax       errors.Kind = "HCL syntax error"
 	ErrTerramateSchema errors.Kind = "terramate schema error"
-	ErrTerraformSchema errors.Kind = "terraform schema error"
 )
 
-// Module represents a terraform module.
```
i4ki, line 39: moved to `tf` package.

hcl/merged_block.go
```diff
@@ -0,0 +1,141 @@
...
+	Type string
+
+	// Attributes are the block's attributes.
+	Attributes Attributes
+
+	// Raw is the list of original blocks that contributed to this block.
+	Raw hclsyntax.Blocks
+
+	// SubBlocks are the block's sub blocks.
+	SubBlocks map[string]*MergedBlock
```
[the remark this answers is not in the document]
katcipis: Hmm I'm not 100% on the naming either. Actually the main issue is the inconsistency, I was reading about Blocks having Blocks before and then a MergedBlock has a SubBlock...so my first instinct was that it was something different, but in the end it seems to be the same thing. So I would advocate for consistency, we could also go with SubBlocks and then use it consistently on other places (like the Block type). Unless I misunderstood something.

errors/list.go
```diff
@@ -150,6 +150,15 @@ func (l *List) Append(errs ...error) {
 	}
 }
 
+// AppendWrap is like Append() but wrap all errors with the provided kind.
+func (l *List) AppendWrap(kind Kind, errs ...error) {
```
katcipis, line 154: Hmmm maybe call it AppenKind and document that it adds the kind to all errors ? Wrapping seems misleading because we are not doing error wrapping in the sense that calling Unwrap will not return the Kind (the kind is not even an error, etc), so we are not wrapping up an error inside another. You could argue that you are wrapping the Kind field inside the error, and that is not necessarily wrong, but on Go idioms it seems misleading since wrapping is used about errors. But Im not sure, feel free to disregard.

hcl/hcl_cfg_run_test.go
```diff
@@ -438,7 +439,7 @@ func TestHCLParserConfigRun(t *testing.T) {
 			},
 			want: want{
 				errs: []error{errors.E(hcl.ErrTerramateSchema,
-					mkrange("cfg.tm", start(9, 15, 147), end(9, 31, 163)))},
+					mkrange("cfg.tm", start(9, 15, 147), end(9, 21, 153)))},
```
katcipis, line 442: The range was wrong before ?

katcipis (approved):
:rocket:

errors/list.go
i4ki: this is the exact wording we used when we designed the `errors` package...
https://github.com/mineiros-io/terramate/blob/main/errors/error.go#L182
when we call `errors.E(arg1, arg2, errlist)` we "wrap" all errors with the arguments provided to `E`.
It's not exactly the same meaning as Go's wrap (even though we support also "wrap" errlist in the Go sense, because we store errlist in the Error.Err which we return in the `Unwrap()` method.
So I think the "word" wrap applies and it's the actual concept we use, it's strange to avoid a word in this case.

I don't like `AppendKind()` because then it looks like we appending a kind to the errlist... which is not the case here.

katcipis (approved):

katcipis (approved):

PR #453, opened 2022-07-06
fix: terraform abspath() uses wd instead of scope basedir

i4ki:


stack/eval.go
```diff
@@ -31,11 +31,15 @@ type EvalCtx struct {
...
-func NewEvalCtx(rootdir string, sm Metadata, globals Globals) *EvalCtx {
-	evalctx := &EvalCtx{evalctx: eval.NewContext(sm.HostPath())}
-	evalctx.SetMetadata(rootdir, sm)
-	evalctx.SetGlobals(globals)
-	return evalctx
+func NewEvalCtx(rootdir string, sm Metadata, globals Globals) (*EvalCtx, error) {
+	evalctx, err := eval.NewContext(sm.HostPath())
+	if err != nil {
+		return nil, err
+	}
```
katcipis, line 38: Here since we are guaranteeing that the NewContext is always created with the stack host path which must always be valid maybe we could panic and keep the API as it was, seems like an internal/severe programming error if this failed.

katcipis (approved):
LGTM, but I think we could panic on the stack.EvalContext since it is more specific and it is guaranteed to be built always with the stack host path as the basedir, if that is invalid there issomething very wrong with Terramate code itself (not an user input etc).

PR #454, opened 2022-07-06
refactor: move stack create default handling to cli pkg

katcipis:
This will make it easier to use the new stack.Create function on tests instead of the old Init one, since for a lot of scenarios the default definitions are unexpected. Since these particular defaults seems to be CLI specific I moved them there.


i4ki (approved):
lgtm

PR #458, opened 2022-07-07
feat: add new git configs

katcipis:
Add new configs to git block:

```
check_untracked   = true
check_uncommitted = true
check_remote      = true
```

When the attributes are not defined in the git block they default to true. Also improved some attributes that had the wrong range when eval failed (wrong type).


hcl/hcl.go
```diff
@@ -1117,8 +1135,6 @@ func parseGitConfig(git *GitConfig, gitBlock *ast.MergedBlock) error {
 
 			git.DefaultBranch = value.AsString()
 		case "default_remote":
-			logger.Trace().Msg("Attribute name was 'default_remote'.")
-
 			if value.Type() != cty.String {
 				errs.Append(errors.E(attr.NameRange,
```
i4ki, line 1139:
```suggestion
				errs.Append(errors.E(attr.Expr.Range(),
```
not related to your change but just noticed this is wrong here.

katcipis: Extracted on a function trying to help being consistent about this.

hcl/hcl.go
```diff
@@ -1456,3 +1501,11 @@ func listTerramateDirs(dir string) ([]string, error) {
 func isTerramateFile(filename string) bool {
 	return strings.HasSuffix(filename, ".tm") || strings.HasSuffix(filename, ".tm.hcl")
 }
+
+func hclAttrEvalErr(attr *hclsyntax.Attribute, msg string, args ...interface{}) error {
```
i4ki, line 1505: top

i4ki (approved):
LGTM

PR #460, opened 2022-07-07
feat: add terramate.config.run.check_gen_code config

katcipis:
Add new configs to run block:

```
check_gen_code = true
```

When the attribute is not defined in the run block it defaults to true.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -50,6 +50,10 @@ type Config struct {
 
 // RunConfig represents Terramate run configuration.
 type RunConfig struct {
+	// CheckGenCode enables generated code is up-to-date check on run.
+	CheckGenCode bool
+
+	// Env contains environment definitions for run.
 	Env *RunEnv
 }
 
@@ -1043,7 +1047,9 @@ func parseRootConfig(cfg *RootConfig, block *ast.MergedBlock) error {
 	if ok {
 		logger.Trace().Msg("Type is 'run'")
 
-		cfg.Run = &RunConfig{}
+		cfg.Run = &RunConfig{
+			CheckGenCode: true,
+		}
 
 		logger.Trace().Msg("Parse run config.")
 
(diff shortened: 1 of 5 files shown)


hcl/hcl.go
```diff
@@ -1062,8 +1068,30 @@ func parseRunConfig(runCfg *RunConfig, runBlock *ast.MergedBlock) error {
...
+				"failed to evaluate terramate.config.run.%s attribute", attr.Name,
+			))
+
+			continue
+		}
+
+		switch attr.Name {
+		case "check_gen_code":
+			if value.Type() != cty.Bool {
+				errs.Append(errors.E(attr.Expr.Range(),
```
katcipis, line 1083: Can use the new error creation function after we merge #458

test/hcl.go
```diff
@@ -106,6 +106,12 @@ func assertTerramateRunBlock(t *testing.T, got, want *hcl.RunConfig) {
 		return
 	}
 
+	if want.CheckGenCode != got.CheckGenCode {
```
i4ki, line 109: now we can do:

```
assert.IsTrue(t, want.CheckGenCode == got.CheckGenCode, 
    "want.Run.CheckGenCode %v != got.Run.CheckGenCode %v", want.CheckGenCode, got.CheckGenCode)
```
but just an idea... feel free to ignore.

i4ki (approved):
lgtm

test/hcl.go
katcipis: Forgot we now have this, wiill do, thanks :heart:

katcipis: I even looked for it but I used go doc and I think go doc showed me the version on the go.mod, which doesn't have this, so I thought I was just going crazy :laughing:

i4ki (approved):

hcl/hcl.go
katcipis: Yeah I just fix it then, already working on it here :+1:

PR #461, opened 2022-07-08
feat: add git checks config through cfg and env

katcipis:
Just as we can pass flags:

- `--disable-check-git-untracked`
- `--disable-check-git-uncommitted`
- `--disable-check-git-remote`
- `--disable-check-gen-code`

We now can have them as project configs:

```hcl
terramate {
  config {
    git {
      check_untracked   = true
      check_uncommitted = true
      check_remote      = true
    }
    run {
      check_gen_code = true
    }
  }
}
```

And environment variables:

- `TM_DISABLE_CHECK_GIT_UNTRACKED`
- `TM_DISABLE_CHECK_GIT_UNCOMITTED`
- `TM_DISABLE_CHECK_GIT_REMOTE`
- `TM_DISABLE_CHECK_GEN_CODE`

A env variable is considered `true` if set and not `0` nor `false`.

Precedence:
- command line
- ENV (needs to be actually set to be considered) 
- config
- built-in default

Docs will be added on a following PR.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -939,6 +939,14 @@ func (c *cli) checkGenCode() bool {
 		}
 	}
 
+	cfg := c.prj.rootcfg
+
+	if cfg.Terramate != nil &&
+		cfg.Terramate.Config != nil &&
+		cfg.Terramate.Config.Run != nil {
+		return cfg.Terramate.Config.Run.CheckGenCode
+	}
+
 	return true
 }
 
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -405,6 +405,19 @@ func (c *cli) generate() {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -405,6 +405,50 @@ func (c *cli) generate() {
...
+		if envVarIsSet(disableCheck) {
+			return false
+		}
+	}
+
+	cfg := c.prj.rootcfg
+
+	if cfg.Terramate != nil &&
+		cfg.Terramate.Config != nil &&
+		cfg.Terramate.Config.Git != nil {
```
i4ki, line 423: we should improve those nil checks later...Now that we have default values in the config, maybe it doesn't need to be a pointer anymore. But we can improve this later.

i4ki (approved):
:rocket:

cmd/terramate/cli/cli.go
katcipis: 👍

PR #471, opened 2022-07-13
feat: add initial stack.Clone

katcipis:
Add initial stack.Clone function that clones stacks. This first version only does a plain clone not handling when the stack being cloned has an ID, this will be added in the next PR. It does all basic validations + it ignores dotfiles/dotdirs.


stack/clone_test.go
```diff
@@ -0,0 +1,151 @@
...
+func TestStackClone(t *testing.T) {
+	type testcase struct {
+		name    string
+		layout  []string
+		src     string
+		target  string
+		wantErr error
+	}
+
+	testcases := []testcase{
```
katcipis, line 38: For now tests are quite simple since we always check that the entire subtrees are identical. For checking for the "almost" identical but with different ID will be tricky =P (also some other scenarios where the stack has other configs inside etc...).

stack/clone.go
```diff
@@ -0,0 +1,105 @@
...
+	// ErrCloneTargetDirExists indicates that the target dir on a clone
+	// operation already exists.
+	ErrCloneTargetDirExists errors.Kind = "clone target dir exists"
+)
+
+// Clone will clone the stack at srcdir into targetdir.
+//
+// - srcdir must be a stack (fail otherwise)
+// - targetdir must not exist (fail otherwise)
+// - All files and directories are copied  (except dotfiles/dirs)
```
i4ki, line 36:
```suggestion
// - All files and directories are copied  (except dotfiles/dotdirs)
```

stack/clone.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := os.MkdirAll(targetdir, createDirMode); err != nil {
+		return errors.E(err, "creating target dir")
+	}
+
+	for _, entry := range entries {
+		if strings.HasPrefix(entry.Name(), ".") {
+			continue
+		}
+		if entry.IsDir() {
+			srcdir := filepath.Join(srcdir, entry.Name())
```
i4ki, line 75: you can create those variables before the `if` and reuse them for copying dir and file.

```
    srcpath := filepath.Join(srcdir, entry.Name())
    dstpath := filepath.Join(dstdir, entry.Name())
    if entry.IsDir() {
        if err := copyDir(dstpath, srcpath); err != nil {
			return errors.E(err, "copying src to target dir")
		}
		continue
    }
    if err := copyFile(dstpath, srcpath); err != nil {
		return errors.E(err, "copying src to target file")
	}
```

stack/clone.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// Clone will clone the stack at srcdir into targetdir.
+//
+// - srcdir must be a stack (fail otherwise)
+// - targetdir must not exist (fail otherwise)
+// - All files and directories are copied  (except dotfiles/dirs)
+// - If cloned stack has an ID it will be adjusted to a generated UUID.
+// - If cloned stack has no ID the cloned stack also won't have an ID.
+func Clone(rootdir, targetdir, srcdir string) error {
```
i4ki, line 39: I think `dstdir` works better with `srcdir`.

i4ki (approved):
lgtm but left some improvement ideas.
Just thinking now if makes sense to use `feat: ` for those intermediate PRs because then we will have repeated `feat` entries for each feature...
In the last release announcement I ignored the intermediate ones to have a concise changelog. What do you think of using `feat` just for the PR that actually exports the feature to be used?

stack/clone.go
katcipis: True true (networking feelings =P)

i4ki (approved):
:rocket:

PR #474, opened 2022-07-14
refactor: renaming and adding new test on hcl parser

katcipis:
# Reason for This Change

As I searched for something that would help me implement the stack clone feature I found some opportunities to make things clearer, at least to me (always debatable if it is better of course).

## Description of Changes

Renamed a few methods to make things more clear. I think it makes sense that a Parse method only parses the HCL itself, without any Terramate specific schema validations, and then a ParseConfig one is a more specific version that will do schema validation and return a more high level Terramate configuration (minimal parse was confusing to me, but that is debatable).

Also make a few other fields more explicit on their names and added a test for the ParseBodies method, I think it will be quite useful on the stack clone so I just double checked it does what I need.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -164,21 +164,23 @@ type PartialEvaluator func(hclsyntax.Expression) (hclwrite.Tokens, error)
 // this API allows you to define the exact set of files (and contents) that are
 // going to be included in the final configuration.
 type TerramateParser struct {
+	// MergedAttributes are the top-level attributes of all files.
+	// This will be available after calling Parse or ParseConfig
+	MergedAttributes ast.Attributes
+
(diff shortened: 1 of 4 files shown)


hcl/hcl.go
```diff
@@ -361,12 +364,25 @@ func (p *TerramateParser) MinimalParse() error {
 	return errs.AsError()
 }
 
+// ParsedBodies returns a map of filename to the parsed hclsyntax.Body.
+func (p *TerramateParser) ParsedBodies() map[string]*hclsyntax.Body {
```
katcipis, line 368: While I was starting to add the new map of parsed blocks I realized we already have this method, should be enough for the cloning feature :100:

i4ki (approved):
lgtm

PR #475, opened 2022-07-15
chore: add stack clone id handling

katcipis:
When we clone a stack that has an ID now we update the cloned stack with a new UUID. This is done using arcane lexing in YOLO mode because we need to keep all comments and keeping all comments after parsing HCL is not possible (at least we don't know how yet).


hcl/lex/lex.go
```diff
@@ -0,0 +1,137 @@
...
+func TokenEquals(token1, token2 *hclwrite.Token) bool {
+	return token1.Type == token2.Type &&
+		string(token1.Bytes) == string(token2.Bytes)
+}
+
+// FindTokenSequence finds the first match of the given token sequence
+// on the given tokens, returning the position of the last token of the
+// matched sequence, so the next token (if any) would be what is after the
+// given sequence. The contents of the tokens are also matched, not only their type.
+func FindTokenSequence(tokens hclwrite.Tokens, seq ...*hclwrite.Token) (int, bool) {
```
i4ki, line 58: For me the design of the algorithm here and the API are quite confusing...
Why we don't treat like a FindSubst() kind of function? that returns the position of first element that matched the whole sequence? Then the caller just knows it needs to skip the size of seq. Example:

```
    begin, ok := FindTokenSequence(tokens, stack, obrace)
    if !ok { }
    stackTokens := tokens[begin+2:] // or tokens[begin+len(seq)]
```
The above API is more useful for other parser tasks.

Regarding the algorithm, it looks more complex than needed, no? Or maybe I'm missing something... The `matchHead`, `pendingMatches` and comparing `TokenEquals()` in both loops seem unneeded for a substr-like function.

What about something like that?

```
func FindTokenSequence(tokens hclwrite.Tokens, seq ...*hclwrite.Token) (int, bool) {
	if len(seq) == 0 {
		return 0, false
	}
outer:
	for i := 0; i < len(tokens); i++ {
		found := false
		for j := 0; j < len(seq); j++ {
			if !TokenEquals(tokens[i+j], seq[j]) {
				continue outer
			}
			found = true
		}
		if found {
			return i + len(seq) - 1, true
		}
	}
	return 0, false
}
```
even though I'd rather return `i` (the position of the first matched token) instead of last... The algorithm above can be way simpler with the other API (I guess).

but no strong feeling, just found it a little odd/different than what I was expecting.
Feel free to ignore, or maybe improve later.

stack/clone.go
```diff
@@ -101,3 +127,104 @@ func copyFile(destfile, srcfile string) error {
...
+	stackFilePath, body := getStackBody(parser)
+	if body == nil {
+		return errors.E("updating stack ID: stack block not found")
+	}
+
+	// WHY oh WHY do you ask ? Parsing HCL always delivers an AST that
+	// has no comments on it, so building a new HCL file from the parsed
+	// AST will lose all comments from the original code.
+	//
+	// - https://raw.githubusercontent.com/katcipis/memes/master/satan.jpg
```
i4ki, line 163: :laughing: :laughing: :laughing:

stack/clone.go
```diff
@@ -101,3 +127,104 @@ func copyFile(destfile, srcfile string) error {
...
+	if !ok {
+		return errors.E(err, "cloned stack doesn't have stack ID")
+	}
+
+	logger.Trace().Msg("updating id attribute")
+
+	// Here we assume that stack IDs are also on the form:
+	// id = "id"
+	// Id's are very constrained and are always strings, so we expect
+	// the tokens: TokenOQuote + TokenQuotedLit + TokenCQuote
```
i4ki, line 206: theoretically... we documented IDs as being strings, then there's a lot of valid cases out there...

Expressions can be parenthesized without any semantic changes, so the examples below are valid ids, no?

```
id = ("id")
id = (("id"))
```

pure interpolations are automatically handled by hcl expr.Value(nil) evaluator.. so the following is also valid?

```
id = "${"part 1"} part 2"
```

i4ki (approved):
I left some questions and things to discuss but feel free to address them in the future.

hcl/lex/lex.go
katcipis: Yeah I had a feeling that it was way too complicated at the time but I was trying to figure out a bunch of different stuff and was not able to come up with the simpler solution and just got happy as soon as it worked :laughing: , will adjust :+1:

katcipis: Told you multiple times I don't have much experience with parsers :laughing:

katcipis: Will just adjust the code a little, I think it can be even simpler and the current code panics when acessing tokens[i+j] (it can be out of bounds on some cases)

stack/clone.go
katcipis: hmm if after eval they satisfy the ID constraints I think they would be, so maybe there are more cases to handle here then.

katcipis: maybe we can add support for other forms of defining the stack ID on follow up PRs, the most common cases today is no ID or the UUID we generate (plain string). so we can have something usable soon and improve (the command itself will be experimental too).

i4ki (approved):
:rocket: :rocket: :rocket: :rocket:

katcipis:
> rocket rocket rocket rocket

thanks again for spotting my very junior/amateur mistakes :rofl:

PR #480, opened 2022-07-18
feat: allow stacks inside stacks

i4ki:


cmd/terramate/e2etests/list_test.go
```diff
@@ -95,6 +96,41 @@ func listTestcases() []testcase {
...
+				"s:mineiros.io/departments/engineering",
+				"s:mineiros.io/departments/accounting",
+				"s:mineiros.io/departments/engineering/terramate",
+				"s:mineiros.io/departments/engineering/terraform-modules",
+				"d:mineiros.io/departments/engineering/docs",
+				"d:mineiros.io/departments/engineering/tests",
+				"s:mineiros.io/departments/engineering/tests/e2e",
+			},
+			want: runExpected{
+				Stdout: strings.Join([]string{
```
katcipis, line 123: Noice :100: , we could even have a stackList func or something that always does the joining with \n and accepts a vargs, so it would look like:

```go
Stdout: stackList(
   "stack1",
   "stack2",
)
```

cmd/terramate/e2etests/run_test.go
```diff
@@ -78,6 +78,27 @@ frita
 `,
 			},
 		},
+		{
+			name: "independent stacks inside other stacks gives consistent ordering (lexicographic by path)",
```
katcipis, line 82: I imagine this will change on the next PR when we introduce the ordering changes (nice to do in multiple steps)

katcipis (approved):
:rocket: :rocket: :rocket:

cmd/terramate/e2etests/run_test.go
i4ki: yep

cmd/terramate/e2etests/list_test.go
i4ki: yep I thought about this, but introducing this make sense update all tests. I can do this in a separate PR.

PR #483, opened 2022-07-18
chore: improve stack clone update

katcipis:
Using hclwrite and some minimal token manipulation there is a much better way to replace the attribute. Sorry for the brainfart on the previous approach, I did something very similar for the formatting on hcl/fmt.go but apparently Im stupid x_x.


i4ki (approved):
:rocket: 

yay much better!! Thanks!

PR #484, opened 2022-07-19
feat: add filesystem hierarchical order to stack orchestration

i4ki:


cmd/terramate/e2etests/run_test.go
```diff
@@ -385,6 +399,40 @@ stack-z
 				StderrRegex: string(dag.ErrCycleDetected),
 			},
 		},
+		{
+			name: "child stack can have explicit after clause to parent",
+			layout: []string{
+				`s:stacks`,
+				`s:stacks/child:after=["../"]`,
```
katcipis, line 406:
```suggestion
				`s:stacks/child:after=["/stacks"]`,
```

to be consistent with the other tests that also use the abspath. Feel free to ignore.

docs/orchestration.md
```diff
@@ -60,17 +60,31 @@ selected for execution, should always be:
...
-To help with that terramate provides a way to explicit declare
-the desired order of execution between stacks.
+To help with that terramate provides two ways to define the desired order of
+execution between stacks.
 
+### Filesystem hierarchical order
 
-### Defining Order Of Execution
+By creating an hierarchy of stacks in the filesystem will make Terramate execute
+the parent stacks before the children ones.
```
katcipis, line 69:
```suggestion
Creating an hierarchy of stacks in the filesystem will make Terramate execute
parent stacks before their children.
```

docs/orchestration.md
```diff
@@ -60,17 +60,31 @@ selected for execution, should always be:
...
 
-Order of execution is declared inside the **stack** block using the
-field **before** and **after**. 
+This is a double-edged sword because it allows you to move stacks around to
+better resemble the cloud infrastructure and the order of which things must
+be executed with zero code change but in counterpart beware that moving stacks 
+just for cosmetic purposes can potentially change order and break stuff.
```
katcipis, line 74:
```suggestion
just for cosmetic purposes can potentially change the execution order and break things.
```

docs/orchestration.md
```diff
@@ -60,17 +60,31 @@ selected for execution, should always be:
...
 
-Each field is a set of strings (**set(string)**),
-where each string is a reference to another stack.
+### Explicit Order Of Execution
+
+Order of execution can be explicitly declared inside the **stack** block using
+the field **before** and **after**.
```
katcipis, line 79:
```suggestion
the fields **before** and **after**. 
```

docs/orchestration.md
```diff
@@ -60,17 +60,31 @@ selected for execution, should always be:
...
+Order of execution can be explicitly declared inside the **stack** block using
+the field **before** and **after**. 
+
+Each field is a set of strings (**set(string)**), where each string is a 
+reference to another stack.
+
+The explicit order can be used in conjunction with the implicit filesystem order
+but have in mind that a parent stack in the filesystem can never execute after a
+child one, and trying to make this using explicit **before** and **after** clauses
+will lead to a cycle error.
```
katcipis, line 87: nice

run/order.go
```diff
@@ -42,6 +45,21 @@ func Sort(root string, stacks []stack.S) ([]stack.S, string, error) {
...
+		for _, other := range stacks {
+			if stack.Path() == other.Path() {
+				continue
+			}
+
+			if strings.HasPrefix(stack.Path(), other.Path()+string(os.PathSeparator)) {
+				other.SetImplicitBeforePath(stack.Path())
+			}
+		}
+	}
```
katcipis, line 61:
```suggestion
	for _, parent := range stacks {
		for _, child := range stacks {
			if parent.Path() == child.Path() {
				continue
			}

			if strings.HasPrefix(parent.Path(), child.Path()+string(os.PathSeparator)) {
				child.SetImplicitBeforePath(parent.Path())
			}
		}
	}
```

They are not always parent/child, but we are looking for parent/child relationships and it was a little hard at first for me to get which was the parent/child, so maybe renaming would improve things.

stack/stack.go
```diff
@@ -55,6 +55,10 @@ type (
 		// before is a list of stack paths that must run after this stack.
 		before []string
 
+		// implicitBefore is a map of paths that must run before this stack but
+		// they were implicitly computed from the project hierarchy.
+		implicitBefore map[string]struct{}
```
katcipis, line 60: why do we need a separate implicit before instead of just adding on the stacks before ? the errors in general are the samy anyway no ? like cycles etc ? (I imagine Im missing something, so just curious).

stack/stack.go
```diff
@@ -85,6 +89,9 @@ type (
 		// RelPathToRoot is the relative path from the stack to root.
 		RelPathToRoot() string
 	}
+
+	// List of stacks.
+	List []S
```
katcipis, line 94: Noice :100:

katcipis (dismissed):
LGTM. I just don't get why we need an extra SetImplicitBeforePath method, keeping the implicit before on a different structure and then adding when getting Before(). It seems that having something like AppendBefore or AddBefore that just allows the caller to add before stacks would be enough, because in the end the implicit before from the hierarchy is just a normal before internally (implicit, but just a before nonetheless) :thinking:

run/order.go
i4ki: yes but as you said, calling them `parent` and `child` is not correct... they are just stacks of the same list being compared. Maybe calling them `s1` and `s2` improves?

i4ki:
I avoided direct mutation of Before() because (I think) it's useful for troubleshooting order issues to have them in a separate field, because otherwise either we lose the information or we need to create a "StackRef" type containing the path and an origin/kind/etc field telling from where this entry comes from.

Will think more about this.

stack/stack.go
katcipis: hmm I feel you...but I find it a little odd that we have extra logic that is unused, nothing on the code uses this difference between implicit/normal before, as soon as I saw it I kept looking for where it is used..and the fact that it is not used feels like unnecessary complexity (extra abstractions/code). But if you are confident it will be useful in the future for improved diagnostics it is all good :+1: 

On directly mutating before, we can have a method to add the Befores (instead of direct mutation) , because in the end mutation happens anyway in the sense that calling Before() returns the mutated (merged) befores, so it already has some mutation complexity but in a way that seems harder to debug, like someone debugging this may log the before stacks and forget about implicitBefore, inspecting before VS Before() will produce different results, etc. I would be more prone to just modeling on the stack a way to clearly add before stacks (but not necessarily exposing the slice directly) and then use that on every place that needs before stacks to be added (the implicit and also when loading config, etc).

But it is just an idea and my initial perception, feel free to disregard.

katcipis (approved):
:rocket:

i4ki:
I will change to directly mutate "before".

PR #487, opened 2022-07-20
feat: add automatic code gen on stack create

katcipis:
# Reason for This Change

On terramate create we want any pending code generation to already be done on the created stack.

## Description of Changes

Add a code generation step after the stack creation.

### DIFF
--- a/cmd/terramate/e2etests/create_test.go
+++ b/cmd/terramate/e2etests/create_test.go
@@ -30,11 +30,12 @@ func TestCreateStack(t *testing.T) {
 	cli := newCLI(t, s.RootDir())
 
 	const (
-		stackID          = "stack-id"
 		stackName        = "stack name"
 		stackDescription = "stack description"
 		stackImport1     = "/core/file1.tm.hcl"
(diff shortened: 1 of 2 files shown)


cmd/terramate/e2etests/create_test.go
```diff
@@ -61,8 +69,12 @@ func TestCreateStack(t *testing.T) {
 			"--import", stackImport2,
 		)
 
+		t.Logf("run create stack %s", stackPath)
```
katcipis, line 72: This helped me debug some issues with duplicated IDs on the test...thinking if maybe we should move some t.Logf inside cli.run, so everytime a test involving cli fails we get a full log of all commands and their output etc... WDYT ?

i4ki: hmm not sure because this could generate a lot of noise for cases where the got vs want diff is enough...
let's discuss this in other channels and improve separate if needed.

i4ki (dismissed):
lgtm but there are some conflicts.

i4ki (approved):
:rocket:

PR #490, opened 2022-07-20
feat: add --no-recursive flag to run command

i4ki:


cmd/terramate/cli/cli.go
```diff
@@ -93,9 +93,10 @@ type cliSpec struct {
 	} `cmd:"" help:"List stacks"`
 
 	Run struct {
-		DisableCheckGenCode   bool     `optional:"true" default:"false" help:"Disable outdated generated code check"`
-		DisableCheckGitRemote bool     `optional:"true" default:"false" help:"Disable checking if local default branch is updated with remote"`
```
i4ki, line 97: they are optional by default.

cmd/terramate/e2etests/run_test.go
```diff
@@ -1392,6 +1392,57 @@ func TestRunContinueOnError(t *testing.T) {
 	})
 }
 
+func TestRunNoRecursive(t *testing.T) {
+	s := sandbox.New(t)
+
+	s.BuildTree([]string{
+		`f:file.txt:root`,
```
katcipis, line 1399: Why the file on root ? :thinking:

katcipis (dismissed):
:1st_place_medal:

cmd/terramate/e2etests/run_test.go
i4ki: see the last `cli.run()` invocation at line 1438.

katcipis: hmm but we never run commands on non stack dirs anyway so it seems we are testing something else :thinking:. If the ---no-recursive flag didn't work the result would be to run cat on the stacks only (all stacks), not on the root anyway since it is not a stack, so I still don't get why root needs a file :thinking:, unless the idea here is to test that run never executes anything on non-stack dirs, but this seems unrelated to the  test :thinking:.

If the idea is to indeed validate that non-stacks never execute anything on this test we could use our test cmd, if you want to be sure that the command won't be executed at all without requiring creating files. Also would make sense to test the behavior when it is recursive but there are no stacks on the tree, which makes me think, when just calling terramate run on a subtree that has no stacks it will only do nothing (no fail), so isn't it inconsistent that with --no-recursive if the current dir is not a stack it fails ? In the recursive mode, if no stack is found we are OK with it :thinking: (not sure, just thinking out loud).

katcipis (approved):

PR #491, opened 2022-07-21
fix: codegen should generate empty files

katcipis:
# Reason for This Change

We want both generate_hcl and generate_file to generate empty files when content is empty, instead of the current behavior of not generating any file.

## Description of Changes

Both code generation and outdated code detection are fixed to handle empty files.


i4ki (approved):
lgtm

PR #492, opened 2022-07-21
chore: remove mac 10.15 from CI

katcipis:
Remove "macos-10.15" from the CI, it started to get cancelled all the time by GHA, maybe they dropped support to it or they are going through some issues with it. We could add it back if it starts working again.


i4ki (dismissed):
lgtm

i4ki (approved):
lgtm haha

PR #493, opened 2022-07-21
feat: add support for directories in the stack order

i4ki:
This PR adds support for using directories in the after/before clauses of the stack block.
This allows for a more flexible mechanism for defining stack ordering because this way users can organize similar stacks inside a common directory and use this directory as a reference in the order.

### DIFF
--- a/cmd/terramate/e2etests/run_test.go
+++ b/cmd/terramate/e2etests/run_test.go
@@ -66,6 +66,15 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
+			name: "after regular file",
+			layout: []string{
+				fmt.Sprintf("s:stack:after=[%q]", test.WriteFile(t, "", "test.txt", `bleh`)),
+			},
+			want: runExpected{
+				Stdout: listStacks(`stack`),
+			},
+		},
+		{
 			name: "independent stacks, consistent ordering (lexicographic)",
 			layout: []string{
 				"s:batatinha",
--- a/cmd/terramate/e2etests/run_test.go
+++ b/cmd/terramate/e2etests/run_test.go
@@ -57,6 +57,15 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
+			name: "after non-existent path",
+			layout: []string{
+				fmt.Sprintf("s:stack:after=[%q]", test.NonExistingDir(t)),
+			},
+			want: runExpected{
+				Stdout: listStacks(`stack`),
+			},
+		},
+		{
 			name: "independent stacks, consistent ordering (lexicographic)",
 			layout: []string{
 				"s:batatinha",
@@ -541,7 +550,7 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
-			name: "before directory containing no stacks",
+			name: "before directory containing no stacks does nothing",
 			layout: []string{
 				`d:dir`,
 				`d:dir/dir2`,
@@ -553,7 +562,7 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
-			name: "after directory containing no stacks",
+			name: "after directory containing no stacks does nothing",
 			layout: []string{
 				`d:dir`,
 				`d:dir/dir2`,
--- a/cmd/terramate/e2etests/run_test.go
+++ b/cmd/terramate/e2etests/run_test.go
@@ -535,6 +535,32 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
+			name: "after stack containing sub-stacks",
+			layout: []string{
+				`s:parent`,
+				`s:parent/s1`,
+				`s:parent/s2`,
+				`s:parent/s3`,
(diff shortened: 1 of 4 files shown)


cmd/terramate/e2etests/run_test.go
```diff
@@ -471,13 +489,14 @@ func TestCLIRunOrder(t *testing.T) {
 			},
 		},
 		{
-			name: "child stack can have explicit after clause to parent",
+			name: "child stack CANNOT have explicit after clause to parent",
```
katcipis, line 492:
```suggestion
			name: "child stack cannot have explicit after clause to parent",
```

or maybe:

```suggestion
			name: "child stack with explicit after clause to parent fails",
```

Using all upper case seems quite off from how most tests are described on the project.

i4ki: it's because it was allowed before... tried to make it explicit the change. But I will change.

docs/orchestration.md
```diff
@@ -78,31 +78,33 @@ just for cosmetic purposes can potentially change the execution order and break
 Order of execution can be explicitly declared inside the **stack** block using
 the fields **before** and **after**. 
 
-Each field is a set of strings (**set(string)**), where each string is a 
-reference to another stack.
+Each field is a set of string paths (**set(string)**), where each string is a 
+reference to another directory, which can be a stack or contain stacks inside.
```
katcipis, line 82:
```suggestion
Each field is a set of string (**set(string)**), where each string is a path that
references another directory, which can be a stack or contain stacks inside.
```

docs/orchestration.md
```diff
@@ -78,31 +78,33 @@ just for cosmetic purposes can potentially change the execution order and break
...
 
 ```
-/path/relative/to/project/root/stack
+/path/relative/to/project/root/dir
 ```
 
 **before** ensures that the configured stack is executed before the
-listed stacks, as the stack you are saying "I execute before these stacks".
+stacks contained in the directory entries. As the stack you are saying: 
+"I execute before these stacks".
```
katcipis, line 103:
```suggestion
stacks contained in the directory paths. As the stack you are saying: 
"I execute before all stacks inside these directories".
```

docs/orchestration.md
```diff
@@ -78,31 +78,33 @@ just for cosmetic purposes can potentially change the execution order and break
...
 
-**after** ensures the opposite, that the listed stacks are executed before
-the configured stack, you are saying "I execute after these stacks".
+**after** ensures the opposite, that the stacks contained in the provided 
+directories are executed before the current stack. As the stack, you are saying:
+"I execute after these stacks".
```
katcipis, line 107:
```suggestion
"I execute after all stacks inside these directories".
```

run/order.go
```diff
@@ -145,22 +146,50 @@ func BuildDAG(
 
 	visited[s.Path()] = struct{}{}
 
-	logger.Trace().
-		Msg("Load all stacks in dir after current stack.")
-	afterStacks, err := loader.LoadAll(root, s.HostPath(), s.After()...)
+	cleanOrderPaths := func(fieldname string, paths []string) []string {
```
katcipis, line 149:
```suggestion
	orderAbsPaths := func(fieldname string, paths []string) []string {
```

Or:

```suggestion
	normalizePaths := func(fieldname string, paths []string) []string {
```

Since you are normalizing all paths (rel/abs to project) to paths absolute to the host. Feel free to disregard.

run/order.go
```diff
@@ -145,22 +146,50 @@ func BuildDAG(
...
+				cleanpaths = append(cleanpaths, path)
+			}
+		}
+		return cleanpaths
+	}
+
+	afterPaths := cleanOrderPaths("after", s.After())
+	beforePaths := cleanOrderPaths("before", s.Before())
+
+	logger.Trace().Msg("Load all stacks in dir after current stack.")
```
katcipis, line 177:
```suggestion
	logger.Trace().Msg("load all stacks in dir after current stack")
```

I think the final dot on the log messages don't help much, lately I have been avoiding them.

katcipis (dismissed):
:rocket:

run/order.go
i4ki: it's actually not normalizing neither making them absolute.. but just filtering out the ones with error (no such file, not directory, no permission, etc).

run/order.go
i4ki: but logs are always capitalized, no?

run/order.go
```diff
@@ -145,22 +146,50 @@ func BuildDAG(
...
+			if err != nil {
+				logger.Warn().
+					Err(err).
+					Msgf("failed to stat %q path %q - ignoring", fieldname, abspath)
+			} else if !st.IsDir() {
+				logger.Warn().
+					Msgf("stack.%s path %q is not a directory - ignoring",
+						fieldname, path)
+			} else {
+				cleanpaths = append(cleanpaths, path)
```
katcipis, line 168: I thought the abspath was appended here, my mistake

katcipis (approved):
:rocket: :rocket: :rocket:

PR #499, opened 2022-07-26
refactor: clone: remove unnecessary lex complexity

katcipis:
# Reason for This Change

Simplify the stack clone code. No need to use lex/tokens for simple strings.

## Description of Changes

Just use cty.Value instead of lex when setting the new stack ID.


stack/clone.go
```diff
@@ -153,16 +152,14 @@ func updateStackID(stackdir string) error {
 
 	logger.Trace().Msg("finding file containing stack definition")
 
-	stackFilePath, body := getStackBody(parser)
-	if body == nil {
-		return errors.E("updating stack ID: stack block not found")
+	stackFilePath := getStackFilepath(parser)
```
i4ki, line 155: maybe call this `stackParsedFile`?

i4ki (approved):
lgtm

stack/clone.go
katcipis: hmm but it is not returning a parsed file :thinking: . Maybe stackBlockFilepath ? wanted to make it clear that it returns the filepath of where the stack block is defined. What is being "get" is kinda hard to explain with a small name :laughing:

i4ki: gotcha! I thought it was returning the hcl.File object.

PR #505, opened 2022-08-01
fix: iterator variable not allowed in the tm_dynamic.labels attribute

i4ki:


katcipis (dismissed):
:rocket: :rocket: :rocket:

katcipis (approved):

PR #513, opened 2022-08-02
feat: add terramate experimental vendor

katcipis:
# Reason for This Change

Adding new `terramate experimental vendor` that will help downloading and vendoring Terraform modules.

## Description of Changes

The overall idea of the design is to implement a very basic first version of vendor but in a way that is easy to extend to things like vendoring all source references in a project etc.

We assessed how Terraform does this using go-getter. [They don't seem very happy with go-getter](https://github.com/hashicorp/terraform/blob/d8c14c4295f4990faf2df90ea6f7b90193eb3f72/internal/getmodules/getter.go#L21) and they don't provide parsing for their specific format of module source, so it is all in in the go-getter behavior for retrieving things or nothing (no parsing to re-use here).

Given that we want a simpler subset of that go-getter does I implemented a very simple parsing just for what we need. The getting is quite simple since we only want git, so it is basically git clone.

## Usage/test script

```
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

terramate {
  config {
  }
}
EOM

terramate experimental vendor download github.com/mineiros-io/terramate main
terramate experimental vendor download git@github.com:mineiros-io/terramate.git v0.1.19
terramate experimental vendor download git::https://github.com/mineiros-io/terramate.git v0.1.18
terramate experimental vendor download git::ssh://git@github.com/mineiros-io/terramate.git v0.1.17

echo "vendor:"
tree -d -L 4 vendor

echo "checking VERSION files of each vendored repo"
cat vendor/github.com/mineiros-io/terramate/main/VERSION
cat vendor/github.com/mineiros-io/terramate/v0.1.19/VERSION
cat vendor/github.com/mineiros-io/terramate/v0.1.18/VERSION
cat vendor/github.com/mineiros-io/terramate/v0.1.17/VERSION
```

The final tree looks like:

```
vendor
└── github.com
    └── mineiros-io
        └── terramate
            ├── main
            ├── v0.1.17
            ├── v0.1.18
            └── v0.1.19
```


modvendor/modvendor.go
```diff
@@ -0,0 +1,221 @@
...
+	return clonedir, nil
+}
+
+// ParseSource parses the given modsource string.
+// The modsource must be a valid Terraform Git/Github source reference as documented in:
+//
+// - https://www.terraform.io/language/modules/sources
+//
+// Source references that are not Git/Github are not supported.
+func ParseSource(modsource string) (Source, error) {
```
katcipis, line 144: The idea is to use ParseSource in the future when we are parsing all module.source references from code and then we only vendor the ones we support.

git/git.go
```diff
@@ -103,31 +103,6 @@ const (
...
-		Str("action", "NewConfigWithPath()").
-		Str("path", programPath).
-		Msg("Make new config.")
-	config := NewConfig(username, email)
-	config.ProgramPath = programPath
-	return config
-}
-
-// NewWrapper creates a new wrapper.
-func NewWrapper(user, email string) (*Git, error) {
```
katcipis, line 127: Only new wrapper was used on tests, these functions confused me on how was I supposed to create a git wrapper.

git/git_test.go
```diff
@@ -29,7 +29,10 @@ import (
 const CookedCommitID = "4e991b55e3d58b9c3137a791a9986ed9c5069697"
 
 func TestGit(t *testing.T) {
-	git, err := git.NewWrapper(test.Username, test.Email)
+	git, err := git.WithConfig(git.Config{
```
katcipis, line 32: I tested creating the cnofig with no username/email and it also works, do we really need them ?

git/git.go
```diff
@@ -445,6 +420,12 @@ func (git *Git) Add(files ...string) error {
 	return err
 }
 
+// Clone will clone the given repo inside the given dir.
+func (git *Git) Clone(repoURL string, dir string) error {
```
i4ki, line 424: by current design, the porcelain commands are guarded by:

```
    if !git.config.AllowPorcelain {
		return fmt.Errorf("Init: %w", ErrDenyPorcelain)
	}
```
I think you should do the same here and then for the case of the vendor command we created the wrapper explicitly allowing porcelain.

i4ki, line 424: by current design, the porcelain commands are guarded by:

```
    if !git.config.AllowPorcelain {
		return fmt.Errorf("Init: %w", ErrDenyPorcelain)
	}
```
I think you should do the same here and then for the case of the vendor command we created the wrapper explicitly allowing porcelain.

git/git_test.go
i4ki: we need them for any write operation in the repo's index objects (git add, git commit, etc).

git/git_test.go
```diff
@@ -140,6 +143,25 @@ func TestRevParse(t *testing.T) {
...
+		filename = "test.txt"
+		content  = "test"
+	)
+	s := sandbox.New(t)
+	s.RootEntry().CreateFile(filename, content)
+	git := s.Git()
+
+	git.CommitAll("add file")
+
+	repoURL := "file://" + s.RootDir()
```
i4ki, line 157: nice

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/mineiros-io/terramate/errors"
+	"github.com/mineiros-io/terramate/git"
+	"github.com/rs/zerolog/log"
+)
+
+// Source represents a module source
+type Source struct {
```
i4ki, line 30: we have this already here: https://github.com/mineiros-io/terramate/blob/main/tf/terraform.go#L31
I think this code should be under a "terraform related package".

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+//
+// Source references that are not Git/Github are not supported.
+func ParseSource(modsource string) (Source, error) {
+	ref := ""
+	splitParams := strings.Split(modsource, "?")
+	modsource = splitParams[0]
+
+	if len(splitParams) > 1 {
+		if len(splitParams) != 2 {
+			return Source{}, errors.E(ErrInvalidModSrc, "unexpected extra '?' on source")
```
i4ki, line 153: the error message is not clear what is "extra" and which ones are supported.
Maybe something like this?

```suggestion
			return Source{}, errors.E(ErrInvalidModSrc, "only `?ref=<val>` query parameter is supported but found %q", strings.Join(splitParams[1:], "?"))
```

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+			return Source{}, err
+		}
+		return Source{
+			URL:  strings.TrimPrefix(modsource, "git::"),
+			Path: path,
+			Ref:  ref,
+		}, nil
+
+	default:
+		return Source{}, errors.E(ErrUnsupportedModSrc)
```
i4ki, line 203:
```suggestion
		return Source{}, errors.E(ErrUnsupportedModSrc, modsource)
```

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+func parseGithubAtPath(modsource string) string {
+	path := strings.TrimPrefix(modsource, "git@")
```
i4ki, line 208: it should be an error if the strings have leading and trailing spaces, no?

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+	path = strings.TrimSuffix(path, ".git")
```
i4ki, line 209: this is not going to trim the `.git` if the URL has the form: `git@github.com:org/owner/repo.git?ref=test-branch`

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+	// This is GH specific, so we don't need to handle specific ports
+	// and we can assume it is always git@github.com:org/path.git.
+	return strings.Replace(path, ":", "/", 1)
```
i4ki, line 212: is this safe? what if user provide a wrong URL like `git@github.com:22/owner/repo.git` ?
maybe we can check here if there's really one instance of `:` in the source and fail if that's not the case?

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func parseGitColonPath(modsource string) (string, error) {
+	rawURL := strings.TrimPrefix(modsource, "git::")
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return "", errors.E(ErrInvalidModSrc, "invalid URL inside git::")
+	}
+	path := filepath.Join(u.Host, u.Path)
+	return strings.TrimSuffix(path, ".git"), nil
```
i4ki, line 222: this also doesn't work if there's a `ref` query parameter.

modvendor/modvendor.go
katcipis: Actually this error indicates that you have more than one ? in the string, isn't it clear enough to say the string has more than one '?' ?

modvendor/modvendor.go
katcipis: The modsource is already stripped of parameter at this point. This works because the tests cover this (we have URLs with refs and the paths are correct).

modvendor/modvendor.go
```diff
@@ -0,0 +1,223 @@
...
+			return Source{}, errors.E(ErrInvalidModSrc, "unknown param %q", splitRefParam[0])
+		}
+
+		ref = splitRefParam[1]
+		if ref == "" {
+			return Source{}, errors.E(ErrInvalidModSrc, "ref param %q is empty", refParam)
+		}
+	}
+
+	switch {
```
i4ki, line 173: `source` most commonly is a relative path but it's not handled here. Is the caller always ensuring this is a git/remote path?

modvendor/modvendor.go
katcipis: Checking for only one instance of : doesn't solve the issue, when we have a port there will be only one :. It would require a more detailed parsing

i4ki:
I know it's still draft but left some comments. Just ignore if it's still incomplete for the cases I mentioned and planned for final version.

modvendor/modvendor.go
i4ki: they are two issues. The first question is not related to the proposed change.

modvendor/modvendor.go
katcipis: It will work (and it is tested currently). In this case I parsed as a URL and path is composed of the Host + Path, that does not include query parameters:

https://go.dev/play/p/rHYvOzuX5Jn

modvendor/modvendor.go
i4ki: I proposed to check for one instance of `:` because the strings.Replace() in this case is intended to replace only the `<host>:<path>` separator, so it should fail if it would replace more stuff.

modvendor/modvendor.go
katcipis: when the source doesn't match any of the git patterns we generated an unsupported source error, which is the same when we have hg:// etc. Why this would be a problem particularly to local files ? (they will be ignored in the end)

git/git_test.go
katcipis: Ah OK, but on this test none of that happens and the test works without it, so maybe here we could remove this ? :thinking:

modvendor/modvendor.go
i4ki: no the error says you have "extra stuff" in the query parameter, you don't say you expect one.

Changing this to `%d extra query parameters` is better but maybe we should be already clear about what we expect there: just the `ref`...

modvendor/modvendor.go
i4ki: fair enough. 
I see that this function can be reused in the `tf.ParseModules()` and then it could have a `Source.Scheme` and a `source.IsLocalPath()` or something.

i4ki: fair enough. 
I see that this function can be reused in the `tf.ParseModules()` and then it could have a `Source.Scheme` and a `source.IsLocalPath()` or something.

git/git_test.go
i4ki: I think so.

tf/mod_source_parse.go
```diff
@@ -0,0 +1,130 @@
...
+			Path: path,
+			Ref:  ref,
+		}, nil
+
+	default:
+		return Source{}, errors.E(ErrUnsupportedModSrc)
+	}
+}
+
+func parseGithubAtPath(modsource string) string {
```
i4ki, line 114: the arg says `modsource` but it should be the `modsource` with stripped query parameters. That's what confused me but I think it's fine.

modvendor/modvendor.go
katcipis: Moved to tf

modvendor/modvendor.go
katcipis: Currently what would happen is that the port number would be part of the path, which is not "unsafe" (it wont break) but also not desireable :laughing:. Will handle removing the port when it is present. Will add the check for how much : we have too...I dont think it is allowed on URLs to have : unless encoded on the path,. but better to ensure on a check indeed.

tf/mod_source_parse.go
katcipis: ah yeah, at some point coming up with the names get tricky if you don't want large names :laughing:. In this case maybe strippedModSource, already a mouthfull but maybe more clear (being specific on what si being stripped would make it REALLY long :laughing: )

modvendor/modvendor.go
katcipis: I answered but it didn't show your comments before, sorry :laughing:. Yeah later I got that indeed there is two issues, thanks for pointing out.

Currently on the code I replace only the first ":", initially I thought that would be enough (then the path would have : on the dir names):

```go
strings.Replace(path, ":", "/", 1)
```

modvendor/modvendor.go
katcipis: Ah...hmm it could be useful to generalize it but I'm not focusing on that right now, sorry :heart:

modvendor/modvendor.go
katcipis: I was trying to be precise that there is an extra `?` but now I get that you are saying that ? makes you think it would be an extra anything ? :laughing:

modvendor/modvendor.go
katcipis: Will try to improve the parsing in general, thanks for pointing this stuff out :heart:

modvendor/modvendor.go
```diff
@@ -0,0 +1,125 @@
...
+		return "", errors.E(err, "creating mod dir inside vendor")
+	}
+
+	logger.Trace().Msg("moving cloned mod from workdir to clonedir")
+	if err := os.Rename(workdir, clonedir); err != nil {
+		// This may leave intermediary created dirs hanging on vendordir
+		// since we just create all and then delete clone dir on a failure to move.
+		// If we get a lot of errors from os.Rename we may need to handle this
+		// more gracefully, here we assume that os.Rename errors are rare since both
+		// dirs where just created.
```
i4ki, line 118:
```suggestion
		// dirs were just created.
```

tf/mod_source_parse.go
```diff
@@ -0,0 +1,123 @@
...
+func ParseSource(modsource string) (Source, error) {
+	switch {
+	case strings.HasPrefix(modsource, "github.com"):
+		u, err := url.Parse(modsource)
+		if err != nil {
+			return Source{}, errors.E(ErrInvalidModSrc, err,
+				"%s is not a URL", modsource)
+		}
+		ref := u.Query().Get("ref")
+		u.RawQuery = ""
```
i4ki, line 61: :smile: 
cool

i4ki (approved):
:rocket: :rocket: :rocket: 
amazing

modvendor/modvendor.go
katcipis: I have some sort of personal issue with were/where :laughing:

i4ki (approved):

PR #514, opened 2022-08-03
fix: warn in case of unexpected terramate block

i4ki:
```
$ cat stack-a/terramate.tm.hcl
stack {
	name = "awesome-stack-a"
}

terramate {

}
$ terramate list
2022-08-04T02:17:37+01:00 WRN  error="/home/i4k/src/mineiros/tm-example/stack-a/terramate.tm.hcl:5,1-10: `terramate` block is only allowed at the project root directory" action=checkConflicts()
stack-a
stack-b
stack-c
```

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -39,7 +39,7 @@ const (
 	ErrTerramateSchema        errors.Kind = "terramate schema error"
 	ErrImport                 errors.Kind = "import error"
 	ErrInvalidDynamicIterator errors.Kind = "invalid dynamic iterator"
-	ErrConfigConflict         errors.Kind = "conflict error"
+	ErrUnexpectedTerramate    errors.Kind = "`terramate` block is only allowed at the project root directory"
 )
 
 const (
@@ -1338,7 +1338,7 @@ func (p *TerramateParser) parseTerramateSchema() (Config, error) {
 
 	logger.Trace().Msg("checking for top-level attributes.")
 
-	rawconfig := p.Imported
+	rawconfig := p.Imported.Copy()
 	err := rawconfig.Merge(p.Config)
 	if err != nil {
 		err = errors.E(err, ErrImport)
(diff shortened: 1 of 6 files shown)


hcl/hcl.go
```diff
@@ -190,6 +191,7 @@ type TerramateParser struct {
 	// parsedFiles stores a map of all parsed files
 	parsedFiles map[string]parsedFile
 
+	strict bool
```
katcipis, line 194: JS feelings :laughing:

hcl/hcl.go
```diff
@@ -1406,6 +1423,48 @@ func (p *TerramateParser) parseTerramateSchema() (Config, error) {
...
+			panic("parsed terramate block not found")
+		}
+
+		errs.Append(
+			errors.E(ErrConfigConflict, stackblock.TypeRange,
+				"`stack` block conflicts with `terramate` block defined at %s",
+				tmblock.RawOrigins[0].TypeRange.String(),
+			),
+
+			errors.E(ErrConfigConflict, tmblock.RawOrigins[0].TypeRange,
```
katcipis, line 1452:
```suggestion
			),
			errors.E(ErrConfigConflict, tmblock.RawOrigins[0].TypeRange,
```

hcl/hcl_stack_test.go
```diff
@@ -330,7 +352,8 @@ func TestHCLParserStack(t *testing.T) {
 			},
 		},
 		{
-			name: "after: empty set works",
+			name:      "after: empty set works",
```
katcipis, line 355:
```suggestion
			name: "after: empty set works",
```

hcl/hcl_stack_test.go
```diff
@@ -160,7 +181,8 @@ func TestHCLParserStack(t *testing.T) {
 			},
 		},
 		{
-			name: "empty name",
+			name:      "empty name",
```
katcipis, line 184:
```suggestion
			name: "empty name",
```

I think the code is not formatted no ? :thinking:

katcipis (dismissed):

i4ki:
@katcipis I changed to report an error if `terramate` is not at the rootdir because my previous assumption was not valid, the docs still allow `stack + terramate` in the same dir **if it's root dir**.
The change broke some tests because `cfg.Terramate` was the easiest one to test the merging at the config level. I removed those testcases because we still have lots of tests for globals + import covering the same merge thing, so I think this is fine.

hcl/hcl.go
```diff
@@ -1406,6 +1423,39 @@ func (p *TerramateParser) parseTerramateSchema() (Config, error) {
 	return config, nil
 }
 
+func (p *TerramateParser) checkConfigSanity(cfg Config) error {
+	logger := log.With().
+		Str("action", "TerramateParser.checkConfigSanity()").
```
katcipis, line 1428: Noice

katcipis (approved):

PR #520, opened 2022-08-04
fix: terramate vendor download fails on os.Rename

katcipis:
# Reason for This Change

When moving the cloned repo to the vendor dir we used os.Rename. But on some systems this fails since it is a cross device operations and rename does not support that on Linux.

## Description of Changes

Instead of using os.Rename from a tmpdir outside the project we do it from a tmpdir inside the project, making it more likely for the fs/device to be the same. We still clone the module outside the project since some git setups will automatically add any cloned repos as submodules.

The overall flow is:

- Clone repo on /tmp
- Remove .git while it is on /tmp
- Copy to <projectroot>/.tmvendor<tmp-unique>
- Rename from <projectroot>/.tmvendor<tmp-unique> to vendor/<etc> (which then could still fail if the vendor dir is on a different mount/dev/etc).

The idea is to reduce as much as possible the chances of rename failing, but still use it since it will change vendor atomically (avoiding to leave vendor with half copied things).


fs/copy.go
```diff
@@ -0,0 +1,85 @@
...
+// If it returns false, the file/dir is ignored.
+// In the case of directories this indicates that the whole dir
+// and its files and subdirs will not be copied.
+type CopyFilterFunc func(os.DirEntry) bool
+
+// CopyTree will copy the whole tree from srcdir to destdir.
+// This means copying all dirs and files recursively.
+// The destdir provided does not need to exist, will be created.
+// The provided filter function allows to filter which files/directories
+// get copied.
```
katcipis, line 36:
```suggestion
// The provided filter function allows to filter which files/directories get copied.
```

mariux (dismissed):
we should still copy it to a temporary directory and rename in the end to make the existence atomic. if something fails in the current approach, we can end with half filled vendor directories.

katcipis:
> we should still copy it to a temporary directory and rename in the end to make the existence atomic. if something fails in the current approach, we can end with half filled vendor directories.

The current approach is cloning in a temporary directory and then renaming, but rename won't work accross different device :thinking: 

The only thing that can leave half copied files would be the process exiting suddently, even in the case of a Go panic we would still delete any leftover files (on tmp + on vendor, using defer). Or if when we try to cleanup the cleanup itself fails.

soerenmartius (approved):
lgtm

PR #521, opened 2022-08-04
fix: vendor download should use env vars on git clone

katcipis:
# Reason for This Change

Currently when running git clone on modvendor we are not loading any of the environment variables since by default our git instances force an empty env. 

## Description of Changes

For clone, on modvendor, we add the parent process env.


mariux (approved):
lgtm

PR #523, opened 2022-08-05
test: generate code for variable with optionals (tf 1.3)

katcipis:
# Reason for This Change

Just double checking that the new variable with optional descriptions work on Terramate. It parses it and partial evaluation also works on it.

## Description of Changes

Just added the test. Here is also a quick script to see it working live:

```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

terramate {
  config {
  }
}
EOM

cat > variable.tm.hcl <<- EOM
globals {
  default = 666
}

generate_hcl "variable_optional.tf" {
  content {
    variable "with_optional_attribute" {
      type = object({
        a = string
        b = optional(string)
        c = optional(number, global.default)
      })
    }
  }
}
EOM

terramate create stack
terramate generate

echo
echo "generated code"
cat stack/variable_optional.tf
```

### DIFF
--- a/generate/genhcl/genhcl_hclwrite_test.go
+++ b/generate/genhcl/genhcl_hclwrite_test.go
@@ -30,6 +30,10 @@ func block(name string, builders ...hclwrite.BlockBuilder) *hclwrite.Block {
 	return hclwrite.BuildBlock(name, builders...)
 }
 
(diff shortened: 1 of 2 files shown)


mariux (approved):
lgtm

PR #524, opened 2022-08-05
test: parsing variables with optionals (tf 1.13)

katcipis:
# Reason for This Change

Added test to check that new Terraform 1.13 feature, variables with object definitions with optionals, doesn't break parsing and detecting changes across modules.

## Description of Changes

Added a test doing change detection on a module when the code also has variables with the new optionals being used.


mariux (approved):
lgtm

PR #526, opened 2022-08-08
refactor: split genhcl tests

katcipis:
# Reason for This Change

Currently we have a quite big set of tests for genhcl on a single file. The idea is to split so it makes it easier to find specific tests, specially as we add more features on generate_hcl.

## Description of Changes

The tests were split on:

* General tests
* Partial Eval specific tests
* Dynamic specific tests

No tests were added/removed, they were just extracted into different files. So changes on behavior or on what is tested.


mariux (approved):
lgtm

PR #527, opened 2022-08-09
feat: add tm_dynamic attributes

katcipis:
# Reason for This Change

Adding support to `attributes` inside `tm_dynamic` as a way to generate blocks using an object/map as the source of the blocks attributes. This is useful in some scenarios where `content` is not enough, like dynamic generation of `terraform.required_providers` (or a scenario where each generated block will have different attributes).

## Description of Changes

Add support to `attributes` and its tests. Also added some tests for the overall `tm_dynamic` that were missing, specially error handling and fixed some issues like `tm_dynamic` allowing unknown attributes instead of failing.

## Test example

Here is a simple test script that serves as an example on how this could be used to generate providers configuration:

```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

terramate {
  config {
  }
}
EOM

cat > providers.tm.hcl <<- EOM
globals {
  used_providers = ["google", "kubernetes"]
  available_providers = {
    google = {
      source  = "hashicorp/google"
      version = "version 1"
    }
    google-beta = {
      source  = "hashicorp/google-beta"
      version = "version 2"
    }
    kubernetes = {
      source  = "hashicorp/kubernetes"
      version = "version 3"
    }
  }
  required_providers = { for k,v in global.available_providers : k => v if tm_contains(global.used_providers, k)}
}

generate_hcl "providers.tf" {
  content {
    terraform {
      tm_dynamic "required_providers" {
        for_each = tm_length(global.required_providers) > 0 ? [true] : []
        attributes = global.required_providers
      }
    }
  }
}
EOM

terramate create stack
terramate generate

echo
echo "generated providers:"
cat stack/providers.tf
```


hcl/hcl.go
```diff
@@ -813,7 +823,7 @@ func validateGenerateFileBlock(block *ast.Block) error {
 // as is (original expression form, no evaluation).
 //
 // Returns an error if the evaluation fails.
-func CopyBody(target *hclwrite.Body, src *hclsyntax.Body, eval Evaluator) error {
+func CopyBody(dest *hclwrite.Body, src *hclsyntax.Body, eval Evaluator) error {
```
katcipis, line 826: While I implemented this I realized that maybe CopyBody should be on genhcl. The logic here is by now pretty genhcl specific, it is not a generic/plain HCL body copying anymore, thanks to all the tm_dynamic logic. But I'm not sure so I didn't moved anything yet, wondering :thinking:

hcl/hcl.go
```diff
@@ -961,15 +1046,99 @@ func appendDynamicBlock(target *hclwrite.Body, block *hclsyntax.Block, eval Eval
...
+			}
+
+			attrName := keyVal.AsString()
+			valExpr, err := eval.GetExpressionTokens(partialEvalAttributes.Bytes(), item.ValueExpr)
+			if err != nil {
+				// Panic here since Terramate generated an invalid expression after
+				// partial evaluation and it is a guaranteed invariant that partial
+				// evaluation only produces valid expressions.
+				log.Error().
+					Err(diags).
```
i4ki, line 1122:
```suggestion
					Err(err).
```

generate/genhcl/genhcl_dynamic_test.go
```diff
@@ -315,83 +941,92 @@ func TestGenerateHCLDynamic(t *testing.T) {
...
+			stack: "/stack",
+			configs: []hclconfig{
+				{
+					path: "/stack",
+					add: generateHCL(
+						labels("tm_dynamic_test.tf"),
+						content(
+							tmdynamic(
+								labels("my_block"),
+								expr("for_each", `[local.a]`),
```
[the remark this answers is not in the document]
katcipis: Hmm interesting, seems possible. Using it as attributes could be tricky since we need proper strings/names as the attribute names, not partial expressions :thinking:

hcl/hcl.go
katcipis: Nice catch, thanks :heart:

i4ki (approved):
lgtm

PR #528, opened 2022-08-10
feat: add global unset

katcipis:
# Reason for This Change

Add support to set a global to `unset`, which unsets the global. References to it will fail as if the global was never defined.

## Description of Changes

Changed globals evaluation, removing any globals set to `unset` before starting the evaluation process.

## Example

```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

terramate {
  config {
  }
}
EOM

cat > globals.tm.hcl <<- EOM
globals {
    a = 1
    b = 2
}
EOM

terramate create stacks/stack-1
terramate create stacks/stack-2

cat > stacks/stack-1/globals.tm.hcl <<- EOM
globals {
    a = unset
}
EOM

cat > stacks/stack-2/globals.tm.hcl <<- EOM
globals {
    b = unset
}
EOM

terramate experimental globals
```

Will result in the globals being:

```
stack "/stacks/stack-1":
        b = 2

stack "/stacks/stack-2":
        a = 1
```


stack/globals.go
```diff
@@ -106,6 +106,23 @@ func (ge *globalsExpr) has(name string) bool {
 	return ok
 }
 
+func removeUnset(globals map[string]expression) {
```
katcipis, line 109: Did this instead of modeling unset as an actual value, like null. Since unset is very specific to undefining globals it seemed simpler to keep it constrained on its usage, instead of allowing it to be used as an value, like inside lists/objs/ternaries etc. But maybe that would be desireable, but then we need to think on what to do when we have unset inside an object and that object is rendered on code generation, etc.

i4ki (approved):
lgtm

PR #534, opened 2022-08-16
fix: check if tm_dynamic.attributes keys are valid identifiers

katcipis:
# Reason for This Change

hclwrite will accept any string as an attribute name with no validation, so using an obj/map to get attribute name will lead to situations where the string is not a valid identifier. When that happens today the error is unclear.

## Description of Changes

Check the identifier explicitly. Also if we generate invalid code that can't be formatted we panic since this is a severe/internal Terramate error.


generate/genhcl/genhcl.go
```diff
@@ -174,9 +174,18 @@ func Load(rootdir string, sm stack.Metadata, globals stack.Globals) ([]HCL, erro
 		}
 		formatted, err := hcl.FormatMultiline(string(gen.Bytes()), loadedHCL.origin)
 		if err != nil {
-			return nil, errors.E(sm, err,
-				"failed to format generated code for block %q", name,
-			)
+			// genhcl must always generate valid code that is formatable
+			// this is a severe internal error
+			logger.Error().
```
katcipis, line 179: Coverage dropped because now we log more details + panic instead of a single return, but the condition was never tested anyway because it requires simulating a situation where terramate generates invalid HCL code, and that should never happen =P

i4ki: so maybe we can remove this `if` altogether and let it crash if it ever occurs?

hcl/hcl.go
```diff
@@ -1108,7 +1108,17 @@ func appendDynamicBlock(target *hclwrite.Body, block *hclsyntax.Block, evaluator
 				return true
 			}
 
+			// hclwrite lib will accept any arbitrary string as attr name
+			// allowing it to generate invalid HCL code, so here we check if
+			// the attribute is a proper HCL identifier.
 			attrName := keyVal.AsString()
+			if !hclsyntax.ValidIdentifier(attrName) {
```
i4ki, line 1115: nice!!

i4ki (approved):

generate/genhcl/genhcl.go
katcipis: hmm the problem here is that if we remove the if it won't crash, it will just save non-formatted/broken things or an empty file (when formatting fails it returns an empty string as result, so mostly it will be an empty file).

PR #537, opened 2022-08-17
fix: tm_dynamic fails with  duplicated labels

katcipis:
# Reason for This Change

Currently using tm_dynamic with duplicated labels on the labels attribute will fail but it shouldn't.

## Description of Changes

Refactored the CopyBody from hcl to genhcl since it contains a lot of logic that is genhcl specific. We still have some parsing that could be on hcl that is now done on genhcl, so it would be better to improve this further, but at least for now it simplifies the already big hcl package.

Instead of using the assignSet function that handle a cty.Value we have a function that handles a cty.Value as a list of strings.

## Reproducing Issue

```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

terramate {
  config {
  }
}
EOM

cat > test.tm.hcl <<- EOM
generate_hcl "test.tf" {
  content {
    tm_dynamic "resource" {
      for_each = ["a"]
      labels = ["a", "a"]
      attributes = {}
    }
  }
}
EOM

terramate create stack
terramate generate

echo
echo "generated providers:"
cat stack/test.tf
```

### DIFF
--- a/generate/genhcl/genhcl.go
+++ b/generate/genhcl/genhcl.go
@@ -430,7 +430,7 @@ func appendDynamicBlock(target *hclwrite.Body, block *hclsyntax.Block, evaluator
 				return true
 			}
 
-			labels, err = hcl.ParseStringList("labels", labelsVal)
+			labels, err = hcl.ValueAsStringList(labelsVal)
 			if err != nil {
 				tmDynamicErr = wrapHCLAttrErr(err, attrs.labels,
 					"parsing tm_dynamic.labels")
--- a/generate/genhcl/genhcl.go
+++ b/generate/genhcl/genhcl.go
@@ -67,7 +67,10 @@ const (
 
 	// ErrInvalidDynamicIterator indicates that the iterator of a tm_dynamic block
 	// is invalid.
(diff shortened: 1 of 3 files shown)


generate/genhcl/genhcl.go
```diff
@@ -265,3 +285,342 @@ func loadGenHCLBlocks(rootdir string, cfgdir string) ([]loadedHCL, error) {
...
+		if attrs.labels != nil {
+			labelsVal, err := evaluator.Eval(attrs.labels.Expr)
+			if err != nil {
+				tmDynamicErr = errors.E(ErrInvalidDynamicLabels,
+					err, attrs.labels.Range(),
+					"failed to evaluate tm_dynamic.labels")
+				return true
+			}
+
+			labels, err = hcl.ValueAsStringList(labelsVal)
```
i4ki, line 442: thanks!!!

i4ki: all the rest is just refactor? or is there more fixing?

i4ki (approved):
lgtm

PR #541, opened 2022-08-19
feat: hcl: support parsing new vendor config

katcipis:
# Reason for This Change

To implement support to manifest configuration on vendoring we need to make the hcl parser aware of the vendor config.

## Description of Changes

Add support to the vendor configuration and its tests.


test/hcl.go
```diff
@@ -121,10 +122,6 @@ func assertTerramateBlock(t *testing.T, got, want *hcl.Terramate) {
 func assertTerramateConfigBlock(t *testing.T, got, want *hcl.RootConfig) {
 	t.Helper()
 
-	if (want == nil) != (got == nil) {
```
katcipis, line 124: This was already checked before calling this function.

docs/config-overview.md
```diff
@@ -158,7 +159,8 @@ More details can be found [here](project-config.md#the-terramateconfigrunenv-blo
 
 # stack block schema
 
-The `stack` block has no labels, supports [merging](#config-merging) and has the following schema:
+The `stack` block has no labels, **does not** support [merging](#config-merging)
```
i4ki, line 162: o.o

docs/config-overview.md
```diff
@@ -213,3 +215,31 @@ and has the following schema:
...
+| [default](#vendormanifestdefault--block-schema) | block | The default manifest |
+
+# vendor.manifest.default block schema
+
+The `vendor.manifest.default` block has no labels, **do not** support [merging](#config-merging)
+and has the following schema:
+
+| name             |      type      | description |
+|------------------|----------------|-------------|
+| files            | list(string)   | The list of [gitignore patterns](https://git-scm.com/docs/gitignore) to match selected files |
```
i4ki, line 245:
```suggestion
| files            | list(string)   | The list of patterns to match selected files. The pattern format is the same of [gitignore](https://git-scm.com/docs/gitignore#_pattern_format) |
```

I'm not sure if this improves but the other way it looks like the option is related to git, when it's not, we only use gitignore pattern format.. feel free to ignore.

hcl/hcl.go
```diff
@@ -1002,6 +1023,115 @@ func parseStack(evalctx *eval.Context, stack *Stack, stackblock *ast.Block) erro
...
+	for _, got := range block.Blocks {
+		for _, want := range blocks {
+			if want == got.Type {
+				if found {
+					errs.Append(errors.E(ErrTerramateSchema,
+						got.DefRange(),
+						"duplicated block %s",
+						got.Type),
+					)
+					continue checkBlocks
```
i4ki, line 1054:
```suggestion
```
This line seems unneeded.

hcl/hcl.go
```diff
@@ -1002,6 +1023,115 @@ func parseStack(evalctx *eval.Context, stack *Stack, stackblock *ast.Block) erro
 	return errs.AsError()
 }
 
+func checkHasSubBlocks(block *ast.Block, blocks ...string) error {
```
i4ki, line 1026: The function is called `checkHasSubBlocks()` but it also fails if the block has attributes and fails if it has labels. Maybe you can make it only checking for sub-blocks and create another function that does the complete check for this kind of block?

hcl/hcl.go
```diff
@@ -1002,6 +1023,115 @@ func parseStack(evalctx *eval.Context, stack *Stack, stackblock *ast.Block) erro
...
+
+	if len(manifestBlock.Blocks) == 0 {
+		return nil
+	}
+
+	defaultBlock := manifestBlock.Blocks[0]
+
+	errs := errors.L()
+
+	for _, block := range defaultBlock.Blocks {
```
i4ki, line 1103: can't you call `checkHasSubBlocks()` with an empty set of wanted blocks?

hcl/hcl.go
```diff
@@ -1231,16 +1361,18 @@ func (p *TerramateParser) parseTerramateSchema() (Config, error) {
...
 	for _, block := range rawconfig.UnmergedBlocks {
 		// unmerged blocks
 
 		logger := logger.With().
 			Str("block", block.Type).
 			Logger()
 
-		if block.Type == StackBlockType {
+		switch block.Type {
+		case StackBlockType:
```
i4ki, line 1375: I'm wondering now why we have this just for "stack" haha I think I added this but don't remember haha

i4ki (approved):
lgtm but left some small improvements (maybe not).
Feel free to ignore.

hcl/hcl.go
katcipis: hmm yeah it does way too much things then naming it was complicated, will improve.

hcl/hcl.go
katcipis: I wanted to avoid setting found again to true. it doesn't change the semantics (it is already true), but it seemed more well behaved, maybe in the future something else is done after this check that doesn't make sense when it is a duplicate.

hcl/hcl.go
katcipis: the function does way too much now, it would also fail because of attributes. I will split them and then it could be used. I started with a schema validation function but it started to be way too much for this case, so I kept changing it to remove some logic and in the end it is an aberration that is not full schema check but also way too complicated :laughing: . Splitting in smaller funcs should make things much more clear.

hcl/hcl.go
katcipis: We use this exported constant on other parts of the code to avoid duplicating the knowledge of what is the stack block name. We don't have this problem with other blocks I think.

PR #544, opened 2022-08-22
feat: metadata: terramate.stacks.list

katcipis:
# Reason for This Change

For some scenarios we need a full list of all stacks. Here we add this list as the metadata `terramate.stacks.list`.

## Description of Changes

Adds the metadata including a redesign since the current design was done with stacks having metadata in mind, but now we have project metadata that does not change accross different stacks.


test/hclwrite/hclwrite.go
```diff
@@ -175,7 +178,13 @@ func AttributeValue(t *testing.T, name string, expr string) BlockBuilder {
 	}
 	body := res.Body.(*hclsyntax.Body)
 
-	val, diags := body.Attributes[name].Expr.Value(nil)
+	scope := &tflang.Scope{}
+	ctx := &hhcl.EvalContext{
+		Functions: scope.Functions(),
```
katcipis, line 183: so we can force an attribute to be a list using "tolist"..or else there is no way to create actual lists only tuples =P

stack/stack.go
```diff
@@ -227,6 +227,18 @@ func LookupParent(root, dir string) (*S, bool, error) {
 	return l.lookupParentStack(dir)
 }
 
+// NewProjectMetadata creates project metadata from a given rootdir and a list of stacks.
+func NewProjectMetadata(rootdir string, stacks List) project.Metadata {
```
katcipis, line 231: This function is to specifically crete the project metadata from a stack.List type, hence why it is on the stack package instead of the project pkg.

stack/globals.go
```diff
@@ -210,20 +210,19 @@ func (ge *globalsExpr) eval(rootdir string, meta Metadata) (Globals, error) {
...
 			logger.Err(err).
 				Str("name", name).
 				Str("origin", expr.origin).
 				Msg("evaluating global")
 		}
-		return Globals{}, errors.E(
-			ErrGlobalEval,
-			"unable to evaluate %d globals", len(pendingExprs),
-		)
+		return Globals{}, errors.E(ErrGlobalEval, "%v", pendingGlobalNames)
```
katcipis, line 225: was getting some eval errors and this made them easier to understand, not perfect but better than just failed to eval 1 var

docs/sharing-data.md
```diff
@@ -258,15 +258,31 @@ accessed through the variable namespace **terramate**.
 This can be referenced from any Terramate code to reference
 information like the path of the stack or its name.
 
-## terramate.root.path.fs.absolute (string)
+
+## Project Metadata
+
+Project metadata is the same independent of stack.
```
i4ki, line 264: This reads funny, did you move this section? maybe it would make more sense if defined after the stack metadata.

generate/generate.go
```diff
@@ -502,15 +503,15 @@ func forEachStack(root, workingDir string, fn forEachStackFunc) Report {
 
 	logger.Trace().Msg("List stacks.")
 
-	stackEntries, err := terramate.ListStacks(root)
+	stacks, err := stack.LoadAll(root)
```
i4ki, line 506: for each found stack we walk/parse/load the full list of stacks again?

generate/genfile/genfile_test.go
```diff
@@ -142,6 +143,7 @@ func TestLoadGenerateFiles(t *testing.T) {
 						labels("test"),
 						expr("content", `<<EOT
 
+stacks_list=${tm_jsonencode(terramate.stacks.list)}
```
i4ki, line 146: :smile:

i4ki (approved):
lgtm but left some questions that can be addressed later.

docs/sharing-data.md
katcipis: This section didn't exist and I added it before the stack because on my mind it would be project wide -> stack specific, but I could change the order, no strong opinion :+1:

generate/generate.go
katcipis: no, it will be loaded once and then it is passed on to the per stack handler. The idea is that forEachStack will call the given function for each stack existing on the project, so this function is called once, we load the stacks, and then call the function for each stack passing the metadata etc.

I could transform this into a loop on generate.Do...it was designed this way because the logic was duplicated on the past, but design changes made it useless, forEachStack is used only in a single place (generate.Do), a simple for would be simpler x_x. Will do in another PR.

generate/generate.go
```diff
@@ -502,15 +503,15 @@ func forEachStack(root, workingDir string, fn forEachStackFunc) Report {
...
 	if err != nil {
 		report.BootstrapErr = err
 		return report
 	}
 
-	for _, entry := range stackEntries {
-		st := entry.Stack
+	projmeta := stack.NewProjectMetadata(root, stacks)
 
+	for _, st := range stacks {
```
katcipis, line 514: @i4ki here is what happens per stack

PR #553, opened 2022-08-24
feat: recursive vendoring

i4ki:
Depends on #555 and #556


modvendor/modvendor.go
```diff
@@ -148,3 +303,81 @@ func Vendor(rootdir string, modsrc tf.Source) (string, error) {
...
+				continue
+			}
+			source := block.Body().GetAttribute("source")
+			if source == nil {
+				continue
+			}
+
+			sourceString := string(source.Expr().BuildTokens(nil).Bytes())
+			sourceString = strings.TrimSpace(sourceString)
+			sourceString = sourceString[1 : len(sourceString)-1] // unquote
```
katcipis, line 348: if tf code has a source = "" won't that parse OK but then panic here ?

modvendor/modvendor_report.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Report with the result of the vendor related functions.
+type Report struct {
+	Vendored map[string]Vendored
+	Ignored  []IgnoredVendor
+	Error    error
+}
+
+// NewEmptyReport returns a new empty report.
+func NewEmptyReport() Report {
```
katcipis, line 48: Maybe just NewReport ? it is kinda usual that New functions create an empty new instance.

modvendor/modvendor.go
i4ki: `source.Expr().BuildTokens(nil).Bytes()` will return `""` and then this line will return an empty string.
But I will add a test for this. I didn't have time to improve that yet. I think I'm gonna do it in a separate PR, I think we can avoid messing with tokens here.

modvendor/modvendor_report.go
i4ki: the original idea was using `NewReport(vendored, ignored, err)` but in the end it wasn't needed, so yeah, I will apply your suggestion. Thanks.

i4ki:
@katcipis I will update the table tests to also support alternate vendorDir now. So I will put it in draft again for a short time.

modvendor/modvendor.go
katcipis: Hmm Im probably missing something stupid, if the sourceString is "" wont the slice be `sourceString[1:-1]` which will panic ?

Like this : https://go.dev/play/p/fl6CqsY2qTq

Again, probably missing something here :laughing:

katcipis: Ah now I got it, you mean the buildtokens never returns an actual empty string, it has always at least the double quotes. Makes sense.

i4ki: yep :smile:

i4ki: I meant it returns `OQUOTE` `STR` `CQUOTE`.

modvendor/modvendor.go
```diff
@@ -15,59 +15,220 @@
...
+	return vendorAll(rootdir, vendorDir, tfdir, NewEmptyReport())
+}
+
+func recVendor(rootdir string, vendorDir string, modsrc tf.Source, report Report, info *modinfo) Report {
+	logger := log.With().
+		Str("action", "modvendor.recVendor()").
+		Str("module.source", modsrc.Raw).
+		Logger()
+
+	moddir, err := doVendor(rootdir, vendorDir, modsrc)
```
katcipis, line 89: hmm `recVendor` + `doVendor` seems confusing. Usually when I have a recursive implementation the name of the function is the same but private, like `vendor`, so you have Vendor calling an internal recursive vendor() function. And then we would need a better name to doVendor, not sure yet, maybe `downloadVendor`.

katcipis (commented):
Just comment, still reviewing

modvendor/modvendor.go
```diff
@@ -15,59 +15,220 @@
...
+	for fname := range originMap {
+		files = append(files, fname)
+	}
+	sort.Strings(files)
+	errs.Append(patchFiles(rootdir, files, sourcemap))
+	report.Error = errs.AsError()
+	return report
+}
+
+// doVendor will doVendor the provided modsrc into the rootdir.
```
katcipis, line 206:
```suggestion
// downloadVendor will download the provided modsrc into the vendorDir.
```

Just testing the new name, seems better =P

modvendor/modvendor.go
```diff
@@ -155,3 +316,81 @@ func Vendor(rootdir, vendordir string, modsrc tf.Source) (string, error) {
...
+					Msg("found relevant module")
+
+				relPath, err := filepath.Rel(
+					filepath.Dir(fname), filepath.Join(rootdir, info.vendoredAt),
+				)
+				if err != nil {
+					errs.Append(err)
+				} else {
+					newexpr := hclwrite.NewExpressionLiteral(cty.StringVal(relPath))
+					block.Body().SetAttributeRaw("source", newexpr.BuildTokens(nil))
```
katcipis, line 376:
```suggestion
					block.Body().SetAttributeValue("source", cty.StringVal(relPath))
```

modvendor/modvendor_report.go
```diff
@@ -0,0 +1,120 @@
...
+	sources := []string{}
+	for source := range r.Vendored {
+		sources = append(sources, source)
+	}
+
+	sort.Strings(sources)
+	for _, source := range sources {
+		vendored := r.Vendored[source]
+		addLine("[+] %s", vendored.Source.URL)
+		addLine("    ref: %s", vendored.Source.Ref)
```
katcipis, line 73: hmm on the generate report I used tabs, would it be better to use spaces ? (oh no tabs vs spaces discussion :laughing: )

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+
+type testcase struct {
+	name         string
+	source       string
+	layout       []string
+	vendordir    string
+	configs      []hclconfig
+	wantVendored []string
+	wantIgnored  []wantIgnoredVendor
+	wantError    error
```
katcipis, line 60: hmm since Vendor returns a modvendor.Report, why we just dont define a want modvendor.Report instead of 3 different want fields ? :thinking:

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+
+	// wantFiles describes file contents that must exist in the vendor directory.
+	// The fmt.Stringer can be a template that gets compiled with text/template
+	// package and having access to the relative paths of the vendored modules.
+	// For example, if `wantVendored` contains [<modsrc1>, <modsrc2>, ...], then
+	// the template will have access to the list [<relative path of modsrc1>,
+	// <relative path of modsrc2>, ...].
+	// Then the template can reference the third element of the list:
+	//   source = {{index 2}}
+	// If this looks like a hack to you, it's because it is.
```
katcipis, line 70: :laughing:

test/sandbox/sandbox.go
```diff
@@ -133,6 +133,7 @@ func NoGit(t *testing.T) S {
 //   <kind>:<relative path>[:data]
 // Where kind is one of the below:
 //   "d" for directory creation.
+//   "g" for local git directory creation.
```
katcipis, line 136: noice :100:

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+	ReasonPattern string
+}
+
+type wantReport struct {
+	Vendored []string
+	Ignored  []wantIgnoredVendor
+	Error    error
+}
+
+func TestModVendorAllRecursive(t *testing.T) {
```
katcipis, line 91:
```suggestion
func TestModVendor(t *testing.T) {
```

I think we only export on the package the Vendor function that is always recursive no ?

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+				{
+					RawSource:     "https://example.com/my-module",
+					ReasonPattern: "unsupported module source",
+				},
+			},
+		},
+		{
+			name:   "module not found",
+			source: "git::file://{{.}}/module-that-does-not-exists?ref=main",
+			wantIgnored: []wantIgnoredVendor{
```
katcipis, line 160: hmm if the reference is git but doesn't exist it is also just ignored ?

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+					data: Module(
+						Labels("test"),
+						Str("source", "git::file://{{.}}/another-module?ref=main"),
+					),
+				},
+			},
+			wantFiles: map[vendorPathSpec]fmt.Stringer{
+				"git::file://{{.}}/module-test?ref=main#main.tf": Module(
+					Labels("test"),
+					Str("source", "{{index . 1}}"),
```
katcipis, line 187: hmm I usually like table tests, but in this case thinking if maybe we are stretching too much to do it, maybe N test functions (even with some duplication) would be easier to understand :thinking: (or a more sophisticated DSL, but then it may be way too much effort for this... IDK).

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+							Str("source", "git::file://{{.}}/another-module?ref=main"),
+						),
+					),
+				},
+			},
+			wantFiles: map[vendorPathSpec]fmt.Stringer{
+				"git::file://{{.}}/module-test?ref=main#main.tf": Doc(
+					Module(
+						Labels("test"),
+						Str("source", "{{index . 1}}"),
```
katcipis, line 272: Hmm if there was a way for these references to be by name, like the name of the module you depend and then it expands do the source, it would improve things considerably, the positional access is quite hard to understand for a newcomer (or someone as limited as me :laughing: ...I think I get it now, and the problem is tricky, we can assess this later).

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+					),
+				},
+			},
+			wantVendored: []string{
+				"git::file://{{.}}/my-module?ref=main",
+				"git::file://{{.}}/awesome-module?ref=main",
+				"git::file://{{.}}/cool-module?ref=main",
+				"git::file://{{.}}/best-module?ref=main",
+			},
+			wantFiles: map[vendorPathSpec]fmt.Stringer{
```
katcipis, line 638: Shouldn't "cool-module" be on the wantFiles list with a rewritten source to my-module ? In case of a cycle I was expecting the module to be vendored + have the source patched too, no ? (sorry if missing something basic here)

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+				test.WriteFile(t, s.RootDir(), cfg.path,
+					fixupString(t, cfg.data.String(), modulesDir))
+
+				git := sandbox.NewGit(t, filepath.Join(modulesDir, cfg.repo))
+				git.CommitAll("files updated")
+			}
+			source := fixupString(t, tc.source, modulesDir)
+			modsrc, err := tf.ParseSource(source)
+			assert.NoError(t, err)
+			got := modvendor.Vendor(rootdir, tc.vendordir, modsrc)
```
katcipis, line 685:
```suggestion
			for _, cfg := range tc.configs {
				test.WriteFile(t, s.RootDir(), cfg.path,
					fixupString(t, cfg.data.String(), modulesDir))

				git := sandbox.NewGit(t, filepath.Join(modulesDir, cfg.repo))
				git.CommitAll("files updated")
			}
			source := fixupString(t, tc.source, modulesDir)
			modsrc, err := tf.ParseSource(source)
			assert.NoError(t, err)
			
			rootdir := t.TempDir()
			got := modvendor.Vendor(rootdir, tc.vendordir, modsrc)
```

Move rootdir closer to where it is used, I was a little confused at first thinking it would be used on the for loop.

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+
+			if tc.vendordir == "" {
+				tc.vendordir = "/vendor"
+			}
+
+			modulesDir := s.RootDir()
+			rootdir := t.TempDir()
+			for _, cfg := range tc.configs {
+				test.WriteFile(t, s.RootDir(), cfg.path,
+					fixupString(t, cfg.data.String(), modulesDir))
```
katcipis, line 677: maybe applyConfigTemplate ? fixupString is quite vague

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+
+				git := sandbox.NewGit(t, filepath.Join(modulesDir, cfg.repo))
+				git.CommitAll("files updated")
+			}
+			source := fixupString(t, tc.source, modulesDir)
+			modsrc, err := tf.ParseSource(source)
+			assert.NoError(t, err)
+			got := modvendor.Vendor(rootdir, tc.vendordir, modsrc)
+			want := fixupReport(t, wantReport{
```
katcipis, line 686: Same here, maybe applyReportTemplate

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+	srctpl, err := template.New("template").Parse(input)
+	assert.NoError(t, err)
+	var buf bytes.Buffer
+	err = srctpl.Execute(&buf, value)
+	assert.NoError(t, err)
+	return buf.String()
+}
+
+func fixupReport(t *testing.T, r wantReport, value string, vendordir string) modvendor.Report {
+	out := modvendor.Report{
```
katcipis, line 708:
```suggestion
func fixupReport(t *testing.T, r wantReport, value string, vendordir string) modvendor.Report {
	t.Helper()
	
	out := modvendor.Report{
```

modvendor/modvendor_test.go
i4ki: In the beginning I was testing like this (several Test* funcs) but there's a lot of duplication even if you abstract in several functions.. for me it was worse, but I agree that the hacks here are not cool as well. But I have some ideas to improve this if we have time in the future.

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+}
+
+func evaluateWantedFiles(
+	t *testing.T,
+	wantFiles map[vendorPathSpec]fmt.Stringer,
+	modulesDir string,
+	rootdir string,
+	vendordir string,
+) map[string]fmt.Stringer {
+	evaluated := map[string]fmt.Stringer{}
```
katcipis, line 739:
```suggestion
) map[string]fmt.Stringer {
	t.Helper()
	
	evaluated := map[string]fmt.Stringer{}
```

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+	for pathSpec, expectedStringer := range wantFiles {
+		pathSpecParts := strings.Split(string(pathSpec), "#")
+		assert.EqualInts(t, len(pathSpecParts), 2)
+		source := pathSpecParts[0]
+		path := pathSpecParts[1]
+
+		modsrc, err := tf.ParseSource(fixupString(t, source, modulesDir))
+		assert.NoError(t, err)
+		absVendorDir := modvendor.AbsVendorDir(rootdir, vendordir, modsrc)
```
katcipis, line 748: noice

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+}
+
+func checkWantedFiles(
+	t *testing.T,
+	tc testcase,
+	modulesDir string,
+	rootdir string,
+	vendordir string,
+) {
+	wantFiles := evaluateWantedFiles(t, tc.wantFiles, modulesDir, rootdir, vendordir)
```
katcipis, line 762:
```suggestion
) {
	t.Helper()
	
	wantFiles := evaluateWantedFiles(t, tc.wantFiles, modulesDir, rootdir, vendordir)
```

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,823 @@
...
+			return nil
+		}
+
+		expectedStringTemplate, ok := wantFiles[path]
+		if ok {
+			// file must be rewritten
+			relVendoredPaths := computeRelativePaths(
+				t, filepath.Dir(path), tc.wantVendored, modulesDir, rootdir, tc.vendordir,
+			)
+			want := fixupString(t, expectedStringTemplate.String(), relVendoredPaths)
```
katcipis, line 787: Ah here would be the index access on the templates right ?

modvendor/modvendor_test.go
katcipis: Hmm true, in the end a really good DSL wins in this case.

modvendor/modvendor_test.go
katcipis: Ahh...would be already an improvement. We can address in the future :+1:

modvendor/modvendor_test.go
```diff
@@ -15,18 +15,843 @@
...
+			moduleName := pathParts[1]
+
+			originalPath = filepath.Join(originalPath, moduleName, strings.Join(pathParts[3:], "/"))
+
+			originalBytes, err := ioutil.ReadFile(originalPath)
+			assert.NoError(t, err)
+
+			gotBytes, err := ioutil.ReadFile(path)
+			assert.NoError(t, err)
+			assert.EqualStrings(t, string(originalBytes), string(gotBytes),
```
katcipis, line 825: hmm maybe using test.AssertDiff would produce a better output for comparing multiline strings

katcipis (commented):
Just a comment, GH interface is annoying

modvendor/modvendor_test.go
```diff
@@ -49,19 +874,23 @@ func TestModVendorWithCommitIDRef(t *testing.T) {
 	gitURL := "file://" + repoSandbox.RootDir()
 	rootdir := t.TempDir()
 
-	const vendordir = "/dir/reftest/vendor"
-	cloneDir, err := modvendor.Vendor(rootdir, vendordir, tf.Source{
-		URL:  gitURL,
-		Ref:  ref,
-		Path: path,
-	})
+	source, err := tf.ParseSource(fmt.Sprintf("git::%s?ref=%s", gitURL, ref))
```
katcipis, line 877: hmm any reason why we need to encode in a source and then parse it right away instead of creating the tf.Source directly ?

modvendor/modvendor.go
```diff
@@ -155,3 +316,81 @@ func Vendor(rootdir, vendordir string, modsrc tf.Source) (string, error) {
...
+				}
+			}
+		}
+
+		newcontent := parsedFile.Bytes()
+		errs.Append(ioutil.WriteFile(fname, newcontent, 0644))
```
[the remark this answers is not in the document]
i4ki: I think it's a good idea. Thanks

modvendor/modvendor_test.go
```diff
@@ -111,31 +948,28 @@ func TestModVendorWithRef(t *testing.T) {
 	// or else the test passes even if the correct ref is not used.
 	repogit.Checkout(ref)
 
-	newCloneDir, err := modvendor.Vendor(rootdir, vendordir, tf.Source{
+	source = tf.Source{
```
katcipis, line 951: hmm here we kept the tf.Source initialization instead of parsing the URL :thinking: (Im probably missing something on the other ones)

modvendor/modvendor_report.go
i4ki: I think tabs eat too much space in the terminal in this case, but I can change if you think it's better.

modvendor/modvendor_test.go
i4ki: because it reads better in the table tests (IMO).. flat > nested in this case I guess. But not a big deal, can change back.

modvendor/modvendor_report.go
katcipis: Hmm no strong opinion, but then we have a inconsistency with the code generation report :thinking:. Lets leave it like that and discuss (the output is experimental anyway).

modvendor/modvendor_test.go
```diff
@@ -164,23 +1004,22 @@ func TestModVendorDoesNothingIfRefExists(t *testing.T) {
...
 	s := sandbox.New(t)
 	gitURL := "file://" + s.RootDir()
 	rootdir := t.TempDir()
 
-	_, err := modvendor.Vendor(rootdir, "/vendor", tf.Source{
-		URL:  gitURL,
-		Path: path,
-	})
+	source, err := tf.ParseSource(fmt.Sprintf("git::%s", gitURL))
+	assert.NoError(t, err)
```
katcipis, line 1012: If we go in the directionof always parsing and never creating tf.Source directly we could have a helper function to do this. Something like:

```suggestion
	source := parseSource(gitURL)
```

And it automatically adds the git:: prefix + fails if it is invalid etc.

modvendor/modvendor_test.go
katcipis: hmm I can see your point...but I do get a sense of extra indirection that can cause some confusion. When it is just want and wantErr it is easier to understand, but in this case it is two slices that are related and will be composed in a final report...IDK :thinking: . Lets leave it like this, it is not bad, was just curious on the reasoning and I never felt annoyed by the nesting :laughing: .

modvendor/modvendor_test.go
```diff
@@ -212,8 +1051,35 @@ func assertNoGitDir(t *testing.T, dir string) {
 	}
 }
 
-func vendoredDir(rootdir, vendordir, path, ref string) string {
-	return filepath.Join(rootdir, vendordir, path, ref)
+func assertVendorReport(t *testing.T, want, got modvendor.Report) {
+	assert.EqualInts(t, len(want.Vendored), len(got.Vendored),
```
katcipis, line 1055:
```suggestion
func assertVendorReport(t *testing.T, want, got modvendor.Report) {
	t.Helper()
	
	assert.EqualInts(t, len(want.Vendored), len(got.Vendored),
```

modvendor/modvendor_test.go
```diff
@@ -212,8 +1051,35 @@ func assertNoGitDir(t *testing.T, dir string) {
...
+		}
+	}
+	for i, wantIgnored := range want.Ignored {
+		if wantIgnored.RawSource != got.Ignored[i].RawSource {
+			t.Errorf("want.RawSource %v is different than %v",
+				wantIgnored.RawSource, got.Ignored[i].RawSource)
+		}
+		if wantIgnored.Reason != "" {
+			if ok, _ := regexp.MatchString(wantIgnored.Reason, got.Ignored[i].Reason); !ok {
+				t.Errorf("want.Reason %v is different than %v",
```
katcipis, line 1074:
```suggestion
				t.Errorf("want.Reason %v pattern does not match %v",
```

modvendor/modvendor_test.go
i4ki: yep

modvendor/modvendor_test.go
i4ki: less code? also if we change the way we encode the source later (or add new fields), we have less places to fix, and the tf.ParseSource() will fill all fields and catch errors, etc.

modvendor/modvendor_test.go
i4ki: I just prefer the other way, as I explained, it could last longer as we use only the Parse API.

modvendor/modvendor_test.go
katcipis: hmm create one struct passing 3 fields VS calling Sprintf + a parseing function...Im not so sure it is less code :thinking:. But not a strong opinion on this...on the spirit of even less code would extract this into a function that already checks the error + does the printf.

I get the point of preparing for future changes...it just feels funny that you already have parsed information, then you glue it together to parse it again..and Im not so sure if this will really change or if we will have wild information being extracted from the parsing in the future.

modvendor/modvendor_test.go
katcipis: fair enough, but then we need to fix it here to use the parse no ?

modvendor/modvendor_test.go
i4ki: the problem is that we have to try to clone() in order to know the issues (not exists, ref doesn't exists, etc), so yeah now it returns in the `ignored` list but we can improve later.

modvendor/modvendor_test.go
i4ki: it depends on if you agree or not... I can revert the code to the way it was before.

modvendor/modvendor_test.go
i4ki: you mean:

```
source := parseSource(t, "git::%s?ref=%s", gitURL, ref)
```
right?

modvendor/modvendor_test.go
katcipis: lets use the parse function :+1:

modvendor/modvendor_test.go
katcipis: Actually no, I think all formats are the same appending git::, so I was thinking on making it simpler and not spread around the whole formatting around the tests, it could even be something like newSource, then the formatting + parsing is just a detail to create the source correctly. Or we can leave as is, it is not bad either, just giving stupid ideas.

modvendor/modvendor_test.go
i4ki: just made this work. Thanks

katcipis (approved):

modvendor/modvendor_test.go
katcipis: Just to explain better at least the point already more code. Initializing the source struct directly in a smilar way to how the parsing option is done would be done like this:

```go
source := tf.Source{ URL:  gitURL, Ref:  ref, Path: path }
funcall(source)
```

It is a single line to create the source with information already available + call something that uses it. The current code was passing it directly on the function call and then we break into multiple lines, but if we compare with the parsing approach that has its own line it would be a more fair comparison. It is also one line and it involves only a simple struct intialization, IMHO less code than this:

```go
source, err := tf.ParseSource(fmt.Sprintf("git::%s?ref=%s", gitURL, ref))
assert.NoError(t, err)
funcall(source)
```

On top of the extra error handling, even though it is a single line you have both a printf + a parsing happening on this single line, IMHO that is more code.

But depending on future changes it could happen that the parsing function could already fill out fields for us, but that is more related to the idea of having a test specific constructor that abstract details than having to parse things necessarily, but could be an advantage. So in general I'm not pushing for not using the parse function, but I wanted to defend a little why I said it is more code =P

PR #558, opened 2022-08-26
fix: panic when parsing terraform files with empty module.source

i4ki:


manager_test.go
```diff
@@ -258,6 +265,21 @@ func singleNotChangedStack(t *testing.T) repository {
 	}
 }
 
+func singleNotChangedStackWithEmptyModule(t *testing.T) repository {
```
katcipis, line 268:
```suggestion
func singleNotChangedStackWithEmptyModuleSrc(t *testing.T) repository {
```

Or else it seems you are creating an empty module block IMHO. But not a huge deal, ignore at will.

katcipis (approved):
Nice catch :100:

katcipis (approved):

manager_test.go
```diff
@@ -258,6 +265,21 @@ func singleNotChangedStack(t *testing.T) repository {
 	}
 }
 
+func singleNotChangedStackWithEmptyModuleSrc(t *testing.T) repository {
```
katcipis, line 268: Ah sorry the suggestion broke the code, you need a proper full replace on the whole code

katcipis (changes requested):
Renaming func broke tests

katcipis (approved):

PR #564, opened 2022-08-31
feat: vendor: support manifest to filter copied files

katcipis:
# Reason for This Change

Add support to manifest files when vendoring, so we don't always vendor the entire project. The order of precedence for configuration loading is:

- .terramate
- root

If there is no manifest by default we filter out .terramate from root (/.terramate).

To check it working e2e you can run:

```
terramate experimental vendor download github.com/mineiros-io/terraform-google-pubsub-topic main
```

Which has a manifest.


modvendor/modvendor_manifest_test.go
```diff
@@ -0,0 +1,386 @@
...
+				"/file2",
+			},
+		},
+	}
+
+	for _, tcase := range testcases {
+		t.Run(tcase.name, func(t *testing.T) {
+			repoSandbox := sandbox.New(t)
+
+			// Remove the default README.md created by the sandbox
```
i4ki, line 288: I think we should get rid of this, right? I used this in the past just to create the first commit and then create the HEAD and main ref.. but we can just add another commit removing it and then all new repos are created clean.
Dont need to change now. Just documenting so we can remember of doing this next time we touch the sandbox.

i4ki (approved):
:rocket: :rocket: :rocket:

modvendor/modvendor_manifest_test.go
katcipis: Agreed, would be cool to remove it, I was quite surprised/puzzled and it took me some time to remember this :laughing:

PR #566, opened 2022-09-02
fix: delete outdated generated code non stacks

katcipis:
# Reason for This Change

Today if you remove the stack configuration from a stack, and it stops being a stack, any generated code is left there forever.

## Description of Changes

After code generation is dont and all stacks are updated we walk through the project deleting any terramate generated code on a dir that is not a stack. A very simple function to check if a dir is a stack has been added.


generate/generate.go
```diff
@@ -182,54 +183,52 @@ func Do(root string, workingDir string) Report {
...
-		if other, ok := genset[file.Name()]; ok && file.Condition() {
-			return errors.E(ErrConflictingConfig,
-				"configs from %q and %q generate a file with same name %q have "+
-					"`condition = true`",
-				file.Origin(),
-				other.Origin(),
-				file.Name(),
-			)
-		}
+	outdatedDirs, err := listGenFilesOutsideStacks(rootdir, rootdir)
```
katcipis, line 187: I'm thinking that we could use this later to do outdated code detection (not actual removal). Didn't fixed it on this PR because it seemed like the PR would be too big, so this one is focused only on actually removing outdated files.

generate/generate.go
```diff
@@ -653,3 +652,90 @@ func (ss *stringSet) slice() []string {
...
+	return nil
+}
+
+type dirGenFiles struct {
+	dir   string
+	files []string
+}
+
+// listGenFilesOutsideStacks returns a map of dir -> generated files.
+func listGenFilesOutsideStacks(rootdir, dir string) ([]dirGenFiles, error) {
```
i4ki, line 697: I think this could be much simpler if you handle both ListGenFiles() and the listGenFilesOutsideStacks() in the same function.
```
func SearchGeneratedFiles(rootdir string) (active []string, orphaned []string, err error) {
    err := filetype.WalkDir(rootdir, func(path string, d filetype.DirEntry, err error) error {
        if err != nil {
            return err
        }
        
        if d.IsDir() && filename.Base(path) == ".git" {
		    return filepath.SkipDir
		}
    
        if d.IsDir() {
            continue
        }
    
        data, err := os.ReadFile(path)
        // err handling
    
        if !hasGenHCLHeader(string(data)) {  
            return nil
        }
          
        if config.IsStack(filepath.Dir(path)) {
            active = append(active, path)
        } else {
            orphaned = append(orphaned, path)
        }

        return nil
    })
    
    if err != nil {
        return nil, nil, err
    }
    
    return active, orphaned, nil
}
```

Because in the end you need to traverse all files in the project anyway.
Would something like this be enough?

i4ki (approved):
lgtm but left some improvement ideas.

generate/generate.go
katcipis: Maybe, perhaps we could improve this in the future.

PR #568, opened 2022-09-02
chore: update macos runners

katcipis:
# Reason for This Change

macos-10.15 is being deprecated. Sorry I added it back again, I wasn't able to find this before:

https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

And apparently the best way to communicate deprecation is by making things fail inside a schedule, so people can observe the behavior of "sometimes it work, sometimes it breaks" and then realize that obviously that means the runner is being deprecated.

And the workflow is just marked as cancelled...with no information on why, kudos to GHA :laughing:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
       fail-fast: false
       matrix:
         # maybe in the future: windows-2019, windows-2022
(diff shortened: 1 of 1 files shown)


i4ki:
oh, it randomly fails, it must mean deprecation...

i4ki (approved):

PR #573, opened 2022-09-05
chore: refactor dag to be independent of the run-order semantics

i4ki:


cmd/terramate/cli/cli.go
```diff
@@ -946,7 +946,7 @@ func generateDot(
 		Logger()
 
 	parent := dotGraph.Node(getLabel(stackval))
-	for _, childid := range graph.ChildrenOf(id) {
```
i4ki, line 949: For general graphs, it really depend on how you look at them, but for directed graphs, the dependency arrow is explicit and calling this method "ChildrenOf" was wrong, it actually return the parentsOf/ancestorsOf... The change has no other implication in the code.

zwopir (approved):

PR #581, opened 2022-09-08
fix: recursive vendor does not patch dependencies when already vendored

i4ki:


modvendor/modvendor.go
```diff
@@ -235,9 +235,11 @@ func vendorAll(rootdir string, vendorDir string, tfdir string, report Report) Re
 
 		info.subdir = modsrc.Subdir
 
-		if v, ok := report.Vendored[Dir(vendorDir, modsrc)]; ok {
```
i4ki, line 238: the report knows only about the current "vendor" run. We must always check the disk to verify which modules are already vendored.

mariux: `Vendored[Dir(vendorDir, modsrc)];` should be true even if no download happened? why do we need to check the existence of the directory twice? once for deciding to download new and the again.. if we find the directory and decide not to download we could just mark it as vendored too?

i4ki: this line is deleted, we are not checking it twice... :thinking:

mariux: i mean we checked it while vendoring why check it again when patching the source? we replace the "cached version" with a check?

i4ki: Have a look in the use case below:

modA and modB depends on modC.

1. terramate experimental vendor download modA
  a. vendors modC
  b. patches modA files to reference /modules/modC/<ref>
3. terramate experimental vendor download modB
  a. Line 238: *This check in the filesystem detects that modC is already vendored by an earlier invocation of Terramate, so doesn't need to vendor again*
  b. patches modB files to reference /modules/modC/<ref>

mariux: i mean if a module sources the same remote 3 times... we check the existence of the directory 3 times: statdir, vendor, statdir again, statdir again... ?

mariux:
can we get this merged? @zwopir @i4ki

mariux (approved):
lgtm

modvendor/modvendor.go
i4ki: for sure we can cache the "existence of modules" once but not sure if it's worth the effort... because still the cache can be inconsistent if other vendor invocations ran in parallel etc, at least this way we minimize the window for those races.

PR #586, opened 2022-09-15
feat: add lets block to generate_* configs

i4ki:
This PR introduces the `lets` block for *generate_hcl* and *generate_file* blocks.
The **lets** is intended to be used as local or temporary variables, so its scope is the generate_hcl/generate_file block where they are defined.

Example:

```
generate_hcl "file.tf" {
    condition = let.length >= let.minimum && let.length <= let.maximum

    lets {
        minimum = 1
        maximum = 5
        length  = tm_length(global.values)
    }
    
    content {
        values = global.values
    }
}
```


generate/genfile/genfile_test.go
```diff
@@ -825,6 +826,140 @@ stack_id=stack-id
...
+							Str("string", "let string"),
+						),
+						Lets(
+							Str("string", "dup"),
+						),
+						Expr("content", `let.string`),
+					),
+				},
+			},
+			wantErr: errors.E(lets.ErrLetsRedefined),
```
katcipis, line 961: Maybe lets.ErrRedefined to avoid stuttering.

lets/lets.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/mineiros-io/terramate/hcl"
+	"github.com/mineiros-io/terramate/hcl/eval"
+	"github.com/rs/zerolog/log"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Errors returned when parsing and evaluating lets.
+const (
+	ErrLetsEval      errors.Kind = "lets eval failed"
+	ErrLetsRedefined errors.Kind = "lets redefined"
```
katcipis, line 30: Maybe use ErrEval/ErrRedefined to avoid stuttering.

lets/lets.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Value is an evaluated let.
+	Value struct {
+		Origin string
+
+		cty.Value
+	}
+
+	// L is an evaluated lets map.
+	L map[string]Value
```
katcipis, line 54: Maybe rename to Map to be consistent with globals.Map

generate/genfile/genfile.go
```diff
@@ -130,6 +132,11 @@ func Load(projmeta project.Metadata, sm stack.Metadata, globals globals.G) ([]Fi
 			Str("origin", origin).
 			Logger()
 
+		err := lets.Load(genFileBlock.lets, evalctx.Context)
```
katcipis, line 135: Inside this loop we call Load with the same evalctx.Context multiple times for each genFileBlock without cleaning up the evalctx.Context "lets" namespace, at least I didn't find where we do this. If we don't cleanup won't the lets namespace loaded from a generate_file block be available to another generate_file block ?

lets.Load only sets the namespace to an empty map if the namespace doesn't exist, so it seems that when it already exists it will only add more (or override) what is already there from previous loads. Sorry if I missed where it is cleanup up, if that is the case just disregard the comment.

lets/lets.go
```diff
@@ -0,0 +1,228 @@
...
+	logger := log.With().
+		Str("action", "Exprs.Eval()").
+		Logger()
+
+	lets := L{}
+	pendingExprsErrs := map[string]*errors.List{}
+	pendingExprs := make(Exprs)
+
+	copyexprs(pendingExprs, letExprs)
+	removeUnset(pendingExprs)
```
katcipis, line 84: Hmm do we support unset on lets ? the scope of lets is so small that it doesn't seem useful, but it makes it consistent with globals :thinking:

generate/genfile/genfile.go
katcipis: On generate_hcl we create a new context per generate_hcl block, maybe we could do the same here (and also have some tests to cover this, the test exists on generate_hcl but I think it is missing here).

katcipis (commented):
Some considerations on cleanup between successives lets.Load calls, maybe I missed something (as usual) :laughing:

generate/genfile/genfile.go
i4ki: yep I fixed this in genhcl but forgot here. Thanks.

katcipis (approved):
:rocket: :rocket: :rocket:

PR #590, opened 2022-09-20
chore: avoid cleanup dotdirs on generate

katcipis:
# Reason for This Change

We are currently checking for files to cleanup on generate inside dot dirs.

## Description of Changes

Filter out dot dirs when cleaning up files.


i4ki (approved):
:rocket:

PR #595, opened 2022-09-21
feat: add support to assert block

katcipis:
# Reason for This Change

We can now parse/validate assert blocks. The next step will be to use the parsed config on code generation to check the assertions.

## Description of Changes

Add assertions parsing on hcl.Config.


test/hcl.go
```diff
@@ -74,6 +74,7 @@ func AssertTerramateConfig(t *testing.T, got, want hcl.Config) {
 
 	assertTerramateBlock(t, got.Terramate, want.Terramate)
 	assertStackBlock(t, got.Stack, want.Stack)
+	assertAssertsBlock(t, got.Asserts, want.Asserts)
```
katcipis, line 77: Don't know how to avoid the stuttering here :rofl:

test/hcl.go
```diff
@@ -93,6 +94,43 @@ func AssertDiff(t *testing.T, got, want interface{}, msg ...interface{}) {
...
+			"warning expr mismatch")
+	}
+}
+
+// exprAsStr gets the original expression as a string for an expression built
+// with our
+func exprAsStr(t *testing.T, expr hhcl.Expression) string {
+	t.Helper()
+
+	if expr == nil {
```
katcipis, line 124: Not sure this is a good idea, we rarely test anything with empty expressions so it seemed safe to do this in order to avoid checking if both are nil before comparing etc.

test/hcl.go
```diff
@@ -93,6 +94,43 @@ func AssertDiff(t *testing.T, got, want interface{}, msg ...interface{}) {
...
+			exprAsStr(t, w.Message), exprAsStr(t, g.Message),
+			"message expr mismatch")
+		assert.EqualStrings(t,
+			exprAsStr(t, w.Warning), exprAsStr(t, g.Warning),
+			"warning expr mismatch")
+	}
+}
+
+// exprAsStr gets the original expression as a string for an expression built
+// with our
```
i4ki, line 120: the comment is not finished.

i4ki (dismissed):
lgtm

config/assert.go
```diff
@@ -0,0 +1,97 @@
...
+func evalBool(evalctx *eval.Context, expr hhcl.Expression, name string) (bool, error) {
+	if expr == nil {
+		return false, errors.E(ErrSchema, "%s must be defined", name)
+	}
+	val, err := evalctx.Eval(expr)
+	if err != nil {
+		return false, errors.E(err, "evaluating %s", name)
+	}
+	if val.Type() != cty.Bool {
+		return false, errors.E(ErrSchema, "%s must be boolean, got %v", name, val.Type())
```
i4ki, line 80:
```suggestion
		return false, errors.E(ErrSchema, "%s must be boolean, got %v", name, val.Type().FriendlyName())
```

config/assert.go
```diff
@@ -0,0 +1,97 @@
...
+func evalString(evalctx *eval.Context, expr hhcl.Expression, name string) (string, error) {
+	if expr == nil {
+		return "", errors.E(ErrSchema, "%s must be defined", name)
+	}
+	val, err := evalctx.Eval(expr)
+	if err != nil {
+		return "", errors.E(err, "evaluating %s", name)
+	}
+	if val.Type() != cty.String {
+		return "", errors.E(ErrSchema, "%s must be string, got %v", name, val.Type())
```
i4ki, line 94:
```suggestion
		return "", errors.E(ErrSchema, "%s must be string, got %v", name, val.Type().FriendlyName())
```

i4ki (approved):
🚀 🚀

i4ki (approved):

PR #605, opened 2022-09-23
feat: generate: integrate assert on code generation

katcipis:


i4ki (approved):
:rocket: :rocket:

generate/generate.go
```diff
@@ -81,6 +87,34 @@ func Do(rootdir string, workingDir string) Report {
...
+		if err != nil {
+			report.err = err
+			return report
+		}
+
+		logger.Trace().Msg("checking stack asserts")
+		errs := errors.L()
+		for _, assert := range asserts {
+			if !assert.Assertion {
+				assertRange := assert.Range
```
katcipis, line 102: maybe we could have our own Range type where the filename is always a project.Path, and then a function to do conversion (we do this on other places too I think, or something similar at least). not sure, just an idea.

i4ki: 👍

PR #606, opened 2022-09-23
chore: add ci aggregator job

katcipis:
Since we have build matrices instead of defining a long list of required checks we use a "ci" job that aggregates all required checks, even dynamic ones like the matrix.


i4ki (approved):
:rocket:

PR #609, opened 2022-09-24
fix: properly emulate windows prompt for signal tests.

i4ki:


cmd/terramate/e2etests/run_signal_test.go
```diff
@@ -33,8 +33,7 @@ func TestRunSendsSigkillIfCmdIgnoresInterruptionSignals(t *testing.T) {
 	git.Add(".")
 	git.CommitAll("first commit")
 
-	tm := newCLI(t, s.RootDir())
-	tm.loglevel = "trace"
+	tm := newCLIWithLogLevel(t, s.RootDir(), "trace")
```
katcipis, line 36: We are going to keep trace level on this test ? (it is usually disabled by default)

i4ki: It was already "trace", I just used the constructor instead of setting the `tm.logLevel` by hand.

i4ki: but I think they could be disabled

katcipis: ah yeah I probably forgot then as trace =P

PR #612, opened 2022-09-25
fix: use portable binaries in tests

i4ki:


cmd/terramate/e2etests/cmd/test/test.go
```diff
@@ -34,6 +35,10 @@ func main() {
 		hang()
 	case "env":
 		env()
+	case "cat":
```
katcipis, line 38: noice

katcipis (approved):

PR #613, opened 2022-09-26
fix: generate: assert loading when proj root is /

katcipis:
# Reason for This Change

If a project root is on / we would loop forever on the HasPrefix logic (you never leave root).

## Description of Changes

When we reach root break explicitely.


i4ki (approved):
🚀

PR #619, opened 2022-09-28
feat: add tm_dynamic condition attribute

katcipis:
# Reason for This Change

Add a condition attribute on tm_dynamic. When it evaluates to false the block is ignored, including any nested tm_dynamic blocks.


i4ki (dismissed):
🚀

i4ki (approved):
lgtm

PR #620, opened 2022-09-28
chore: update linter

katcipis:
# Reason for This Change

Update golangci-linter. Even though there are not config changes the linter now validates for package comments. Since it is nice to have those I introduced them on all missing packages, we definitely can improve these docs in the future.

## Description of Changes

Update golangci-lint and add documentation for packages on the project.


cmd/terramate/main.go
```diff
@@ -12,6 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/*
```
i4ki, line 15: why this comment style?

i4ki (approved):
lgtm

cmd/terramate/main.go
katcipis: I copied from the docs on how to document commands, usually because the commands docs can get bigger, but on this case we can go with the more normal style:

https://go.dev/doc/comment (the Commands section)

i4ki: ok no problem, I was curious but indeed makes sense for huge comments.

i4ki (approved):

cmd/terramate/main.go
katcipis: I changed it for now since the comment is small, then we adapt accordingly in the future :+1:

PR #621, opened 2022-09-28
feat: generate_hcl: make tm_dynamic for_each optional

katcipis:
# Reason for This Change

Makes tm_dynamic.for_each optional.

Depends on #619


i4ki (approved):
🚀

PR #624, opened 2022-09-29
feat: generate: support subdirs

### LINKED ISSUE #503: [FEATURE] Add support for folder in generate_hcl block
Hi,

When dealing with multiple generated files from different modules, it could interesting to isolate generated files into some folders instead of simply use a "_" to order files.

Currently, the char "/" is not allowed in the file name making it impossible.

Many thanks

katcipis:
# Reason for This Change

Add support to sub dirs when generating code. So now we can do something like:

```hcl
generate_hcl "_gen/name.tf" {
}
```
And generate code inside a directory.

Closes #503 

## Description of Changes

Updated code generation, cleanup procedures and outdated code checks to work with subdirs.
The approach was to make stacks responsible for any subdirs inside them that have generated code that can't be traced to any configuration belonging to the stack. So stacks are owners of all their subdirs, except subdirs that are stack themselves.

Then we also have orphan directories, which have generated code but have no parent stack anywhere. Those are also detected now (it was not being detected before).

## Example

```
#!/bin/bash

set -o errexit
set -o nounset

workdir=$(mktemp -d)
cd "${workdir}"

echo "project: ${workdir}"

cat > terramate.tm <<-EOF
terramate {
  config {
  }
}
EOF

terramate create stack-1
terramate create stack-1/child
terramate create stack-1/child/grandchild
terramate create stack-2

cat > gen.tm <<-EOF
generate_hcl "file.hcl" {
    content {
        a = "data"
    }
}

generate_hcl "_gen/file.hcl" {
    content {
        a = "data"
    }
}

generate_file "_gen/file.txt" {
    content = "data"
}

generate_file "_gen/dir/file.txt" {
    content = "data"
}
EOF

terramate generate

echo
echo "project final structure"
tree .
```


generate/report_test.go
```diff
@@ -329,7 +329,13 @@ func assertEqualReports(t *testing.T, got, want generate.Report) {
 
 	if diff := cmp.Diff(got.Successes, want.Successes); diff != "" {
 		t.Errorf("success results differs: got(-) want(+)")
-		t.Fatal(diff)
+		// Don't call fatal to avoid vague errors like
```
katcipis, line 332: Don't call fatal to avoid vague errors like:

```
generate_test.go:61: success results differs: got(-) want(+)
generate_test.go:61:   []generate.Result(
		-   nil,
		+   {{Dir: s"/stacks/stack", Created: []string{"dir/file.hcl", "dir/sub/file.txt"}}},
)
```
With no info on what was the error/failure.

generate/generate_file_test.go
```diff
@@ -256,77 +256,6 @@ func TestGenerateFile(t *testing.T) {
 				},
 			},
 		},
-		{
```
katcipis, line 259: I moved tests related to subdir to a single place instead of testing on both genfile/genhcl. Seemed less error prone on this case.

generate/genhcl/genhcl.go
```diff
@@ -82,9 +82,9 @@ const (
 	ErrDynamicConditionEval errors.Kind = "evaluating tm_dynamic.condition"
 )
 
-// Name of the HCL code.
-func (h HCL) Name() string {
-	return h.name
+// Label of the original generate_hcl block.
```
katcipis, line 85: Trying to abstract as a name only made it confusing to me, so I changed it to be more concrete.

generate/report.go
```diff
@@ -147,6 +153,20 @@ func (r Report) empty() bool {
 		len(r.Successes) == 0
 }
 
+func (r *Report) sort() {
```
katcipis, line 156: now we manipulate the report in ways that is easy to generate it with random order (using maps). Other changes could also change the order, so it seems safer to just always sort the report before returning on code generation.

cmd/terramate/cli/cli.go
```diff
@@ -1346,34 +1346,19 @@ func (c *cli) checkOutdatedGeneratedCode(stacks stack.List) {
...
-			Logger()
-
-		logger.Trace().Msg("checking stack for outdated code")
+	outdatedFiles, err := generate.Check(c.root())
 
-		outdated, err := generate.CheckStack(projmeta, stack)
-		if err != nil {
-			logger.Fatal().Err(err).Msg("checking stack for outdated code")
-		}
+	fatalerr(logger, "failed to check outdated code on project", err)
```
katcipis, line 1353: I'm thinking on ways to log error lists better, this is my first take at it.

cmd/terramate/e2etests/run_test.go
```diff
@@ -1308,6 +1308,50 @@ func TestRunFailIfGitSafeguardUntracked(t *testing.T) {
...
+		testHelperBin,
+		"env",
+	), runExpected{
+		IgnoreStdout: true,
+	})
+
+	orphanEntry.RemoveFile("config.tm")
+	orphanEntry.DeleteStackConfig()
+
+	git.CommitAll("generated code")
```
katcipis, line 1343:
```suggestion
	git.CommitAll("deleted stack")
```

generate/generate.go
```diff
@@ -222,94 +200,222 @@ func Do(rootdir string, workingDir string) Report {
...
-				report.Failures = append(report.Failures, FailureResult{
-					Result: Result{
-						Dir:     relpath,
-						Deleted: deleted,
-					},
-					Error: err,
-				})
-				continue outdatedDirsLoop
+	deletedFiles := map[project.Path][]string{}
+	deleteFailures := map[project.Path]*errors.List{}
```
katcipis, line 210: Since these are not stacks each dir is reported independently, so we need to organize the report per directory here. Was on the fence between doing this on the report itself with a method or doing it here, in the end did it here with these maps.

generate/generate.go
```diff
@@ -646,24 +751,83 @@ func hasGenHCLHeader(code string) bool {
...
+
+		abspath := filepath.Join(stackpath, relFilename)
+
+		switch {
+		case strings.HasPrefix(relFilename, "/"):
+			errs.Append(errors.E(ErrInvalidGenBlockLabel,
+				"%s: %s: starts with /",
+				file.Origin(), file.Label()))
+			continue
+		case strings.HasPrefix(relFilename, "./"):
```
katcipis, line 777: I added this constraint because it makes some things simpler, or else ./name and name are actually the same, so a simple label checking is not enough. Still doable but it seemed OK to constrain on a uniform way to define the paths. But can change this, not a 100% sure it is a good idea.

generate/report_test.go
```diff
@@ -329,7 +329,7 @@ func assertEqualReports(t *testing.T, got, want generate.Report) {
 
 	if diff := cmp.Diff(got.Successes, want.Successes); diff != "" {
 		t.Errorf("success results differs: got(-) want(+)")
-		t.Fatal(diff)
+		t.Error(diff)
```
katcipis, line 332: it was hard to understand erros when success already didn't match (lack of differences on Failures).

generate/generate_test.go
```diff
@@ -57,6 +59,455 @@ func (s stringer) String() string {
...
+					},
+				},
+			},
+		},
+		{
+			name: "if path is symlink fails",
+			layout: []string{
+				"s:stacks/stack",
+				"d:somedir",
+				"l:somedir:stacks/stack/symlink",
```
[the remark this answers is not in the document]
katcipis: ah OK it is related to the test working on windows only if the user is admin, will fix this. Thanks :heart:

PR #629, opened 2022-10-07
feat: log: add disabled log level

katcipis:
# Reason for This Change

Add a "disabled" log level. When logging is disabled fatal calls still will abort the process, but silently. This can be tested by running in an empty dir:

```sh
terramate list --log-level=fatal
2022-10-07T13:03:11+02:00 FTL project root not found action=newCli() workingDir=/tmp/test
```

Versus:

```sh
terramate list --log-level=disabled
```

The status code indicates error, but not output. Our current version of zerolog would not abort the process when Fatal is called, which caused panics, but the newest 1.28.0 implements the desired behavior.


i4ki (approved):
lgtm

PR #630, opened 2022-10-07
refactor: extract cleanup of orphaned files

katcipis:
# Reason for This Change

It is easier to check on the benchmark how long we are taking to cleanup orphaned files after code generation. Also I think the code is easier to understand now.


i4ki (approved):
lgtm

PR #631, opened 2022-10-07
chore: load configuration once

i4ki:


config/config.go
```diff
@@ -32,28 +36,43 @@ const (
 	ErrSchema errors.Kind = "config has an invalid schema"
 )
 
-// TryLoadRootConfig try to load the Terramate root config. It looks for the
+// Tree is the config tree.
+type Tree struct {
+	// Root is the configuration of this tree node.
+	Root hcl.Config
```
katcipis, line 42: Maybe call it Node instead of Root, but not sure, feel free to disregard (just tossing the idea).

config/config.go
```diff
@@ -32,28 +36,43 @@ const (
...
+
+// TryLoadConfig try to load the Terramate configuration tree. It looks for the
 // the config in fromdir and all parent directories until / is reached.
 // If the configuration is found, it returns configpath != "" and found as true.
-func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found bool, err error) {
+func TryLoadConfig(fromdir string) (tree *Tree, configpath string, found bool, err error) {
 	for {
 		logger := log.With().
-			Str("action", "TryLoadRootConfig()").
+			Str("action", "config.TryLoadConfig()").
```
katcipis, line 59: noice

config/config.go
```diff
@@ -62,26 +81,147 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
 		}
 		fromdir = parent
 	}
-	return hcl.Config{}, "", false, nil
+	return nil, "", false, nil
 }
 
-// IsStack returns true if the given directory is a stack, false otherwise.
-func IsStack(rootdir, dir string) (bool, error) {
+func LoadTree(rootdir string, cfgdir string) (*Tree, error) {
```
katcipis, line 87:
```suggestion
// LoadTree loads a configuration tree, parsing all Terramate files. 
func LoadTree(rootdir string, cfgdir string) (*Tree, error) {
```

config/config.go
```diff
@@ -62,26 +81,147 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	return loadTree(rootdir, cfgdir, nil)
+}
+
+func (tree *Tree) Rootdir() string {
```
katcipis, line 91:
```suggestion
// Rootdir returns the root directory from where the tree was parsed.
func (tree *Tree) Rootdir() string {
```

config/config.go
```diff
@@ -62,26 +81,147 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	return tree.rootdir
+}
+
+func (tree *Tree) IsStack() bool {
```
katcipis, line 95:
```suggestion
IsStack returns true if the tree current node is a stack
func (tree *Tree) IsStack() bool {
```

config/config.go
```diff
@@ -62,26 +81,147 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	return tree.Root.Stack != nil
+}
+
+func (tree *Tree) Stacks() []*Tree {
```
katcipis, line 99:
```suggestion
Stacks returns all Tree nodes that are stacks.
func (tree *Tree) Stacks() []*Tree {
```

config/config.go
```diff
@@ -62,26 +81,147 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	for _, children := range tree.Children {
+		stacks = append(stacks, children.Stacks()...)
+	}
+
+	return stacks
+}
+
+func (tree *Tree) Lookup(path project.Path) (*Tree, bool) {
+	pathstr := path.String()
+	if len(pathstr) == 0 || pathstr[0] != '/' {
```
katcipis, line 113: hmm doesn't project.Path already guarantees these invariants ? maybe it should have a Valid method to check if the path is sane (because in Go it is hard to guarantee invariants on types etc). Feel free to disregard and move on, not important.

i4ki: There's this:
https://github.com/mineiros-io/terramate/blob/main/project/project.go#L43

but making project.Path a struct instead of a string alias would enforce even more.

config/config.go
i4ki: thanks, I left docs to the final. I still need to go over all of them and adjust.

katcipis: I imagined something like that, so I tried to help a little with them.

cmd/terramate/cli/cli.go
```diff
@@ -1667,13 +1682,13 @@ func lookupProject(wd string) (prj project, found bool, err error) {
 
 			logger.Trace().Msg("Load root config.")
 
-			cfg, err := hcl.ParseDir(rootdir, rootdir)
+			cfg, err := config.LoadTree(rootdir, rootdir)
```
katcipis, line 1685: Thinking if reloadcfg could be called loadcfg and be called here too, it seems to just load the config in general (there is no reload specific behavior I think).

cmd/terramate/e2etests/create_test.go
```diff
@@ -89,6 +89,8 @@ func TestCreateStack(t *testing.T) {
 
 		got := s.LoadStack(stackPath)
 
+		t.Logf("stack: %s", got)
```
katcipis, line 92: Is this log permanent ?

config/config.go
```diff
@@ -62,26 +87,173 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	sort.Slice(l, func(i, j int) bool {
+		return l[i].RootDir() < l[j].RootDir()
+	})
+}
+
+func loadTree(rootdir string, cfgdir string, rootcfg *hcl.Config) (*Tree, error) {
 	logger := log.With().
-		Str("action", "config.IsStack()").
-		Str("rootdir", rootdir).
+		Str("action", "config.LoadTree()").
```
katcipis, line 175:
```suggestion
		Str("action", "config.loadTree()").
```

config/config.go
```diff
@@ -62,26 +87,173 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	}
+	for _, children := range tree.Children {
+		stacks = append(stacks, children.Stacks()...)
+	}
+
+	stacks.Sort()
+	return stacks
+}
+
+// Lookup a node from the tree.
```
katcipis, line 123: Maybe we should make it explicit on the doc that it will ignore the root node.

cmd/terramate/cli/cli.go
i4ki: new implementation has no need to reload configuration anymore, now the commands that do change configuration also reload the sub portion of the tree that changed.

config/config.go
i4ki: it doesn't skip the root node, the old code had a misleading comment, I just changed it.
thanks

cmd/terramate/cli/cli.go
```diff
@@ -932,10 +929,11 @@ func (c *cli) generateGraph() {
 
 		if err := run.BuildDAG(
 			graph,
-			c.root(),
+			c.cfg(),
 			e.Stack,
-			loader,
```
katcipis, line 937: no more loader :partying_face:

cmd/terramate/cli/cli.go
```diff
@@ -1569,7 +1568,7 @@ func (c cli) checkVersion() {
 
 	logger.Trace().Msg("checking if terramate version satisfies project constraint")
 
-	rootcfg := c.prj.rootcfg
+	rootcfg := c.prj.cfg.Node
```
katcipis, line 1571: noice :100:

config/config.go
```diff
@@ -62,26 +88,259 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+func (tree *Tree) Dir() string {
+	return tree.dir
+}
+
+// RootDir returns the tree root directory..
+func (tree *Tree) RootDir() string {
+	if tree.Parent != nil {
+		return tree.Parent.RootDir()
+	}
+	return tree.dir
```
katcipis, line 110:
```suggestion
	return tree.Root().dir
```

config/config.go
```diff
@@ -62,26 +88,259 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	return cfg, true
+}
+
+// StacksByPaths returns the stacks from the provided relative paths.
+func (tree *Tree) StacksByPaths(base project.Path, relpaths ...string) List {
 	logger := log.With().
-		Str("action", "config.IsStack()").
-		Str("rootdir", rootdir).
+		Str("action", "tree.StacksByPath").
+		Str("basedir", base.String()).
```
katcipis, line 173:
```suggestion
		Stringer("basedir", base).
```

config/config.go
```diff
@@ -62,26 +88,259 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+			return nil, errors.E(err, "failed to load config from %s", dir)
+		}
+
+		node.Parent = tree
+		tree.Children[name] = node
+	}
+	return tree, nil
+}
+
+// LoadSubTree loads a subtree into the current tree. The destdir
```
katcipis, line 278:
```suggestion
// LoadSubTree loads a subtree located at cfgdir into the current tree.
```

config/config_test.go
```diff
@@ -32,18 +33,231 @@ func TestIsStack(t *testing.T) {
...
+			relpaths: []string{
+				"../../stack",
+				"../../stack",
+			},
+			want: []string{
+				"/stack",
+			},
+		},
+		{
+			name: "single stack, 1 matching relpath",
```
katcipis, line 170: Isn't this test a duplicate from [here](https://github.com/mineiros-io/terramate/pull/631/files#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R139) ?

config/config_test.go
```diff
@@ -32,18 +33,231 @@ func TestIsStack(t *testing.T) {
...
+			},
+			want: []string{
+				"/a",
+				"/a/lot",
+				"/a/lot/of",
+				"/a/lot/of/stacks",
+				"/a/lot/of/stacks/with-non-stack-dir/and",
+				"/a/lot/of/stacks/with-non-stack-dir/more",
+				"/a/lot/of/stacks/with-non-stack-dir/stacks",
+				"/b/is/not/stack/but-this-is-a-stack",
```
katcipis, line 211: :laughing:

config/config.go
```diff
@@ -62,26 +88,259 @@ func TryLoadRootConfig(fromdir string) (cfg hcl.Config, configpath string, found
...
+	return nil
+}
+
+// IsEmptyConfig tells if the configuration is empty.
+func (tree *Tree) IsEmptyConfig() bool {
+	return tree.Node.IsEmpty()
+}
+
+// IsStack returns true if the given directory is a stack, false otherwise.
+func IsStack(cfg *Tree, dir string) bool {
```
katcipis, line 326:
```suggestion
// IsStack returns true if the given directory is a stack, false otherwise.
// The given dir is an host absolute path.
func IsStack(cfg *Tree, dir string) bool {
```

errors/error.go
```diff
@@ -421,6 +421,16 @@ func IsKind(err error, k Kind) bool {
 	return Is(err, E(k))
 }
 
+// IsAnyKind returns true if err is of any of the provided kinds.
+func IsAnyKind(err error, kinds ...Kind) bool {
```
katcipis, line 425: noice :100:

generate/generate.go
```diff
@@ -72,24 +71,24 @@ const (
 // on code generation, any failure found is added to the report but does not abort
 // the overall code generation process, so partial results can be obtained and the
 // report needs to be inspected to check.
-func Do(rootdir string, workingDir string) Report {
-	report := forEachStack(rootdir, workingDir, func(
+func Do(cfg *config.Tree, workingDir string) Report {
```
katcipis, line 74: As we talked in a call, we can improve the workingDir thing here later.

generate/generate.go
```diff
@@ -850,60 +832,41 @@ func validateGeneratedFiles(rootdir string, stackpath string, generated []genCod
...
 	errs := errors.L()
 
-	for strings.HasPrefix(curdir, meta.Rootdir()) {
+	for {
 		logger = logger.With().
-			Str("curdir", curdir).
+			Stringer("curdir", curdir).
 			Logger()
 
-		logger.Trace().Msg("parsing config")
```
katcipis, line 875: noice :100:

generate/generate_file_check_test.go
```diff
@@ -121,6 +121,7 @@ func TestCheckStackForGenFile(t *testing.T) {
 			),
 		).String(),
 	)
+	s.ReloadConfig()
```
katcipis, line 124: hmm maybe move s.ReloadConfig() inside assertOutdatedFiles ? Seems like we reload for the assert.

generate/generate_file_check_test.go
```diff
@@ -204,7 +210,10 @@ func TestCheckOutdatedDetectsEmptyGenerateFileContent(t *testing.T) {
 		).String(),
 	)
 
+	s.ReloadConfig()
```
katcipis, line 213: same thing here, maybe it could be on the assert function.

generate/generate_file_check_test.go
```diff
@@ -204,7 +210,10 @@ func TestCheckOutdatedDetectsEmptyGenerateFileContent(t *testing.T) {
...
 	assertOutdatedFiles([]string{"test.txt"})
+
+	s.ReloadConfig()
```
katcipis, line 216: hmm if generate has the same issues with outdatedconfig maybe we could always reload the config also inside generate, or else I imagine this will be an issue most of the time we change files and usually we want s.Generate to be updated.

katcipis: I know it can make things slower, but not sure if in this case it will be worth the complexity of always ensuring the reload everywhere + when you forget I imagine that debugging can be challenging if you don't remeber this detail etc. But feel free to disregard, not a huge issue.

generate/genhcl/genhcl.go
```diff
@@ -248,49 +247,44 @@ type dynBlockAttributes struct {
...
-	blocks, err := hcl.ParseGenerateHCLBlocks(rootdir, cfgdir)
-	if err != nil {
-		return nil, errors.E(ErrParsing, err, "cfgdir %q", cfgdir)
-	}
-
-	logger.Trace().Msg("Parsed generate_hcl blocks.")
 	res := []loadedHCL{}
+	cfg, ok := tree.Lookup(cfgdir)
+	if ok && !cfg.IsEmptyConfig() {
+		blocks := cfg.Node.Generate.HCLs
```
katcipis, line 262: noice :100:

hcl/hcl.go
```diff
@@ -648,7 +656,9 @@ func (c Config) AbsDir() string { return c.absdir }
 
 // IsEmpty returns true if the config is empty, false otherwise.
 func (c Config) IsEmpty() bool {
-	return c.Stack == nil && c.Terramate == nil
+	return c.Stack == nil && c.Terramate == nil &&
+		c.Vendor == nil && len(c.Asserts) == 0 &&
+		len(c.Generate.Files) == 0 && len(c.Generate.HCLs) == 0
```
katcipis, line 661: Isn't the empty check missing Globals and Imported ? (if a config only has globals or just an import block etc).

katcipis (dismissed):

katcipis (dismissed):
:rocket: :rocket: :rocket:

katcipis (approved):

PR #638, opened 2022-10-11
chore: improve logging for git checks

katcipis:
# Reason for This Change

Improve logging of warning/fatals when we have uncommited/untracked files. We have each file logged individually to avoid long lists inside a single log message using the debug level and then a simple warn/fatal with no file information.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -645,38 +645,35 @@ func (c *cli) checkGitUncommited() bool {
 	return true
 }
 
-func (c *cli) gitSafeguards(checks terramate.RepoChecks, shouldAbort bool) {
-	logger := log.With().
-		Str("action", "gitSafeguards()").
-		Logger()
+func debugFiles(files []string, msg string) {
+	for _, file := range files {
+		log.Debug().
+			Str("file", file).
+			Msg(msg)
+	}
+}
 
+func (c *cli) gitSafeguards(checks terramate.RepoChecks, shouldAbort bool) {
 	if c.parsedArgs.Run.DryRun {
 		return
 	}
 
-	if c.checkGitUntracked() && len(checks.UntrackedFiles) > 0 {
-		if shouldAbort {
-			logger.Fatal().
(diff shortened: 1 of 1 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -645,38 +645,35 @@ func (c *cli) checkGitUncommited() bool {
...
+
+	var logger *zerolog.Event
+	if shouldAbort {
+		logger = log.Fatal()
+	} else {
+		logger = log.Warn()
+	}
 
+	if c.checkGitUntracked() && len(checks.UntrackedFiles) > 0 {
+		logger.Msg("repository has untracked files")
```
i4ki, line 672: hmm I don't like this approach much because now this unsuspicious logging can abort... the other way it was explicit when it aborts... but maybe it's just me.

cmd/terramate/cli/cli.go
```diff
@@ -645,38 +645,35 @@ func (c *cli) checkGitUncommited() bool {
...
-		if shouldAbort {
-			logger.Fatal().
-				Strs("files", checks.UntrackedFiles).
-				Msg("repository has untracked files")
-		} else {
-			logger.Warn().
-				Strs("files", checks.UntrackedFiles).
-				Msg("repository has untracked files")
-		}
+	debugFiles(checks.UntrackedFiles, "untracked file")
```
i4ki, line 661: cool

i4ki (approved):
lgtm, feel free to ignore the comments as they are very personal.

cmd/terramate/cli/cli.go
katcipis: hmm the logger is created pretty close to here and you can see clearly when it is fatal and warning, maybe if the function was bigger and had more stuff on it but on this case it seemed OK to me. but can change it too, no hard feelings on this.

katcipis: better to be explicit, take a look now :-)

i4ki (approved):
lgtm

PR #639, opened 2022-10-11
chore: improve warning messages

katcipis:
# Reason for This Change

The idea is to log less detailed information on warn messages, leaving details to other levels like debug/trace. For info/warning/fatal I think we could drop the "action" field since it is quite specific to details of the code (private function names, etc).


i4ki (approved):
lgtm

PR #640, opened 2022-10-12
chore: add time flag to make bench

i4ki:


i4ki:
I've noticed that the default bench counter for e2e tests is suboptimal for testing in unpredictable machines such as our laptops/desktops then this flag would help to ensure enough tests are taken to average better the result.

katcipis (approved):

PR #642, opened 2022-10-12
chore: improve info logging on run

katcipis:


run/exec.go
```diff
@@ -140,7 +137,7 @@ func Exec(
 		}
 
 		if interruptions > 0 {
-			logger.Trace().Msg("interrupting execution of further stacks")
+			logger.Info().Msg("interrupting execution of further stacks")
```
katcipis, line 140: Reviewing here it seems this is more like an Info, if other stacks are interrupted.

i4ki (approved):
lgtm

PR #643, opened 2022-10-12
chore:logging: improve fatal msgs

katcipis:
The idea is to improve logging of fatal msgs. We added functions to help logging our Terramate specific errors more nicely, so when we have error lists (like when parsing failed) it will log each error on its own log entry with the error level, so we don't lose information but we also don't have a clobbered single fatal message.

A parsing error when doing vendoring for example:

```
terramate experimental vendor download github.com/mineiros-io/terramate main
2022-10-13T15:19:54+02:00 ERR terramate schema error: unrecognized block "wrong"
2022-10-13T15:19:54+02:00 ERR terramate schema error: unrecognized block "wrong2"
2022-10-13T15:19:54+02:00 FTL parsing vendor dir configuration on .terramate
```

It shows all the parsing errors and then a final fatal with the message. Also added a function to be used for warnings but will change warnings in a different PR later.


i4ki (approved):
lgtm

PR #645, opened 2022-10-14
chore: generate: improve info logging

katcipis:
# Reason for This Change

Add some initial info logging on generate to provide info on creation/change/deletion of files per stack.


i4ki (approved):
lgtm

PR #646, opened 2022-10-14
feat: add .tmskip

katcipis:
# Reason for This Change

So we can add root context configurations in some directory and then import it without processing the configuration twice. This is also useful to ignore vendored content (in the future we can add .tmskip on vendored dirs).


i4ki (approved):
lgtm

PR #648, opened 2022-10-17
feat: add globals extend syntax

i4ki:
This PR introduces the ability to extend global objects in the same level or in child scopes.

Example:

```hcl
# /parent.tm

globals {
    project = {
        name = "my awesome infrastructure"
        tags = {
            # project-level tags here
        }
    }
}
```

```hcl
# /teams/marketing/globals.tm

globals "project" "tags" {
    team = "marketing"
}
```


generate/genhcl/genhcl.go
```diff
@@ -24,9 +24,9 @@ import (
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/mineiros-io/terramate/config"
 	"github.com/mineiros-io/terramate/errors"
-	"github.com/mineiros-io/terramate/globals"
 	"github.com/mineiros-io/terramate/hcl"
 	"github.com/mineiros-io/terramate/hcl/ast"
+	mcty "github.com/mineiros-io/terramate/hcl/cty"
```
katcipis, line 29: really not a big deal (or a small one), but thinkin if tmcty would not be more clear

katcipis: as we discussed on a call, maybe "github.com/mineiros-io/terramate/hcl/runtime" and we would have runtime.Object etc. not sure, but using cty seems misguided since these are our own representations of types not cty (some functions help integrating, but it is not cty I think..it would even be cool to be decoupled as much as possible from cty).

hcl/ast/merged_block.go
```diff
@@ -39,31 +43,57 @@ type MergedBlock struct {
 	RawBlocks map[string]Blocks
 }
 
+// BlockType represents a block type.
+type BlockType string
+
+// LabelBlockType represents a labelled block type.
+type LabelBlockType struct {
+	Type   BlockType // Type of the block
+	Labels string    // Labels are comma separated.
```
katcipis, line 52:
```suggestion
	Labels string    // Labels are dot separated.
```

It seems that Labels are Joined by "." not ","

hcl/ast/merged_block.go
```diff
@@ -39,31 +43,57 @@ type MergedBlock struct {
...
 		Blocks:     make(map[string]*MergedBlock),
 		RawBlocks:  make(map[string]Blocks),
 	}
 }
 
+// NewLabelBlockType returns a new LabelBlockType.
+func NewLabelBlockType(typ string, labels []string) LabelBlockType {
+	return LabelBlockType{
+		Type:   BlockType(typ),
+		Labels: strings.Join(labels, "."),
```
katcipis, line 76: maybe we could extract a newLabels that does the joining, the "." as the separator is duplicated here.

hcl/ast/merged_block.go
```diff
@@ -39,31 +43,57 @@ type MergedBlock struct {
...
+func (mb *MergedBlock) MergeBlock(other *Block, isLabelled bool) error {
 	errs := errors.L()
-
-	// Currently all merged blocks do not support labels.
-	// This should not be handled here if changed in the future.
-	if len(other.Labels) > 0 {
+	if !isLabelled && len(other.Labels) > 0 {
 		errs.Append(errors.E(other.LabelRanges, "block type %q does not support labels"))
+	} else {
+		otherLabels := strings.Join(other.Labels, ".")
```
katcipis, line 86: label building is also here, so maybe the extract could be used here too if it happens.

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+	//   {
+	//       val = {}
+	//   }
+	//
+	// Then values can be set inside obj.val by doing:
+	//
+	//   obj.SetAt("val.test", 1)
+	Object struct {
+		// Keys is a map of key names to values.
+		Keys map[string]interface{}
```
katcipis, line 48:
```suggestion
		Keys map[string]any
```

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// DotPath represents a path inside the object using a dot-notation.
+	DotPath string
+)
+
+// NewObject creates a new object.
+func NewObject() *Object {
+	return &Object{
+		Keys: make(map[string]interface{}),
```
katcipis, line 65:
```suggestion
		Keys: make(map[string]any),
```

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+}
+
+// Set a key value into object.
+func (obj *Object) Set(key string, value interface{}) {
```
katcipis, line 70:
```suggestion
func (obj *Object) Set(key string, value any) {
```

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+			newobj := NewObject()
+			newobj.SetFrom(vvalue.Raw().AsValueMap())
+			value = newobj
+		}
+	}
+	obj.Keys[key] = value
+}
+
+// GetKeyPath retrieves the value at path.
+func (obj *Object) GetKeyPath(path DotPath) (interface{}, bool) {
```
katcipis, line 82:
```suggestion
func (obj *Object) GetKeyPath(path DotPath) (any, bool) {
```

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+	parts := strings.Split(string(path), ".")
```
katcipis, line 83: Definitely not important, but here it seems to be the common use case for the new function [strings.Cut](https://pkg.go.dev/strings#Cut). It would help us avoid the re-joining.

hcl/cty/object.go
```diff
@@ -0,0 +1,174 @@
...
+			subtree.SetFrom(v.AsValueMap())
+			obj.Set(k, subtree)
+		} else {
+			obj.Set(k, v)
+		}
+	}
+}
+
+// SetAt sets a value at the specified path key.
+func (obj *Object) SetAt(path DotPath, value interface{}) error {
```
katcipis, line 116:
```suggestion
func (obj *Object) SetAt(path DotPath, value any) error {
```

hcl/cty/object_test.go
```diff
@@ -0,0 +1,181 @@
...
+			want: &cty.Object{
+				Keys: map[string]interface{}{
+					"key": &cty.Object{
+						Keys: map[string]interface{}{
+							"test": "child value",
+						},
+					},
+				},
+			},
+		},
```
katcipis, line 97:
```suggestion
		{
			name: "set at an existing child object",
			obj: &cty.Object{
				Keys: map[string]any{
					"key": cty.NewObject(),
				},
			},
			path: "key.test",
			val:  "child value",
			want: &cty.Object{
				Keys: map[string]any{
					"key": &cty.Object{
						Keys: map[string]any{
							"test": "child value",
						},
					},
				},
			},
		},
```

Can also do in other tests, it gets a little shorter.

katcipis (dismissed):
:rocket: :rocket: :rocket:

katcipis (approved):

katcipis (approved):

PR #651, opened 2022-10-18
chore: improve docs for errors pkg

katcipis:
# Reason for This Change

Add some extra information + format the docs according to some new features on Go 1.19 for lists/links. It should still work well for older Go versions too.


i4ki (approved):
lgtm

PR #652, opened 2022-10-18
chore: avoid loading entire os.Environ on tests

katcipis:
# Reason for This Change

Some tests could fail depending on which env vars the host has exported. We only need PATH since we use git. This should make the tests less dependent on the env.


i4ki (approved):
lgtm

PR #654, opened 2022-10-19
feat: add support to assert blocks inside generate_* blocks

katcipis:
# Reason for This Change

So we can have locally scoped assert blocks that can access the generate_hcl/generate_file lets block. They have the same behavior as the root based assert.

## Description of Changes

* Add support for it on parser
* Add eval support on genhcl
* Add eval support on genfile
* Integrate on generate


generate/genhcl/genhcl.go
```diff
@@ -181,66 +184,76 @@ func Load(
...
+			origin:    origin,
 			body:      formatted,
 			condition: condition,
+			asserts:   asserts,
 		})
 	}
 
-	sort.Slice(hcls, func(i, j int) bool {
-		return hcls[i].String() < hcls[j].String()
+	sort.SliceStable(hcls, func(i, j int) bool {
```
katcipis, line 249: Changed to just using the label, I think we started using String when it was simpler. Currently it would also use condition/body/etc on the ordering and that seemed odd. I think just ordering by label ir organized enough for testing etc. Since labels can be repeated I used a stable sort to keep the origina order things are processed.

hcl/hcl.go
```diff
@@ -719,34 +727,51 @@ func parseGenerateHCLBlock(block *ast.Block) (GenHCLBlock, error) {
...
-		switch b.Type {
+	errs := errors.L()
+
+	for _, subBlock := range block.Blocks {
+		switch subBlock.Type {
 		case "lets":
-			lets = append(lets, b)
+			lets = append(lets, subBlock.Block)
+		case "assert":
+			// TODO(KATCIPIS): error handling tests
```
i4ki, line 745: if we miss tests, isn't it better to be a TODO in the hcl_test.go instead?

hcl/hcl.go
```diff
@@ -719,34 +727,51 @@ func parseGenerateHCLBlock(block *ast.Block) (GenHCLBlock, error) {
...
-				)
+				))
+				continue
 			}
-			content = b
+			content = subBlock.Block
 		default:
 			// already validated but sanity checks...
-			panic("unreachable")
+			panic(errors.E("terramate internal error: unexpected block type %s", subBlock.Type))
```
i4ki, line 762: nice, maybe we can standardize this for internal panics by creating an `errors.Kind` of "terramate internal error".

generate/genfile/assert_test.go
```diff
@@ -0,0 +1,196 @@
...
+			want: []result{
+				{
+					name: "asserts.hcl",
+					file: genFile{
+						origin:    "/stack/generate.tm",
+						condition: true,
+						body:      "value",
+						asserts: []config.Assert{
+							{
+								Range:     Mkrange("/stack/generate.tm", Start(7, 17, 88), End(7, 34, 105)),
```
i4ki, line 70: maybe put the filename in a `const` (or use a helper func) and reuse here to avoid repetition, but feel free to  ignore.

generate/genhcl/genhcl.go
```diff
@@ -181,66 +184,75 @@ func Load(
...
-			logger.Error().
-				Err(err).
-				Stringer("origin", loadedHCL.origin).
-				Str("code", string(gen.Bytes())).
-				Str("label", name).
-				Msg("internal error formatting generated code")
-
 			panic(errors.E(sm, err,
-				"internal error: formatting generated code for generate_hcl %q", name,
+				"internal error: formatting generated code for generate_hcl %q:%s", name, string(gen.Bytes()),
```
i4ki, line 236: here we have a different message, maybe it would be nice to be:

```go
panic(errors.E(errors.ErrInternal, sm, err, "formatting generated code for generate_hcl %q:%s", name, string(gen.Bytes()))
```

i4ki (approved):
lgtm

hcl/hcl.go
katcipis: Ah this TODO is a leftover, sorry and thanks for the catch :+1:

i4ki (approved):

PR #655, opened 2022-10-19
feat: create labeled globals if no attribute is defined

i4ki:


globals/globals.go
```diff
@@ -91,6 +91,19 @@ func LoadExprs(tree *config.Tree, cfgdir project.Path) (Exprs, error) {
 
 	globalsBlocks := cfg.Node.Globals.AsList()
 	for _, block := range globalsBlocks {
+		if block.Labels != "" && len(block.Attributes.SortedList()) == 0 {
+			expr, _ := eval.ParseExpressionBytes([]byte(`{}`))
```
katcipis, line 95: I imagine there is no way to create an empty object and then get it as an expression :laughing:

katcipis (dismissed):

katcipis (approved):
:rocket: :rocket: :rocket:

PR #656, opened 2022-10-21
feat: add --quiet

katcipis:
# Reason for This Change

Running terramate with --quiet will silence terramate output on stdout.


i4ki (approved):
lgtm

PR #608, opened 2022-09-24
feat: add windows support

i4ki:
- [x] add windows to the GHA.
- [x] Portable Makefiles #601 
- [x] Make project paths OS-independent #610 
- [x] Fix test runner terminal emulator #609 
- [x] Fix file URI in the modvendor tests #611 
- [x] Make all tests portable (avoid the use of `cat`, `pwd`, etc) #612 

Tests progress:

- [x] errors
- [x] git (skipped for now)
- [x] project
- [x] run
- [x] run/dag
- [x] stack (fixed by #610)
- [x] tf (fixed by #610)
- [x] config
- [x] globals
- [x] hcl (fixed by #610)
- [x] hcl/eval (fixed by #610)
- [x] modvendor (fixed by #611)
- [x] generate (fixed by #610)
- [x] generate/genfile (fixed by #610)
- [x] generate/genhcl (fixed by #610)
- [x] lets
- [x] fs
- [x] hcl/ast
- [x] test
- [x] cmd/terramate/cli
- [x] cmd/terramate/e2etests (fixed by #612)
- [x] cmd/terramate/e2etests/cmd (fixed by #612)

PR #662, opened 2022-10-23
fix: stack.S doesn't implement errors.StackMeta

i4ki:
Running `terramate run` gave me this error messages:

```
2022-10-23T15:23:58+01:00 ERR running /stack%!(EXTRA *exec.Cmd=/usr/bin/make): exit status 2
2022-10-23T15:23:58+01:00 FTL one or more commands failed
```

The reason was that #608  broke the assumption that `stack.S` implements `errors.StackMeta`.

This line was added to ensure this doesn't happen again: https://github.com/mineiros-io/terramate/pull/662/files#diff-3255e7799f07c4f38eb7328b59a25c5a9d25b9661dcf83b1ea9a817d2bde6629R107

### DIFF
--- a/errors/errlog/errlog.go
+++ b/errors/errlog/errlog.go
@@ -79,7 +79,7 @@ func logerr(
 		ctx = ctx.Stringer("file", tmerr.FileRange)
 	}
 	if tmerr.Stack != nil {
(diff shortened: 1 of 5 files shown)


soerenmartius (approved):
lgtm

PR #663, opened 2022-10-24
refactor: introduce info.Range and use it instead of Origin

katcipis:
# Reason for This Change

We need more information about parsed blocks than a single origin filename. So instead of having a single origin string we will have our own ast.Range type that contains both paths, host and project path, and all the range information. 

Then a feature listing the origin of each generated code should be fairly simple.

It required a lot of changes because I also changed the approach to testing range information. Tried to make it as easy as it is to test as the hcl.Range tests on errors and now we don't test this everywhere but only on tests specific to range information.

We still use hcl.Range on the errors pkg for a lot of reasons we discussed in private. Improving that would make the changes much bigger (and it is already quite big). So the scope is the minimum to enable the feature.


run/env.go
```diff
@@ -83,19 +83,17 @@ func LoadEnv(cfg *config.Tree, projmeta project.Metadata, st *stack.S) (EnvVars,
 
 		val, err := evalctx.Eval(attr.Expr)
 		if err != nil {
-			return nil, errors.E(ErrEval, attr.NameRange,
-				err, "attribute origin %s", attr.Origin)
+			return nil, errors.E(ErrEval, err)
```
katcipis, line 86: Seemed to be best to use the range provided by Eval.

hcl/info/range_test.go
```diff
@@ -0,0 +1,81 @@
...
+	end := End(0, 0, 0)
+	hclrange := Mkrange(filepath.Join(rootdir, path), start, end)
+	tmrange := info.NewRange(rootdir, hclrange)
+
+	assert.EqualStrings(t, hclrange.Filename, tmrange.HostPath())
+	assertRangePath(t, tmrange, path)
+}
+
+func TestRangeOnRootWithFileOnRootdir(t *testing.T) {
+	rootdir := string(os.PathSeparator)
```
katcipis, line 63: Not sure this is the best to test root / on windows. Let me know if we can improve this (didn't wanted to have a specific test for windows..but maybe that is necessary).

hcl/info/range.go
```diff
@@ -0,0 +1,109 @@
...
+// counted starting at 1 and incremented for each newline character
+// encountered.
+func (p Pos) Line() int {
+	return p.line
+}
+
+// Column is the source code column where this position points, in
+// unicode characters, with counting starting at 1.
+//
+// Column counts characters as they appear visually, so for example a
```
i4ki, line 93: hashicorp docs?

hcl/info/range_test.go
i4ki: I think this work fine on recent windowezsss

i4ki (approved):
LGTM

hcl/info/range.go
katcipis: yup, since I just copied the same values I imagined we retain the same semantics/docs/etc

PR #666, opened 2022-10-25
chore: document globals extend feature

i4ki:


docs/sharing-data.md
```diff
@@ -23,10 +23,56 @@ add a **globals** block to your [Terramate configuration file](config-overview.m
...
+
+```hcl
+{
+  value = "string"
+  number = 10
+  child = {
+    grandchild = {
+      list = []
+    }
+  }
```
katcipis, line 71: maybe add something to make it clear that his extensions behavior also applies on the hierarchy, that globals inherited from the parent can be extended (disregard if I missed this info).

katcipis (approved):

PR #669, opened 2022-10-25
fix: outdated code detection not handling condition

katcipis:
# Reason for This Change

Outdated code detection was not handling multiple blocks with different conditions properly. It was working before because of details on how the blocks were ordered that were changed. Made the code be not dependent on the ordering since ordering is mostly used to test deterministically and to output results nicelly.

### DIFF
--- a/generate/generate_hcl_check_test.go
+++ b/generate/generate_hcl_check_test.go
@@ -369,15 +369,10 @@ func TestCheckOutdatedIgnoresWhenGenHCLConditionIsFalse(t *testing.T) {
 	stackEntry := s.CreateStack("stacks/stack")
 	stack := stackEntry.Load(s.Config())
 
-	changedStacks := false
-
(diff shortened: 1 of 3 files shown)


generate/generate_hcl_check_test.go
```diff
@@ -369,15 +369,10 @@ func TestCheckOutdatedIgnoresWhenGenHCLConditionIsFalse(t *testing.T) {
 	stackEntry := s.CreateStack("stacks/stack")
 	stack := stackEntry.Load(s.Config())
 
-	changedStacks := false
```
i4ki, line 372: I did this to only reload if there were changes, so tests are faster. Does it has any problem?

i4ki (approved):
lgtm

generate/generate_hcl_check_test.go
katcipis: It was to avoid me making mistakes =P

PR #672, opened 2022-10-26
refactor: improve outdated detection

katcipis:
# Reason for This Change

Mostly improving the tests. The old tests were really hard to extend and adding new scenarios was too cumbersome. The new tests try to make it easy to add new scenarios, modeling a set of steps, so we can express complex scenarios like the configurations changing in a project.

The tests are also all together now instead of separated by generate block, so we can focus on the outdated behavior, with stacks or outside stacks, with all generate blocks, etc, all together and integrated.


i4ki (approved):
lgtm

PR #673, opened 2022-10-27
chore: terramate create output improvements

katcipis:
This PR reduces verbosity of `terramate create`. It also introduces a minimal version of the code generation report, that then is used on `terramate create` when `-v` is passed.

Default output:

```sh
terramate create stack-4
Created stack stack-4
```

Verbose:

```sh
terramate create -v stack-5
Created stack stack-5
Created /stack-5/_module.tf
```


i4ki (approved):
lgtm

PR #674, opened 2022-10-27
feat: add terramate experimental generate debug

katcipis:
Add new `terramate experimental generate debug` command to aid debugging code generation. It shows all generated blocks that have a true condition and their respective origin. It integrates with --changed, so when running --changed only changed stacks will be on the output.

Example:

```sh
#!/bin/bash

set -o errexit
set -o nounset

project="$(mktemp -d)"
cd "${project}"

cat > terramate.tm <<- EOM
terramate {
  config {
  }
}
EOM

terramate create stack-1
terramate create stack-2

cat > config.tm <<- EOM
generate_hcl "test.hcl" {
    content {
        a = "data"
    }
}
generate_hcl "test.hcl" {
    condition = false
    content {
        a = "data"
    }
}
generate_file "file.txt" {
    content = "data"
}
EOM

mkdir modules

cat > modules/config.tm <<- EOM
generate_hcl "imported.hcl" {
    content {
        a = "data"
    }
}
EOM

cat > stack-1/config.tm <<- EOM
import {
  source = "../modules/config.tm"
}
EOM

echo
echo "running on root"
terramate experimental generate debug

echo
echo "running on stack-1"
cd stack-1
terramate experimental generate debug

echo
echo "running on stack-2"
cd ../stack-2
terramate experimental generate debug
```

Will generate this output:

```sh
running on root
Generated files for /stack-1:
        - file.txt origin: /config.tm:12,1-14,2
        - imported.hcl origin: /modules/config.tm:1,1-5,2
        - test.hcl origin: /config.tm:1,1-5,2
Generated files for /stack-2:
        - file.txt origin: /config.tm:12,1-14,2
        - test.hcl origin: /config.tm:1,1-5,2

running on stack-1
Generated files for /stack-1:
        - file.txt origin: /config.tm:12,1-14,2
        - imported.hcl origin: /modules/config.tm:1,1-5,2
        - test.hcl origin: /config.tm:1,1-5,2

running on stack-2
Generated files for /stack-2:
        - file.txt origin: /config.tm:12,1-14,2
        - test.hcl origin: /config.tm:1,1-5,2
```


cmd/terramate/e2etests/exp_generate_test.go
```diff
@@ -0,0 +1,233 @@
...
+							Bool("condition", true),
+							Content(
+								Str("content", "data"),
+							),
+						),
+					),
+				},
+			},
+			want: runExpected{
+				Stdout: `Generated files for /stack-1:
```
i4ki, line 103: as this is an experimental feature (and we are in the process of changing outputs of some commands), testing against stdout could be a problem.
Don't need to change, but if a request to change the output comes later, maybe makes sense to just test the data structure (non e2e) and then maybe re-add this test once it becomes a non-exp feature.

i4ki (approved):
lgtm

cmd/terramate/e2etests/exp_generate_test.go
katcipis: The in structure tests already exist, it would be the load_test, because all the info/logic is there. This was an extra just to ensure the output is sane, but indeed changing it later will be more work.

PR #675, opened 2022-10-28
tool-versions should match go.mod major version

lsc:
# Reason for This Change

Major versions should match so we don't confuse developers or break local builds.


katcipis (approved):

katcipis:
Hmm the option to allow the CI to run is not showing anymore, and then the CI is stuck :thinking:

katcipis:
Created another PR for this (#685). Not sure why the CI is never running here. Thanks for pointing this out @lsc

PR #681, opened 2022-10-31
fix: make check_remote safeguard run from any branch

i4ki:


cmd/terramate/e2etests/general_test.go
```diff
@@ -312,13 +312,30 @@ func TestFailsOnChangeDetectionIfCurrentBranchIsMainAndItIsOutdated(t *testing.T
...
+		), wantRes)
+	}
+
+	testrun()
+
+	git.CheckoutNew("branch")
+	stack.CreateFile("tempfile2", "any content")
+	git.CommitAll("add tempfile2")
+
+	testrun()
```
katcipis, line 338: hmm so now we always check with and without --changed + we are also checking the detection if you are not on an outdated main, but on an outdated branch.

cmd/terramate/cli/project.go
```diff
@@ -231,10 +231,6 @@ func (p *project) checkLocalDefaultIsUpdated() error {
 		Str("workingDir", p.wd).
 		Logger()
 
-	if !p.isDefaultBranch() {
```
katcipis, line 234: hmm so now the function is not exactly "checkLocalDefaultIsUpdated" no ? maybe "checkBranchIsUpdated" ?

i4ki: current code checks if HEAD and remote origin/main has the same ancestor commit, or in other words, it checks if remote origin/main commit is reachable from HEAD.
So the branch name is not important at all, but I agree we should improve the naming here.

But the current name `checkLocalDefaultIsUpdated()` is not technically wrong, because the code fetches the remote origin/main commit and verify if HEAD can reach it, if not, it's because the LocalDefaultBranch is not updated...

katcipis (approved):

cmd/terramate/e2etests/general_test.go
i4ki: yeah but the branch you are is not important anymore, we only check if local origin/main is updated.

katcipis: thanks for clarifying :heart:

katcipis (approved):

PR #683, opened 2022-10-31
chore: improve create output

katcipis:
# Reason for This Change

Errors should have more context now and look like this:

```
terramate create stack-3
2022-10-31T19:15:02+01:00 FTL can't create stack: stack already exists stack=/stack-3
```

```
terramate create stack-1
2022-10-31T19:15:17+01:00 FTL can't create stack: default cfg file already exists file=stack.tm.hcl stack=/stack-1
```


stack/create.go
```diff
@@ -36,6 +36,10 @@ const (
 
 	// ErrStackAlreadyExists indicates that the stack already exists and cant be created.
 	ErrStackAlreadyExists errors.Kind = "stack already exists"
+
+	// ErrStackDefaultCfgFound indicates that the dir already has a default
+	// stack configuration.
+	ErrStackDefaultCfgFound errors.Kind = "default cfg file already exists"
```
i4ki, line 42:
```suggestion
	ErrStackDefaultCfgFound errors.Kind = "default configuration file for stack already exists"
```
not sure if better, just trying to make it clear that this is the default file we use for stacks...

i4ki (dismissed):
lgtm

i4ki (approved):

PR #687, opened 2022-11-01
feat: add --ignore-existing on terramate create

katcipis:
# Reason for This Change

Allow `terramate create --ignore-existing` which then will not fail if the stack already exists (but won't do any changes on the stack).

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -724,7 +724,8 @@ func (c *cli) createStack() {
 	})
 
 	if err != nil {
-		if c.parsedArgs.Create.IgnoreExisting {
+		if c.parsedArgs.Create.IgnoreExisting &&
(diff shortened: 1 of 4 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -723,6 +724,10 @@ func (c *cli) createStack() {
 	})
 
 	if err != nil {
+		if c.parsedArgs.Create.IgnoreExisting &&
+			errors.IsKind(err, stack.ErrStackAlreadyExists) {
+			return
```
i4ki, line 729: maybe we should log something about this? so the user knows that the file was not copied?

i4ki (approved):

cmd/terramate/cli/cli.go
katcipis: Makes sense, thanks :heart:

i4ki (approved):

Issue #692, opened 2022-11-03
[BUG] Issue with metadata interpretation in stack configuration


joshuacook-tamu:
**Describe the bug**
I am attempting to define orchestration / dependency relationships between stacks. I defined it via the `after` statement in the stack config and used the metadata such as `${terramate.stack.path.to_root}`. Terramate fails to evaluate stating there is no variable named "terramate"

**To Reproduce**

1. Define a stack as

```
 stack {
  name = "test vm"
  after = [
    "${terramate.root.path.fs.absolute}/stacks/resources/vm-admin-keypair",
    #"../../resources/vm-admin-keypair",
    "${terramate.stack.path.to_root}/resources/vm-security-group",
    #"../../resources/vm-security-group",
  ]
 }
```

Note, I tried two different approaches above, but both fail with the same error.

2. Run command `terramate list`
3. See error
```
terramate list
2022-11-03T11:50:18-05:00 ERR terramate schema error: loading from /redacted/path/to/stacks: loading from /redacted/path/to/stacks/staging: loading from /redacted/path/to/stacks/staging/redacted: eval expression: failed to evaluate "after" attribute: There is no variable named "terramate". file=/redacted/path/to/stacks/staging/redacted/stack.tm.hcl:4,8-17
2022-11-03T11:50:18-05:00 ERR terramate schema error: loading from /redacted/path/to/stacks: loading from /redacted/path/to/stacks/staging: loading from /redacted/path/to/stacks/staging/redacted: eval expression: failed to evaluate "after" attribute: There is no variable named "terramate". file=/redacted/path/to/stacks/staging/redacted/stack.tm.hcl:6,8-17
2022-11-03T11:50:18-05:00 FTL looking up project root
```

**Expected behavior**

I expected the variable interpolation to work here as it does in multiple other configuration elements within Terramate. Of course, it will work if I uncomment the lines above with the `../..` and comment out the lines with the Terramate metadata elements. This is what I have done for now.

**Log Output**
Add logs from Terramate to help debug your problem.

**Environment (please complete the following information):**
 - OS: MacOS 
 - OS Version: 12.6.1
 - Git Version: 2.37.1
 - Terramate Version: 0.1.40

mariux:
Hi @joshuacook-tamu,

thanks for sharing. The problem is that `terramate` namepace is not available in `after`/`before` as of today. 

We will for sure improve the error message in this case ;) We are actually currently refactoring all output to improve the user experience.

[...]

PR #693, opened 2022-11-04
chore: improve stack docs

katcipis:
Document the stack block better, should avoid confusion as what happened at #692


soerenmartius (approved):
lgtm

PR #702, opened 2022-11-08
feat: allow eval context to be extended with tm_vendor function

katcipis:
# Changes

Add a new extended context that has the `tm_vendor` function and the function itself, with some tests to check that both the new extended context + the function work properly. This PR only adds the new eval context and the function but it is not used yet on the rest of the code, so it introduces no change on behavior. The next step will be to integrate NewExtContext everywhere we need tm_vendor to be available.

# Solution

To avoid circular dependencies when including modvendor related logic into hcl/eval (to calculate the paths) I also refactored the modvendor pkg, creating 2 new more specific pkgs with most of the logic/dependencies while modvendor only has very basic definitions, like the TargetDir, which is then used by the tm_vendor function.

Also extracted all events into the event pkg, so consumer/producers don't need to know each other directly, which can easily end up with annoying circular dependency issues or odd imports, like modvendor/download importing hcl/eval.


mariux:
test https://app.shortcut.com/mineiros/story/6564/feat-add-tm-vendor

hcl/eval/functions.go
```diff
@@ -67,6 +72,52 @@ func tmAbspath(basedir string) function.Function {
...
+		Type: function.StaticReturnType(cty.String),
+		Impl: func(args []cty.Value, retType cty.Type) (cty.Value, error) {
+			// Param spec already enforce modsrc to be string.
+			source := args[0].AsString()
+			modsrc, err := tf.ParseSource(source)
+			if err != nil {
+				return cty.NilVal, errors.E(err, "tm_vendor: invalid module source")
+			}
+			targetPath := modvendor.TargetDir(vendordir, modsrc)
+			result, err := filepath.Rel(basedir.String(), targetPath.String())
```
i4ki, line 92: both paths are UNIX paths then this should not be a failing operation but I didn't find a `path.Rel()`, unfortunately, then panic is our only option =(

modvendor/download/download.go
```diff
@@ -0,0 +1,503 @@
...
+func vendor(
+	rootdir string,
+	vendorDir project.Path,
+	modsrc tf.Source,
+	report Report,
+	info *modinfo,
+	events EventStream,
+) Report {
+	logger := log.With().
+		Str("action", "modvendor.vendor()").
```
i4ki, line 113:
```suggestion
		Str("action", "download.vendor()").
```

modvendor/download/download.go
```diff
@@ -0,0 +1,503 @@
...
+
+func vendorAll(
+	rootdir string,
+	vendorDir project.Path,
+	tfdir string,
+	report Report,
+	events EventStream,
+) Report {
+	logger := log.With().
+		Str("action", "modvendor.vendorAll()").
```
i4ki, line 190:
```suggestion
		Str("action", "download.vendorAll()").
```

modvendor/download/download.go
```diff
@@ -0,0 +1,503 @@
...
+// This function is not recursive, so dependencies won't have their dependencies
+// vendored. See Vendor() for a recursive vendoring function.
+func downloadVendor(
+	rootdir string,
+	vendorDir project.Path,
+	modsrc tf.Source,
+	events EventStream,
+) (string, error) {
+	logger := log.With().
+		Str("action", "modvendor.downloadVendor()").
```
i4ki, line 297:
```suggestion
		Str("action", "download.downloadVendor()").
```

modvendor/download/download.go
```diff
@@ -0,0 +1,503 @@
...
+		// Assuming that the whole Terramate project is inside the
+		// same fs/mount/dev.
+		return "", errors.E(err, "moving module from tmp dir to vendor")
+	}
+	return modVendorDir, nil
+}
+
+func patchFiles(rootdir string, files []string, sources *sourcesInfo) error {
+	logger := log.With().
+		Str("action", "modvendor.patchFiles").
```
i4ki, line 436:
```suggestion
		Str("action", "download.patchFiles").
```

modvendor/download/download.go
```diff
@@ -0,0 +1,503 @@
...
+			}
+			source := block.Body().GetAttribute("source")
+			if source == nil {
+				continue
+			}
+
+			sourceString := string(source.Expr().BuildTokens(nil).Bytes())
+			sourceString = strings.TrimSpace(sourceString)
+			sourceString = sourceString[1 : len(sourceString)-1] // unquote
+			// TODO(i4k): improve to support parenthesis.
```
i4ki, line 476: hahaha oh no you are exposing me

i4ki (approved):
lgtm

modvendor/download/download.go
katcipis: nice catch, thanks :100:

i4ki (approved):

PR #705, opened 2022-11-10
feat: add support for generating files outside stacks

i4ki:
Add support for `generate_file` block with `context=root` attribute (the default being `stack`) that allows for generating files outside of stacks.

When using `context=root` the limitations below apply:

- label must be an absolute path
- label must not point to a path inside a stack
- globals are not available
- stack metadata is not available

The only metadata available is the [project metadata](https://github.com/mineiros-io/terramate/blob/main/docs/sharing-data.md#project-metadata).

The `generate_file` with `context=root` can be defined anywhere, even inside stacks, but the file cannot be generated inside a stack (for that case you should use `context=stack`).

Example below will generate a file called `/some/dir/msg.txt` with content `Hello World from Terramate`.

```
generate_file "/some/dir/msg.txt" {
    context = root
    lets {
        name = "Terramate"
        msg = "hello world"
    }
    
    content = "${tm_title(let.msg)} from ${let.name}"
}
```


config/config.go
```diff
@@ -169,6 +174,18 @@ func (tree *Tree) Lookup(abspath project.Path) (*Tree, bool) {
 	return cfg, true
 }
 
+// Dirs returns all the configuration directories that have configurations.
```
katcipis, line 177: There is some stuttering here, thinking if we can't just assume that subdirs/dirs inside the tree always have configurations, because our tree is a config tree already right ? unless we do have some dirs sometimes that have no config at all, but if it represents only configuration it is already an invariant that anything inside has configs and is config related etc (it is not a fs tree). But I'm not sure, just thinking out loud (also not fundamental or anything).

If what Im saying is not completely senseless, maybe this could be just:

```suggestion
// Dirs returns a list with this configuration and all its children.
```

Now not so sure about calling this Dirs...you are actually getting the whole tree as a list no ? Maybe Collect ? Then it would be: "Collect returns the entire tree collected as a depth first list".

Sorry if this is all stupid, feel free to disregard.

generate/generate.go
```diff
@@ -115,6 +115,34 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
 		res.Files = generated
 		results[i] = res
 	}
+
+	for _, dircfg := range cfg.Root().Dirs() {
+		res := LoadResult{Dir: dircfg.ProjDir()}
+		evalctx, err := eval.NewContext(dircfg.Dir())
```
katcipis, line 121: The suggestion is not complete, just me thinking out loud. Feel free to ignore, just getting my head around the dirs/nodes/tree/etc abstractions.

```suggestion
	for _, node := range cfg.Root().Collect() {
		res := LoadResult{Dir: node.ProjDir()}
		evalctx, err := eval.NewContext(node.Dir())
```

Or instead of node just redefine `cfg`. Not sure it is better, just tossing the idea around.

generate/generate.go
```diff
@@ -115,6 +115,34 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+			if err != nil {
+				res.Err = errors.L(res.Err, err).AsError()
+				results = append(results, res)
+				continue
+			}
+
+			generated = append(generated, file)
+			res.Files = generated
+			results = append(results, res)
+		}
```
katcipis, line 144:
```suggestion
			
		}
		res.Files = generated
		results = append(results, res)
```

Unless we want to have a result with 1 file, then another result with 2 files, etc (for N generate_file blocks).

generate/generate.go
```diff
@@ -135,150 +163,300 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
-	) dirReport {
-		stackpath := stack.HostPath()
-		logger := log.With().
-			Str("action", "generate.Do()").
-			Stringer("stack", stack.Path()).
-			Logger()
+	stackReport := forEachStack(cfg, workingDir, doStackGeneration)
+	dirReport := forEachDir(cfg, doDirGeneration)
+	report := mergeReports(stackReport, dirReport)
+	return cleanupOrphaned(cfg, report)
```
katcipis, line 169: Nice :100:

generate/generate.go
```diff
@@ -135,150 +163,300 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+	failureReport := func(r dirReport, err error) dirReport {
+		r.err = err
+		for filename := range removedFiles {
+			r.addDeletedFile(filename)
 		}
+		return r
+	}
 
-		var removedFiles map[string]string
+	removedFiles, err = removeStackGeneratedFiles(cfg, stack.HostPath(), generated)
```
katcipis, line 236: Sorry for the remove all files hack, hopefully we can improve this soon

generate/generate.go
```diff
@@ -135,150 +163,300 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+		return report
+	}
 
-		err = validateGeneratedFiles(cfg, stackpath, generated)
-		if err != nil {
-			report.err = err
-			return report
+	var removedFiles map[string]string
+
+	failureReport := func(r dirReport, err error) dirReport {
```
katcipis, line 228: this failureReport thing would not be necessary without the remove all files ugly hack, sorry for this x_x

generate/generate.go
```diff
@@ -135,150 +163,300 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+	for _, file := range files {
+		dirReport := dirReport{}
+		target := filepath.Join(cfg.RootDir(), file.Label())
+		basedir := project.NewPath(path.Dir(file.Label()))
+		filename := filepath.Base(file.Label())
+		logger := logger.With().
+			Str("filename", target).
+			Logger()
+
+		if !file.Condition() {
```
katcipis, line 380: Will this work if you have one generate_file with condition = false and another one with condition = true ? On the stack generation I remove all files, manipulate the map and report deleted files only in the end because of the multiple conditions case. It seems that here if I have one condition = true and one = false I will always report it was deleted (maybe even report the same file was deleted and then created ?).

Also multiple blocks with the same label are allowed when they have condition = false, and it seems this code would add the file multiple times in this case.

But I could be missing something.

generate/generate.go
```diff
@@ -623,7 +801,45 @@ func readFile(path string) (string, bool, error) {
...
+	}
+
+	report := Report{}
+	projmeta := project.NewMetadata(root.RootDir(), stackpaths)
+	evalctx, err := eval.NewContext(cfg.Dir())
+	evalctx.SetNamespace("terramate", projmeta.ToCtyMap())
+	if err != nil {
+		report.BootstrapErr = err
+		return report
+	}
```
katcipis, line 831:
```suggestion
	evalctx, err := eval.NewContext(cfg.Dir())
	if err != nil {
		report.BootstrapErr = err
		return report
	}
	evalctx.SetNamespace("terramate", projmeta.ToCtyMap())
```

generate/generate.go
```diff
@@ -732,6 +948,51 @@ func removeStackGeneratedFiles(
...
+	cfg *config.Tree,
+	genfiles []GenFile,
+) (map[string]string, map[string]error) {
+	logger := log.With().
+		Str("action", "generate.removeRootGeneratedFiles()").
+		Str("root", cfg.RootDir()).
+		Logger()
+
+	errs := make(map[string]error)
+	removedFiles := map[string]string{}
```
katcipis, line 961:
```suggestion
	errs := map[string]error{}
	removedFiles := map[string]string{}
```

config/config.go
i4ki: The docs in stupid, thanks for catching this.

> There is some stuttering here, thinking if we can't just assume that subdirs/dirs inside the tree always have configurations, because our tree is a config tree already right ? 

Currently the `config.Tree` maintains the same filesystem structure of the project, then directories with no configuration are kept in the tree, then `node.Parent` points to the parent directory (if any).

But if we change to ignore the directories with no configuration in the tree, then maybe with the gaps it will not be enough to express all use cases. 
Eg.:
```
s:stacks
s:stacks/orgs/mineiros
s:stacks/orgs/google
s:stacks/orgs/meta
```
The `stacks/orgs` directory has no configuration, then in this case, it will not be possible to call `generate.Do(root.Lookup("/stacks/orgs"))` to execute generate from that directory.

That's why (I think) we should keep the configuration tree resembling the filesystem tree, because sometimes we want to execute something using the *view* of some directory (eg: the working dir filter).

Each Tree is actually the directory, and it can hold a config or not... the `dir` is ommited in the Tree name but it's there. I just thought that calling it `config.Dir` would be funny.

But I agree calling this `Dirs()` but only returning dirs with config is not correct.
What about `cfg.AsList()`? and then it returns all dir configs and the client filters out by `!cfg.IsEmpty()`?

```suggestion
// AsList returns a list with this configuration and all its children.
```

katcipis: I think I was thinking on something like the skipdir feature, today because of skipdir we already don't load the whole fs tree, so it is not a perfect representation, but makes sense that a lot of nodes will be empty.

AsList should do the trick, I prefer Collect because AsX always reminds me of casting, it looks like you are just to cast something to a different type (like errors.As()), but it is something only on my head/subjective and totally ignorable.

generate/generate.go
i4ki: thanks, yeah this is not even tested yet

generate/generate_root_test.go
```diff
@@ -0,0 +1,332 @@
...
+				Successes: []generate.Result{
+					{
+						Dir:     "/target",
+						Created: []string{"empty.txt"},
+					},
+				},
+			},
+		},
+		{
+			name: "generate_file with false condition generates nothing",
```
katcipis, line 59: Maybe add some tests interleaving generate_file with condition = true + condition = false, to test that mixing them works properly (we have some tricky logic around that, so the tests could be beneficial).

generate/generate_root_test.go
```diff
@@ -0,0 +1,332 @@
...
+						Labels("/target/file.txt"),
+						Expr("context", "root"),
+						Bool("condition", false),
+						Str("content", "content"),
+					),
+				},
+			},
+		},
+		{
+			name: "generate.context=root has access to project metadata",
```
katcipis, line 73: As a corner case maybe it would be cool to test access to the stacks list when the project has no stacks at all.

generate/report.go
```diff
@@ -288,3 +311,25 @@ func (s dirReport) empty() bool {
 		len(s.deleted) == 0 &&
 		s.err == nil
 }
+
+type result interface {
+	Result | FailureResult
+}
+
+func joinResults[T result](results ...[]T) []T {
```
katcipis, line 319: Maybe just do [T any] and drop the constraint ? There is nothing on the algorithm that imposes a constraint on the type.

generate/generate.go
i4ki: there's a `checkFileConflict()` function called before this that checks if only one target has condition=true, otherwise it fails. The at this point it just saves again the files that do need to be created.

katcipis: hmm I'm not so sure this is related to file conflicts or even saving/deleting files, I'm thinking on the report itself, but Ill check the code again and will try to explain better my concerns (could be missing something here).

katcipis: I get it now, sorry for the noise =P

generate/generate_root_test.go
i4ki: done

generate/generate_root_test.go
i4ki: done

generate/generate.go
```diff
@@ -135,150 +172,280 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+	}
 
-		logger.Debug().Msg("finished generating files")
-		return report
-	})
+	logger.Debug().Msg("no conflicts found")
 
-	return cleanupOrphaned(cfg, report)
+	wd := project.PrjAbsPath(cfg.RootDir(), workingDir)
+	files = filterGenFilesByTargetDir(files, wd)
```
katcipis, line 396: Nice extracted filter function :100:

generate/generate.go
```diff
@@ -135,150 +172,280 @@ func Load(cfg *config.Tree) ([]LoadResult, error) {
...
+
+	generateRootFiles(cfg, files, &report)
+	return report
+}
+
+func filterGenFilesByTargetDir(files []GenFile, dir project.Path) []GenFile {
+	logger := log.With().
+		Str("action", "generate.filterGenFilesByTargetDir()").
+		Stringer("workingDir", dir).
```
katcipis, line 405:
```suggestion
		Stringer("targetDir", dir).
```

generate/generate.go
```diff
@@ -751,6 +923,149 @@ func removeStackGeneratedFiles(
...
+			logger := genFileLogger(logger, file)
+			logger.Debug().Msg("file must be generated (if needed)")
+
+			mustExistFiles[file.Label()] = file
+		}
+	}
+
+	// this computes the files that must be deleted if they are present on disk
+	for _, file := range genfiles {
+		if _, ok := mustExistFiles[file.Label()]; !ok {
```
katcipis, line 953: nice way to check if from N blocks at least one is true, cool :100:

generate/generate.go
```diff
@@ -751,6 +923,149 @@ func removeStackGeneratedFiles(
...
+	genfiles []GenFile,
+	report *Report,
+) {
+	logger := log.With().
+		Str("action", "generate.generateRootFiles()").
+		Logger()
+
+	diskFiles := map[string]string{}       // files already on disk
+	mustExistFiles := map[string]GenFile{} // files that must be present on disk
+	mustDeleteFiles := map[string]bool{}   // files to be deleted
```
katcipis, line 937: noice :100:

generate/generate.go
```diff
@@ -751,6 +923,149 @@ func removeStackGeneratedFiles(
...
+		logger.Debug().Msg("file content read successfully")
+
+		diskFiles[label] = string(body)
+	}
+
+	// this deletes the files that exist but have condition=false.
+	for label, mustDelete := range mustDeleteFiles {
+		logger := logger.With().Str("file", label).Logger()
+
+		if mustDelete {
```
katcipis, line 991: do we need a map to boolean + an if here ? It seems that mustDeleteFiles is just a set of files that must be deleted, why would any entry be on the map and not be supposed to be deleted ? So it could be just map[string]struct{}, and then here we could just iterate it and delete everything (unless we do have a few entries with bool = false, but I dont think that is the case).

generate/generate_stack_file_test.go
```diff
@@ -48,13 +48,13 @@ func TestGenerateFile(t *testing.T) {
 			},
```
katcipis, line 48: hmm we have generate_root_test and generate_stack_file_test, the naming doesn't seem consistent. Maybe it could be:

* generate_file_root_test.go
* generate_file_stack_test.go

Or:

* generate_root_file_test.go
* generate_stack_file_test.go

katcipis (dismissed):

generate/generate_stack_file_test.go
i4ki: true!
I changed to use the second option because then we can type less when using file completion to open them. Does it make sense?

generate/generate.go
i4ki: I thought it'd be clearer with the bool but yeah it makes sense to just use struct{}.
thanks

generate/generate_stack_file_test.go
katcipis: Totally :100:

katcipis (approved):
:rocket: :rocket: :rocket:

PR #706, opened 2022-11-11
fix: generate_hcl: when stack on root generated files get deleted on cleanup

katcipis:
# Reason for This Change

When the stack is on root we are creating/deleting the file. To reproduce:

```sh
#!/bin/bash

set -o errexit
set -o nounset

terramate {
  config {
  }
}
EOM

cat > config.tm.hcl <<- EOM
stack {
  name        = "terraform-google-service-account"
  description = "github.com/mineiros-io/terraform-google-service-account"
}

generate_hcl "versions.tf" {
  content {
  }
}
EOM

terramate generate

```

Outdated detection was also not working properly on this scenario.

## Description of Changes

Fix both outdated detection and generation with regression tests.


i4ki (approved):
LGTM

PR #707, opened 2022-11-11
chore: add handler of vendor request events on modvendor.download

katcipis:
# Reason for This Change

The handler will be used in order to vendor all tm_vendor calls while generating code. Integration will be done in a follow up PR.


i4ki (approved):
lgtm

PR #710, opened 2022-11-14
feat: generate_hcl: remove origin info from header

katcipis:
# Reason for This Change

We want to avoid triggering stacks on change detection when we just move generate blocks to different files.
This is a breaking change.

## Description of Changes

Remove the origin information from the header and update the outdated detection test.


i4ki (approved):
:rocket:

PR #715, opened 2022-11-15
feat: generate: add tm_vendor

katcipis:
```sh
#!/bin/bash

set -o errexit
set -o nounset
set -o pipefail

workdir="$(mktemp -d)"
cd "${workdir}"

cat > terramate.tm <<- EOM
terramate {
    config {
    }
}
EOM

terramate create stack

cat > vendor.tm <<- EOM
generate_hcl "vendor.hcl" {
    content {
        vendor = tm_vendor("github.com/mineiros-io/terramate?ref=main")
    }
}
generate_file "vendor.txt" {
    content = tm_vendor("github.com/mineiros-io/terramate?ref=main")
}
EOM

terramate generate

echo
echo "=== generated HCL ==="
cat stack/vendor.hcl

echo
echo "=== generated file ==="
cat stack/vendor.txt

echo
echo
echo "=== vendor dir ==="
tree -d -L 4 ./modules
```


i4ki:
I think we are missing info the vendor events. In the case it tries to vendor for each stack but the output looks like it's retrying.
WDYT?

[asciicast](https://asciinema.org/a/538635)

generate/generate_hcl_test.go
```diff
@@ -958,7 +959,7 @@ func TestWontOverwriteManuallyDefinedTerraform(t *testing.T) {
 		fmt.Sprintf("f:stack/%s:%s", genFilename, manualTfCode),
 	})
 
-	report := generate.Do(s.Config(), s.RootDir())
+	report := generate.Do(s.Config(), s.RootDir(), project.NewPath("/modules"), nil)
```
i4ki, line 962: we have this several times, maybe it could be a:
```
const TestVendorDir project.Path = "/modules"
```
inside the `generate_test.go`.

generate/generate_hcl_test.go
```diff
@@ -1312,7 +1313,7 @@ func TestGenerateHCLCleanupOldFilesIgnoreSymlinks(t *testing.T) {
 
 	cfg, err := config.LoadTree(rootEntry.Path(), rootEntry.Path())
 	assert.NoError(t, err)
-	report := s.GenerateAt(cfg, rootEntry.Path())
+	report := s.GenerateAt(cfg, rootEntry.Path(), project.NewPath("/modules"))
```
i4ki, line 1316: here

generate/generate_test.go
```diff
@@ -752,15 +754,19 @@ func testCodeGeneration(t *testing.T, tcases []testcase) {
 			}
 
 			workingDir := filepath.Join(s.RootDir(), tcase.workingDir)
-			report := generate.Do(s.Config(), workingDir)
+			vendorDir := project.NewPath("/modules")
```
i4ki, line 757: and here.

modvendor/download/report.go
```diff
@@ -109,6 +109,41 @@ func (r Report) Verbose() string {
...
+
+// IsEmpty returns true if the report is empty (nothing to report).
+func (r Report) IsEmpty() bool {
+	return len(r.Vendored) == 0 &&
+		len(r.Ignored) == 0 &&
+		r.Error == nil
+}
+
+func (r *Report) merge(other Report) {
+	if other.Error != nil {
```
i4ki, line 132: can't it be simplified using errors.List? like:
```
errs := errors.L(r.Error, other.Error)
r.Error = errs.AsError()
```
?
because the errors.L() and errs.Append() ignores nil errors.

i4ki: or in one line:
```
r.Error = errors.L(r.Error, other.Error).AsError()
```
I didn't test but I think this (should) work.

modvendor/download/report.go
```diff
@@ -109,6 +109,41 @@ func (r Report) Verbose() string {
...
+		if r.Error == nil {
+			r.Error = other.Error
+		} else {
+			r.Error = errors.L(r.Error, other.Error)
+		}
+	}
+
+	for k, v := range other.Vendored {
+		r.Vendored[k] = v
```
i4ki, line 141: merging the reports only keep the last report entries, is this ok? sorry if it's a stupid question but I don't remember the details in this case. Can you explain?

test/assert.go
```diff
@@ -0,0 +1,39 @@
...
+import "testing"
+
+// AssertEqualSets checks if two sets contains the same elements
+// independent of order (handles slices as sets).
+func AssertEqualSets[T comparable](t *testing.T, got, want []T) {
+	if len(got) != len(want) {
+		t.Fatalf("got: %+v; want: %+v", got, want)
+	}
+
+loop:
```
i4ki, line 26: yay!! hahaha

test/assert.go
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package test
+
+import "testing"
+
+// AssertEqualSets checks if two sets contains the same elements
+// independent of order (handles slices as sets).
+func AssertEqualSets[T comparable](t *testing.T, got, want []T) {
```
i4ki, line 21: top :rocket:

i4ki (dismissed):
well done.
I know it's still missing small things (exit codes, etc). Let me know when done and I review it again.
Cool stuff.
:rocket:

cmd/terramate/cli/cli.go
```diff
@@ -597,14 +604,59 @@ func (c *cli) cloneStack() {
 }
 
 func (c *cli) generate(workdir string) {
-	report := generate.Do(c.cfg(), workdir)
+	report, vendorReport := c.gencodeWithVendor(workdir)
+
 	c.output.Msg(out.V, report.Full())
 
+	vendorReport.RemoveIgnoredByKind(download.ErrAlreadyVendored)
```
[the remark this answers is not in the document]
katcipis: hmm yeah I was very on the fence on different ways to handle this, for now I tried to not mess around the core vendor logic much, but refactoring it/improving makes a lot of sense.

cmd/terramate/cli/cli.go
```diff
@@ -597,14 +604,59 @@ func (c *cli) cloneStack() {
 }
 
 func (c *cli) generate(workdir string) {
-	report := generate.Do(c.cfg(), workdir)
+	report, vendorReport := c.gencodeWithVendor(workdir)
```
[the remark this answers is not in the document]
katcipis: Hmm maybe I missed something, but at first I can't merge them, they are separated exactly because we handle the codegenReport + vendorReport differently on stack creation, so I can't have a single function that does everything (which would be more ideal).

I do agree that the name is bad...really struggling with it, because yeah vendor is always done, it is just a matter of having default report handling or customized report handling.

cmd/terramate/cli/cli.go
```diff
@@ -597,14 +604,59 @@ func (c *cli) cloneStack() {
...
+		vendorProgressEvents,
+	)
+
+	mergedVendorReport := download.MergeVendorReports(vendorReports)
+
+	log.Debug().Msg("generating code")
+
+	report := generate.Do(c.cfg(), workdir, c.vendorDir(), vendorRequestEvents)
+
+	log.Debug().Msg("stopping vendor request handler")
```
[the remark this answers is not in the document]
katcipis: Yeah, this one is for debug situation where odd things are happening, like maybe something si getting stuck forever (which already happened :laughing: ), the message does say it is "stopping", so if something still happens after the log I don't think it is misleading, stopping is always about sending the stop message (like closing the channel), and then we wait for confirmation it stopped (in that case I think then I log something "now it stopped" -P). But no strong felings on any of this, feel free to propose different/better messages :100:

cmd/terramate/cli/cli.go
```diff
@@ -783,7 +835,8 @@ func (c *cli) createStack() {
 	log.Info().Msgf("created stack %s", stackPath)
 	c.output.Msg(out.V, "Created stack %s", stackPath)
 
-	report := generate.Do(c.cfg(), stackDir)
+	// TODO(KATCIPIS): handle vendorReport
+	report, _ := c.gencodeWithVendor(stackDir)
```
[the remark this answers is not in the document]
katcipis: gencodeWithVendor already does exactly what you described, it does generation but does not abort, it returns both reports (there is no single error on codegen/vendoring). So the proposal would be to rename gencodeWithVendor ?

modvendor/download/report.go
katcipis: I think it can, yeah we do filter out nil automatically, cool :100:

modvendor/download/report.go
katcipis: I had to define some semantics for this, but in practice it would not happen today since when we vendor something with success any attempt of vendoring it again will result in a IgnoredVendor. But changes on the code could make this happen and right now we just override indeed, to avoid that we would require some more changes on the report to allow duplicated entires (the current map design doesn't work then).

Maybe this could be addressed later, really not sure what to do here (just saw it was safe for now to do overriding).

katcipis:
> I think we are missing info the vendor events. In this case, it tries to vendor for each stack but the output looks like it's retrying because precisely the same line prints several times. WDYT?
> 
> [asciicast](https://asciinema.org/a/538635)

I think you are right but the issue is not related to tm_vendor itself, it will only make it happen more often when the source is failing to be downloaded. So I think we can address this issue in a different PR in the future, improving output when retrying etc.

generate/genhcl/genhcl.go
```diff
@@ -167,6 +172,11 @@ func Load(
 			return nil, err
 		}
 
+		vendorTargetDir := project.NewPath(path.Join(
+			sm.Path().String(),
+			filepath.Dir(name)))
```
i4ki, line 177:
```suggestion
			path.Dir(name)))
```

generate/genfile/genfile.go
```diff
@@ -143,7 +148,13 @@ func Load(
 			continue
 		}
 
+		name := genFileBlock.Label
 		evalctx := stack.NewEvalCtx(projmeta, sm, globals)
+		vendorTargetDir := project.NewPath(path.Join(
+			sm.Path().String(),
+			filepath.Dir(name)))
```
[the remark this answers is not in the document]
i4ki: as `name` is a forward-slashed project path.

i4ki (approved):
LGTM

:rocket: :rocket: :rocket:

cmd/terramate/cli/cli.go
```diff
@@ -783,15 +834,23 @@ func (c *cli) createStack() {
 	log.Info().Msgf("created stack %s", stackPath)
 	c.output.Msg(out.V, "Created stack %s", stackPath)
 
-	report := generate.Do(c.cfg(), stackDir)
+	report, vendorReport := c.gencodeWithVendor(stackDir)
```
i4ki, line 838: you're not calling `vendorReport.RemoveIgnoredByKind()`. Is there a reason?

cmd/terramate/cli/cli.go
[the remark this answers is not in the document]
i4ki: > they are separated exactly because we handle the codegenReport + vendorReport differently on stack creation

When I added this review the stack creation handling was not implemented yet... I thought they would be the same.

cmd/terramate/cli/cli.go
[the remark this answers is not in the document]
katcipis: Hmm I get it, on my side, thinking on this:

"Having said that, if I see the message stopping vendor request handler then either we are abruptly stopping the "vendor request handler" or the "vendor request handler" itself finished and is now stopping"

I'm thinking more in terms of process A is sending a message for process B to stop, which is the channel being closed, it doesn't seem as confusing to me because in Go there is no way to abruptly stop another goroutine anyway, you can just send a message instructing it to stop, like sending a signal, etc. If the message said the goroutine is stopped then it would be misleading for sure, because it isn't (like you said), but Im just indicating that the stopping process started.

When the actual goroutine stops then we can have a debug saying it stopped. So I would see on the logs:

* "Stopping X"
* Some things happened
* "X stopped/finished"

So in general I don't find it misleading, but I also like the other message, will change it then :+1:

cmd/terramate/cli/cli.go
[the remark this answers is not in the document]
katcipis: Ah all good :+1:

cmd/terramate/cli/cli.go
katcipis: In this case we only output the reports if tehre is any errors on them, when there is no errors they go to output as -v, so I thought that for -v it was OK to keep everything since it is verbose output. But maybe on the HasFailures case it would also make sense to remove the ignored, was just not that sure here :thinking:

i4ki (approved):
lgtm

PR #716, opened 2022-11-16
fix: globals extended with empty block clearing values

i4ki:
This PR fixes the case where an existent globals object is extended with an empty labeled globals block. Example:

```
globals obj {
   a = 1
   b = 2
}

globals obj {

}
```

Before this patch, the final `global.obj` would contain only `{}`.

### DIFF
--- a/globals/globals.go
+++ b/globals/globals.go
@@ -195,7 +195,7 @@ func (globalExprs Exprs) Eval(ctx *eval.Context) EvalReport {
 
 		sort.SliceStable(sortedKeys, func(i, j int) bool {
 			expr1, expr2 := pendingExprs[sortedKeys[i]], pendingExprs[sortedKeys[j]]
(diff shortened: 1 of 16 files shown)


globals/globals.go
```diff
@@ -279,6 +282,23 @@ func (globalExprs Exprs) Eval(ctx *eval.Context) EvalReport {
...
+				// all the `attr = expr` inside global blocks become an entry
+				// in the globalExprs map but we have the special case that
+				// an empty globals block with labels must implicitly create
+				// the label defined object...
+				// then as it does not define any expression, an implicit
+				// expression for an empty object block is added to the map.
+				// This special entry sets the key accessor.isattr = false
+				// which means this expression doesn't come from an attribute.
+
+				// this if only happens if there's already an old value set
```
katcipis, line 295:
```suggestion
				// this only happens if there's already an old value set
```

katcipis (approved):

katcipis (approved):

PR #717, opened 2022-11-18
fix: allow terraform code with any kind of module definition

### LINKED ISSUE #697: [BUG] Terramate cannot parse overrides the same as terraform does
**Describe the bug**

Terramate doesn't parse override files the same as Terraform and errors on for example module override blocks without `source` specifications.

**To Reproduce**

1. Place a module block inside a `vpc.tf` file:
    ```terraform
    module "vpc" {
      source = "terraform-aws-modules/vpc/aws"

      name = "my-vpc"
      cidr = "10.0.0.0/16"

      azs             = ["eu-west-1a", "eu-west-1b", "eu-west-1c"]
      private_subnets = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
      public_subnets  = ["10.0.101.0/24", "10.0.102.0/24", "10.0.103.0/24"]
    
      enable_nat_gateway = true
      enable_vpn_gateway = true
    
      tags = {
        Terraform = "true"
        Environment = "dev"
      }
    }
    ```
3. Add a `vpc_override.tf` file, and override the module block. This module block is without a `source =` argument:
    ```terraform
    module "vpc" {
      name = "my-vpc-override"
    }
    ```
4. Run `terramate list --changed` and see example error:
    ```
    2022-11-07T16:30:49+01:00 ERR listing changed stacks error: checking module changes: applying operation to file "foo_override.tf": parsing modules: terraform schema error: module must have a "source" attribute file=/home/dev/dev/tf-stack/foo_override.tf:5,22-8,2
    ```

Adding a "duplicated" source line on all override blocks avoids this issue.

**Expected behavior**
Same parsing functionality as Terraform. Overides to work.

 - Terramate Version 0.1.40

katcipis:
# Reason for This Change

Terraform provides different was to define modules, as it happens with override described in #697. Instead of assuming and validating what is valid Terraform code we just look for information we need and discard any module blocks that we don't recognize but that could be valid Terraform (we don't assume anymore that we know how to validate Terraform code, Terraform is the one that knows this, we just look for information that is interesting to change detection and that is it).

Closes #697

### DIFF
--- a/tf/terraform.go
+++ b/tf/terraform.go
@@ -17,7 +17,6 @@ package tf
 import (
 	"os"
 
-	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/mineiros-io/terramate/errors"
@@ -32,11 +31,8 @@ type Module struct {
 	Source string // Source is the module source path (eg.: directory, git path, etc).
 }
 
-// Errors returned during the terraform parsing.
-const (
-	ErrHCLSyntax       errors.Kind = "HCL syntax error"
-	ErrTerraformSchema errors.Kind = "terraform schema error"
-)
+// ErrHCLSyntax represents a HCL syntax error
+const ErrHCLSyntax errors.Kind = "HCL syntax error"
 
 // IsLocal tells if module source is a local directory.
 func (m Module) IsLocal() bool {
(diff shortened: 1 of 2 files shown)


tf/terraform.go
```diff
@@ -86,21 +82,27 @@ func ParseModules(path string) ([]Module, error) {
...
+			continue
 		}
 		if !ok {
-			errs.Append(errors.E(ErrTerraformSchema,
-				hcl.RangeBetween(block.OpenBraceRange, block.CloseBraceRange),
-				"module must have a \"source\" attribute",
-			))
+			logger.Debug().Msg("ignoring module block without source")
+
+			continue
```
katcipis, line 105: I think we had a bug here, it was appending the error on the err list but also adding the module with an empty source on the result.

tf/terraform.go
```diff
@@ -86,21 +82,27 @@ func ParseModules(path string) ([]Module, error) {
...
+			Logger()
 
 		logger.Trace().Msg("Get source attribute.")
 		source, ok, err := findStringAttr(block, "source")
 		if err != nil {
-			errs.Append(errors.E(ErrTerraformSchema, err,
-				"looking for module.%q.source attribute", moduleName))
+			logger.Debug().
+				Err(err).
+				Msg("ignoring module block without source")
```
i4ki, line 98: in this case, it's not a missing source but it failed to evaluate the source expression.

tf/terraform.go
i4ki: yeah, luckily this didn't lead to a real bug because all the code using the module list is guarded by checking the source:
https://github.com/mineiros-io/terramate/blob/main/manager.go#L463
https://github.com/mineiros-io/terramate/blob/main/modvendor/download/download.go#L273

thanks for fixing this!

i4ki (approved):
lgtm but just a debug logging that maybe needs to change.

PR #720, opened 2022-11-19
chore: make a type distinction for root and subtree config.

i4ki:
# Reason for This Change

The #631 introduced the `config.Tree` type for the configuration tree but there was no distinction for the root tree.

The issue is that functions/methods received a `*config.Tree` as the first argument list but some of them expected a root tree and some any part of the tree (any subtree).
This is a potential issue because the compiler cannot catch the mistake of providing the wrong tree.

## Description of Changes

This PR introduces the `config.Root` type which shall be used whenever a root configuration is expected (eg.: generate family of routines).


cmd/terramate/cli/cli.go
```diff
@@ -637,7 +638,7 @@ func (c *cli) checkGitUncommited() bool {
 		}
 	}
 
-	cfg := c.prj.cfg.Node
+	cfg := c.prj.rootcfg.Tree().Node
```
katcipis, line 641: We seem to get the root config node multiple times, maybe we could have a method like rootNode and we cloud call:

```go
c.rootNode()
```

Or some other name.

katcipis (approved):
:rocket: :rocket: :rocket:

PR #721, opened 2022-11-19
chore: avoid re-parsing on stack.TryLoad()

i4ki:
# Reason for This Change

The `stack.TryLoad()` function still parses HCL code from disk instead of using the configuration tree.

## Description of Changes

Make the function uses the existing `config.Root` tree.


katcipis (approved):
top !!! :rocket: :rocket: :rocket:

katcipis (approved):

PR #722, opened 2022-11-19
chore: avoid re-parsing the configuration when creating stack

i4ki:
# Reason for This Change

When creating a new stack, the target directory is being re-parsed to check if a stack is defined there already but this is not needed as the configuration tree is available.

## Description of Changes

The `stack.Create()` was changed to lookup the configuration tree instead of calling `hcl.ParseDir()`.


katcipis (approved):
top !!! :rocket: :rocket: :rocket:

katcipis (approved):

PR #735, opened 2022-11-28
fix: check assertions every time stack code is loaded

katcipis:
# Reason for This Change

Assertions were just being checked on code generation, other logic loading code like outdated code detection and generate.Load were not executing assertions. On any generated code situation if an assertion is failing we should fail. In order to do that I moved the assertion loading/eval to where stack generated configs are loaded.


i4ki (approved):
lgtm

PR #737, opened 2022-11-30
chore: refactor output functions

mariux:


i4ki (commented):
Having a helper for outputs using no verbosity is nice but I'd not introduce N helpers for N verbosity levels.

We can have this in the cli:

```
func (c *cli) msgOut(format string, args ...interface{}) {
    c.output.Msg(out.V, format, args...)
}
func (c *cli) msgErr(format string, args ...interface{}) {
    c.output.Err(out.V, format, args...)
}
```

The API as is can be used for any verbosity, but we provide the maximum `VVVV` as a helper.

```
output.Msg(10, "very verbose message")
```

After this change, for verbosity > 3 we need to introduce more helpers that are pretty much just copy-paste.. with copy-pasted docs...

About the `out.write()` and `out.writeV()`, can you explain why you created a writer that only works with the exact verbosity passed? Like, what's the use case for `out.MsgOutV0()`? It looks like not used in the project.

The naming is also confusing because now we have `out.MsgOutV1()` and `out.MsgOutV()`, both have `verbosity=1` but different implementation. The difference is subtle in the docs.

We know the old naming is not the best, we were on the fence about the `out.V` being the "no verbosity" but Go has the philosophy of making the default value useful then the way it is you can have a data structure containing its own verbosity level and it defaults nicely to "no verbosity" if not initialized:

```
// O represents an output instance
type O struct {
	verboseness int
	stdout      io.Writer
	stderr      io.Writer
}
```

mariux:
we want to show some messages only in some verbosity...  so we can have different or additional output... atm the interface just supports additional line based output.. this will not let us refactor any output we currently have. i tried to change some things and it was not possible. so introducing additional and exclusive messages for each level.


just tried to come up with a proposal.. feel free to refactor.. out.V was misleading as we lack non verbose output atm...

mariux:
@i4ki what is the proposed alternative... to be able to refactor soem outputs i would need such functionality..

mariux:
> After this change, for verbosity > 3 we need to introduce more helpers that are pretty much just copy-paste.. with copy-pasted docs...

yes, but i doubt we will ever reach that state ;) normal + 3 verbosity levels is already a lot.

mariux:
we should also remove the line based output as on verbosity you might want to add more information in the same line?

mariux:
@i4ki how do we continue to get the needed functionality working, i will need to refactor some outputs soon.

i4ki:
I guess we can merge it and improve it later on if needed.

i4ki:
I will update the PR.

i4ki (approved):
lgtm

PR #738, opened 2022-11-30
fix: extending globals with importing exhibits random behavior

katcipis:
# Reason for This Change

When trying to extend a global on a parent directory, when the global being extended was imported, we get random behavior. SOmetimes extending works, sometimes it doesn't. A small script that reproduces the problem:

```sh
#!/bin/bash

set -o errexit
set -o nounset

workdir="$(mktemp -d)"
cd "${workdir}"

cat > terramate.tm <<- EOM 
terramate {
  config {
  }
}
EOM

mkdir module

cat > module/config.tm <<- EOM
globals "gclz_config" "terraform" "providers" {
  google = {
    enabled = true
    source  = "hashicorp/google"
  }
}
EOM

cat > config.tm <<- EOM 
import {
  source = "/module/config.tm"
}
EOM

terramate create stacks/stack-1
terramate create stacks/stack-2

cat > stacks/config.tm <<-EOM
globals "gclz_config" "terraform" "providers" "google" {
  enabled = false
  xxx = 666
}

globals {
  test = true
}
EOM

terramate experimental globals
```

On the output sometimes we get:

```
Created stack /stacks/stack-1
Created stack /stacks/stack-2

stack "/stacks/stack-1":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = true
                source  = "hashicorp/google"
              }
            }
          }
        }
        test = true

stack "/stacks/stack-2":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = true
                source  = "hashicorp/google"
              }
            }
          }
        }
        test = true
```

Sometimes:

```
Created stack /stacks/stack-1
Created stack /stacks/stack-2

stack "/stacks/stack-1":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = true
                source  = "hashicorp/google"
              }
            }
          }
        }
        test = true

stack "/stacks/stack-2":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = false
                source  = "hashicorp/google"
                xxx     = 666
              }
            }
          }
        }
        test = true
```

Sometimes:

```
Created stack /stacks/stack-1
Created stack /stacks/stack-2

stack "/stacks/stack-1":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = true
                source  = "hashicorp/google"
              }
            }
          }
        }
        test = true

stack "/stacks/stack-2":
        gclz_config = {
          terraform = {
            providers = {
              google = {
                enabled = false
                source  = "hashicorp/google"
              }
            }
          }
        }
        test = true
```

## Description of Changes

We need to define a proper order of which global to evaluate in order for the extension feature to work properly. The current ordering criteria would not work with import blocks since it depended on the origin of the expression, which in the imported global case will be the path from where it was imported (hence why it was failing only when integrating global extension with importing).

I tried to minimize the changes in order to fix this quickly and avoid introducing new errors, since this is critical, but with more time and rethinking some things we can probably think on something better and more efficient/etc.

The proposed change is to first have a loaded expressions map that retains the original configuration from where the global was loaded (not the same as origin), then we can have that map sorted by the config dir. Given the expressions sorted by config dir, then we can sort by the access path length (like we were already doing), but that is done for each config dir only, no global ordering, so we end up with parent first, then child dirs, which is the desired behavior since child dirs can extend what was defined on the parent.


globals/globals.go
```diff
@@ -48,18 +48,14 @@ type (
 		hhcl.Expression
 	}
 
-	// GlobalPathKey represents a global object accessor to be used as map key.
+	// GlobalPathKey represents a global object accessors to be used as map key.
```
katcipis, line 51:
```suggestion
	// GlobalPathKey represents a global object accessor to be used as map key.
```

mariux:
nice will test fixes asap.. you rock...

katcipis:
> nice will test fixes asap.. you rock...

tested here on our projects + the test script and so far it seems to have solved the issue and it didn't introduce new issues =P.

mariux:
lgtm ... tested on my use case and it creates consistent results now.

hcl/eval/object.go
```diff
@@ -151,6 +153,23 @@ func (obj *Object) SetAt(path ObjectPath, value Value) error {
 		path = path[1:]
 	}
 
+	old, ok := obj.GetKeyPath(path)
+	if ok {
+		// TODO(i4k): document this properly.
```
katcipis, line 158: On top of documenting that this is about parent detection we also need to make it clear that origin always must be inside the hierarchy of a stack or else this will not work (like the other bug related to the origin being from the import block file, outside the hierarchy).

Another idea I had at the time when I was thinking on something similar is to keep the hierarchical level of values, not the origin itself, so the object API would be more explicit about the hierarchical concept and how it affects behavior, but Im not sure it is a good idea, just sharing one of the ideas I had.

katcipis:
LGTM, thanks for the help on this @i4ki :heart:

i4ki (approved):

PR #745, opened 2022-12-07
fix: incorrect globals conflict handling

i4ki:
# Reason for This Change

The fix #738 introduced a bug in the globals conflict checks.
It can be reproduced with the configuration below:

```hcl
# modules/imported.tm
globals hello {
    world = {
       a = 1
    }
}
```

```hcl
# /file.tm
import {
    source = "/modules/imported.tm"
}

globals hello world {
    a = 2
}
```

When running Terramate it gives the error below:
```
global eval: terramate schema error: global.hello.world.a attribute redefined: previously defined at /
```

The reason is that when fixing #738 we switched to store the "config dir" in the eval.Object values because it was needed to solve the problem but the check for conflicting attributes (at runtime) needs the actual source file (not from where it's being imported).

## Description of Changes

The fix just stores both the configuration dir and the origin source path in the eval.Object.

### DIFF
--- a/hcl/eval/object.go
+++ b/hcl/eval/object.go
@@ -51,15 +51,15 @@ type (
 	//       }
 	//   }
 	Object struct {
-		origin Origin
+		origin Info
 		// Keys is a map of key names to values.
 		Keys map[string]Value
 	}
 
 	// Value is an evaluated value.
 	Value interface {
-		// Origin is the origin of the configuration value.
-		Origin() Origin
+		// Info provides extra information for the value.
+		Info() Info
 
 		// IsObject tells if the value is an object.
 		IsObject() bool
@@ -67,16 +67,16 @@ type (
 
 	// CtyValue is a wrapper for a raw cty value.
 	CtyValue struct {
-		origin Origin
+		origin Info
 		cty.Value
 	}
 
-	// Origin of a configuration value.
-	Origin struct {
-		// ConfigFrom defines where the value is being instantiated.
-		ConfigFrom project.Path
+	// Info provides extra information for the configuration value.
+	Info struct {
+		// Dir defines the directory from there the value is being instantiated.
+		Dir project.Path
 
-		// DefinedAt defines where the value is defined.
+		// DefinedAt defines the source file where the value is defined.
 		DefinedAt project.Path
 	}
 
@@ -85,7 +85,7 @@ type (
 )
 
 // NewObject creates a new object for configdir directory.
-func NewObject(origin Origin) *Object {
+func NewObject(origin Info) *Object {
 	return &Object{
 		origin: origin,
 		Keys:   make(map[string]Value),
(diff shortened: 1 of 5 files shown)


hcl/eval/object.go
```diff
@@ -51,38 +51,44 @@ type (
...
 
 	// CtyValue is a wrapper for a raw cty value.
 	CtyValue struct {
-		configdir project.Path
+		origin Origin
 		cty.Value
 	}
 
+	// Origin of a configuration value.
+	Origin struct {
```
katcipis, line 75:
```suggestion
	// Origin of a configuration value.
	// This represents the fact that [Object] has two origin related information that are
	// important to global conflict detection and global extensions. The two information are usually the
	// same paths, but will be different when "import" is used. In that case [Origin.DefinedAt] is the path
	// thas imported (the importee) and [Origin.ConfigFrom] is the path of the importer.
	Origin struct {
```

hcl/eval/object.go
```diff
@@ -290,8 +295,8 @@ func NewValue(val cty.Value, origin project.Path) Value {
 	return NewCtyValue(val, origin)
 }
 
-// ConfigOrigin is the configuration directory which defines the value.
-func (v CtyValue) ConfigOrigin() project.Path { return v.configdir }
+// Origin is the configuration directory which defines the value.
```
katcipis, line 298:
```suggestion
// Origin is all the origin related information about the value definition.
```

hcl/eval/object_test.go
```diff
@@ -41,8 +45,13 @@ func TestCtyObjectSetAt(t *testing.T) {
 		wantErr error
 	}
 
+	defaultOrigin := eval.Origin{
+		DefinedAt:  "/file.tm",
```
katcipis, line 49: DefinedAt is supposed to be a file not a dir ? in that case won't conflict detection fail ? globals defined on the same dir but on different files are supposed to conflict no ?

katcipis: Ah OK I saw you get the Dir(), my bad :laughing:

hcl/eval/object.go
```diff
@@ -205,7 +210,7 @@ func (obj *Object) SetAt(path ObjectPath, value Value) error {
 		// where's the definition of each value and overwrite the values if they
 		// come from a child scope or ignore new values that comes from the parent
 		// but conflicts with definitions in the child.
-		if !strings.HasPrefix(value.ConfigOrigin().String(), old.ConfigOrigin().String()) {
+		if !strings.HasPrefix(value.Origin().ConfigFrom.String(), old.Origin().ConfigFrom.String()) {
```
katcipis, line 213: hmm what did confuse me a little is that ConfigFrom is a dir and then DefinedAt is a file. but I get it now =P

katcipis (dismissed):
:rocket: :rocket: :rocket:

katcipis (approved):

PR #753, opened 2022-12-14
refactor: extract trigger e2e tests

katcipis:
# Reason for This Change

Specially since the trigger is an experimental feature it makes sense to test it separatedly from the core list/run tests. Just moved the tests to a separate file, no changes on the tests themselves.


i4ki (approved):
lgtm

PR #754, opened 2022-12-14
fix: missing build tags for GOOS=js

i4ki:
# Reason for This Change

Terramate is not building with `GOOS=js GOARCH=wasm` due to missing build rules.


waxb (approved):
lgtm

PR #756, opened 2022-12-15
fix: make iterator.old undefined on first iteration of a given key

i4ki:
# Reason for This Change

The `element.old` was set to `null` on the first iteration for a given key, but then it cannot be used with `tm_try()` in all cases.

## Description of Changes

Make the `element.old` undefined and only ever set when there were a value already set for the given key.


mariux (approved):
lgtm

PR #757, opened 2022-12-15
fix: properly postpone evaluation of expressions with unknowns

i4ki:
# Reason for This Change

The `Variables()` method of the `hcl.Expression` interface was not implemented, then the `tm_try()` inside `map` blocks referencing outer variables (`terramate`, `globals`, etc) was not correctly handled.

## Description of Changes

The `Variables()` method of the `MapExpr` is implemented by accumulating the `Variables()` of each of its attributes and subblock but ignoring local scoped variables (ie the iterator namespace).


mariux (approved):
lgtm

PR #758, opened 2022-12-16
chore: move tm functions to stdlib package

i4ki:
# Reason for This Change

The `tm_` functions were defined in the `eval` package for no important reason and this complicates the evolving of both evaluation and stdlib, independently.

## Description of Changes

A `stdlib` package was created to declare all Terramate functions.


mariux (approved):
lgtm

PR #759, opened 2022-12-16
chore: instantiation of stdlib functions should not fail

i4ki:
# Reason for This Change

The instantiation of stdlib can only ever fail in the case of programming errors, then this PR makes it a panic if the provided `basedir` is not correct.


mariux (approved):
lgtm

mariux (approved):
lgtm

PR #763, opened 2022-12-19
feat: add support for lets.map blocks

i4ki:
This PR introduces the `map` block inside the `lets` block.


stack/globals_test.go
```diff
@@ -2795,54 +2796,6 @@ func TestLoadGlobals(t *testing.T) {
 				),
 			},
 		},
-		{
```
i4ki, line 2798: tests moved to mapexpr/test/testcase.go so they could be reused with `lets` (with both genfile and genhcl).

mariux (approved):
lgtm

PR #772, opened 2022-12-23
chore: move stack.S to config.Stack.

i4ki:


stack/eval.go
```diff
@@ -61,15 +62,15 @@ func (e *EvalCtx) SetEnv(environ []string) {
...
+func stackMetaToCtyMap(m config.StackMetadata) map[string]cty.Value {
 	logger := log.With().
 		Str("action", "stack.stackMetaToCtyMap()").
 		Logger()
 
 	logger.Trace().Msg("creating stack metadata")
 
 	stackpath := cty.ObjectVal(map[string]cty.Value{
-		"absolute": cty.StringVal(m.Path().String()),
+		"absolute": cty.StringVal(m.Dir().String()),
```
mariux, line 73: we are using path all over the place why do we need to change it? i would propose to keep the naming aligned..

mariux (changes requested):

cmd/terramate/cli/cli.go
```diff
@@ -1037,7 +1037,7 @@ func (c *cli) generateGraph() {
 
 	visited := dag.Visited{}
 	for _, e := range c.filterStacksByWorkingDir(entries) {
-		if _, ok := visited[dag.ID(e.Stack.Path())]; ok {
+		if _, ok := visited[dag.ID(e.Stack.Dir())]; ok {
```
mariux, line 1040: this change seems out of scope of the PR?

mariux (changes requested):

stack/eval.go
i4ki: Today we use `Path()` all over the place but it's unclear when it's a directory or a file. For the configuration nodes, like the stack object, the path is actually a directory (always), but in some places the `Path()` is a filename.

Then, on main we have two list types with the same implementation:
- https://github.com/mineiros-io/terramate/blob/3e8fa9987e7fb5299f81961824c57eb023f1aac5/stack/stack.go#L352
- https://github.com/mineiros-io/terramate/blob/3e8fa9987e7fb5299f81961824c57eb023f1aac5/config/config.go#L313

both sorted by the directory path.

Now that I moved the evaluated stack object to the config package, we can reuse the same List type for different kind of configs (config.Tree, config.Stack, etc) but the nodes use `.Dir()` already, so in order to reuse the list type we need to make them alike (same interface):

so I created this interface:

```
// DirElem represents an element of the list which could represented by a
// directory. Eg.: stack.
type DirElem interface {
	Dir() project.Path
	HostDir() string
}

// List of directory nodes.
type List[T DirElem] []T
```

because atm all list configs are represented by directories, not arbitrary paths (filenames, etc).

If you'd like to keep the `.Path()` for everything, then we need to change `config.Tree` to use `.Path()` anyway in order to reuse the List.

But... `.Dir()` says more than `.Path()`, so when dealing with different objects at same time, it's less likely to incorrectly handle filenames as directories and vice-versa.

i4ki: let me know what you think.

mariux: why not use names that are super clear:
- StackHostPath (absolute host path) (matches `terramate.stack.path.fs.absolute` if exists ;))
- RootHostPath (matches `terramate.root.path.fs.absolute`)
- StackPathAbsolute (matches `terramate.stack.path.absolute`)
- StackPathRelative (matches `terramate.stack.path.relative`)
- StackPathBasename (matches `terramate.stack.path.basename`)

i4ki: If those proposed method names are for the `config.Stack` struct, then it's super weird (like not Go idiomatic).

1. if the struct is called `Stack` then its method doesn't need the "Stack" prefix on its methods.

2. Go interface types match by method name + signature (params + return types).
Then if we use a different name for each method on each object, we cannot create abstractions for reducing code duplication.

Currently, we have a different type for project paths (the `project.Path`) and then the compiler complains when they are misused (see #773) as a host path. So IMHO this is clear already.

Personally, I still feel we should differentiate the names when it's a directory (not talking about user-facing things like `terramate.stack.*`, this can continue the same), and your proposed names don't help with that.

mariux: okay... lets go :)

mariux (approved):
lgtm

waxb (approved):
lgtm

PR #773, opened 2022-12-24
chore: make project.Path a struct to better type check misuses of the API.

i4ki:
# Reason for This Change

The `project.Path` was an alias to `string` but a project.Path must never be a relative path, then the previous implementation allowed for the construction of invalid paths like `project.Path("relative/dir")`. This is impossible now.

Additionally, the `Root.Dir()` was renamed to `Root.HostDir()` to avoid confusion with project dir paths.

## Description of Changes

Make the project.Path a struct with an unexported `path` field, so now it's impossible to create invalid project paths.


modvendor/download/download_test.go
```diff
@@ -1162,24 +1162,6 @@ func TestModVendorNoRefFails(t *testing.T) {
 	}, report)
 }
 
-func TestModVendorVendorDirIsRelativeFails(t *testing.T) {
```
i4ki, line 1165: now it's impossible to have vendorDir as a relative path, so this test doesn't make sense anymore.

mariux (approved):
lgtm

PR #774, opened 2022-12-25
chore: make config.Stack independent of the rootdir

i4ki:
# Reason for This Change

Currently, the `config.Stack` needs to know the rootdir at creation time but this should not be needed, and this makes it difficult to create test stack objects in the sandbox, because in the test table we don't know the temporary directory in which tests would take place. 
An upcoming PR will make use of this when testing the evaluated config with stack objects.

## Description of Changes

Remove the embedded `hostdir` from the `config.Stack` and add the `*config.Root` as a parameter to `stack.HostDir(root)` and `stack.RelPathToRoot(root)` so the values are only computed when needed.


mariux (approved):
lgtm

PR #775, opened 2022-12-27
fix: allow relative stacks in experimental trigger

i4ki:
# Reason for This Change

Allow the use of relative paths when triggering stacks but fail if the provided directory resolves to a path outside the current project.

Example:

```
$ terramate experimental trigger /some/deep/stack # absolute project path works

$ terramate experimental trigger some/deep/stack # relative path works

$ cd some/deep && terramate experimental trigger stack # works
```

when providing paths with `../`, the resolved path must be inside the current project which Terramate is executing otherwise it fails.


mariux (approved):
lgtm cc @waxb please test ;)

PR #776, opened 2022-12-27
chore: change brand URL on README.md to point directly to the raw image.

i4ki:
# Reason for This Change

Use the GH raw URL instead of the `/blob` URL because the latter is not recognized as an image in other markdown renderers (eg.: markedjs).


i4ki:
@soerenmartius 

The url below returns HTML:
https://github.com/mineiros-io/brand/blob/38f6e5c20a176c62dc673f707a3987c9ec3691cd/terramate-vertical-logo-black-font.svg

whereas this returns an image:
https://raw.githubusercontent.com/mineiros-io/brand/38f6e5c20a176c62dc673f707a3987c9ec3691cd/terramate-vertical-logo-black-font.svg

soerenmartius (approved):
lgtm!

soerenmartius:
good catch!

PR #777, opened 2022-12-27
fix: require valid stack paths in triggers

i4ki:
# Reason for This Change

Validate only valid stack paths can be triggered.

# Description of changes

In the `trigger.Create()` the provided `path` is now lookup from the config tree and checked if it's indeed a stack and fail otherwise.


mariux (approved):
lgtm

mariux (approved):
lgtm

PR #780, opened 2022-12-27
fix: disallow symlinks in trigger stack path

i4ki:
# Reason for This Change

We must disallow symlinks in the trigger stack paths.

## Description of Changes

Fail if symlinks are resolved in any part of the stack path.


mariux (approved):
lgtm

PR #787, opened 2022-12-28
fix: multiple trigger parser issues (panics, errors ignored, etc)

i4ki:
# Reason for This Change

The Terramate panics (no trigger block) or ignores errors (wrong attribute types) with malformed trigger files.

## Description of Changes

Fixed the parser and added tests.


mariux (approved):
lgtm

PR #788, opened 2022-12-29
chore: remove project.Metadata and config.StackMetadata in favor of simple project.Runtime

i4ki:
# Reason for This Change

There was a lot of boilerplate and complexity in dealing with project and stack Metadata interfaces. The stack Metadata was required in the past to avoid cycles that are now resolved.

## Description of Changes

It was introduced the `project.Runtime` as a simple alias to `map[string]cty.Value`, then setting up the `terramate.*` namespace is just a matter of merging values in this map.


mariux (approved):
lgtm

PR #789, opened 2022-12-29
chore: remove the hcl.StackID type.

i4ki:
# Reason for This Change

We can use `stack.ID` as a plain string because the `id` cannot be empty and then we can check if the string is empty to figure out if it was set or not.

## Description of Changes

Remove the hcl.StackID type and use a string for the ID type.


mariux (approved):
lgtm

PR #790, opened 2022-12-29
chore: remove errors.StackMeta

i4ki:
# Reason for This Change

The `errors.E()` API supports providing a stack-like object which must implement the `error.StackMeta` interface.

This idea never took off because it was very brittle and then we had to make sure `errors.StackMeta` and `stack.Metadata` were compatible or messages would get wrongly formatted and other issues. There was only 2 places in the codebase using this feature of `errors.E()`, then it was removed and the stack path was explicitly provided in the error message.

Now we're moving on making the `config.Stack` a simple struct with evaluated public fields (which for instance can be set in tests), then the idea of matching by interface doesn't make sense anymore.

## Description of Changes

The `errors.StackMeta` was removed.

### DIFF
--- a/cmd/terramate/e2etests/run_test.go
+++ b/cmd/terramate/e2etests/run_test.go
@@ -1819,7 +1819,6 @@ func TestRunContinueOnError(t *testing.T) {
 	t.Parallel()
 
 	s := sandbox.New(t)
-
 	s.BuildTree([]string{
 		`s:s1`,
 		`s:s2`,
@@ -1836,8 +1835,8 @@ func TestRunContinueOnError(t *testing.T) {
 
 	cli := newCLI(t, s.RootDir())
 	assertRunResult(t, cli.run("run", "cat", "main.tf"), runExpected{
-		IgnoreStderr: true,
-		Status:       1,
+		StderrRegex: "one or more commands failed",
+		Status:      1,
 	})
 
 	assertRunResult(t, cli.run("run", "--continue-on-error", "cat", "main.tf"), runExpected{
--- a/errors/error.go
+++ b/errors/error.go
@@ -21,12 +21,12 @@ package errors
 import (
 	"errors"
 	"fmt"
+	"reflect"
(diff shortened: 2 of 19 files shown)


cmd/terramate/e2etests/run_test.go
```diff
@@ -1836,8 +1835,8 @@ func TestRunContinueOnError(t *testing.T) {
 
 	cli := newCLI(t, s.RootDir())
 	assertRunResult(t, cli.run("run", "cat", "main.tf"), runExpected{
-		IgnoreStderr: true,
-		Status:       1,
+		StderrRegex: "one or more commands failed",
```
i4ki, line 1838: this test was not behaving correctly. While testing this PR, I forced a panic() in the core Terramate and this test passed with the panic.

errors/error.go
```diff
@@ -162,6 +149,13 @@ func E(args ...interface{}) *Error {
 
 	errs := L()
 	for _, arg := range args {
+		if arg == nil {
+			panic("errors.E() called with nil value in argument")
+		}
+		// explicit check for deprecated usage of the API.
+		if reflect.TypeOf(arg).String() == "*config.Stack" {
+			panic("errors.E() does not support type *config.Stack anymore")
```
i4ki, line 157: I checked all the occurrences and this should never trigger, but it doesn't hurt to keep this panic just in case someone is using this type (it's public).

mariux: lets add information that this is a bug and users should create an issue...

i4ki: I improved the panic message, now it prints:
```
panic: BUG: Deprecated use of argument type hcl.Range in the errors.E() function.
If this panic was triggered in the Terramate project, then please open an issue
at https://github.com/mineiros-io/terramate/issues.
```
wdyt?

mariux: great ;) thanks

mariux (approved):
lgtm

PR #792, opened 2022-12-29
feat: introduce the stack.tags attribute

i4ki:
# Reason for This Change

Implements the `stack.tags` attribute for the `stack` block.

Now stacks can have a `set(string)` of tags which are exported in the `terramate.stack.tags` runtime variable, ie, can be referenced in `lets` and `globals`, and can also be used inside partial evaluated HCL content in `generate_hcl.content` blocks.

The `tags` can also be visible in the cli command `terramate experimental metadata`.

Upcoming features will make use of `tags` in the orchestration!

## Description of Changes

Added the `tags` attribute to the parser, exported it in the `terramate.stack.tags` variable and outputs them in `terramate experimental metadata` command.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -39,6 +39,8 @@ import (
 	"github.com/mineiros-io/terramate/stack/trigger"
 	"github.com/mineiros-io/terramate/stdlib"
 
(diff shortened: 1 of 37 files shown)


mariux (approved):
lgtm

PR #794, opened 2023-01-03
feat: add --tags filter to multiple commands

i4ki:
This PR introduces the `--tags` flag for filtering stacks by tags.

The tags filter uses the syntax below:

```
EXPR    = TAGNAME [ OP EXPR]
TAGNAME = <string>
OP      = ":" | ","
```
Where `:` and `,` are used for `AND` and `OR` logical operations, respectively.

Examples:

```
--tags "app:prod"           # selects all stacks with app AND prod tag.
--tags "k8s,nomad,docker"   # selects all stacks with k8s OR nomad OR docker tag.
```

The `:` has precedence over `,` so if multiple are provided, the `:` are grouped first:

```
--tags A:B,C    # (A && B) || C
--tags A,B:C:D  # A || (B && C && D)
```

If multiple `--tags` are provided, then they are grouped in a `OR` query:

```
--tags app --tags prod            # is the same as: --tags app,prod
--tags app:prod --tags infra:prod # is the same as: --tags app:prod,infra:prod
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -27,6 +27,7 @@ import (
 	hhcl "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/mineiros-io/terramate/cmd/terramate/cli/out"
+	"github.com/mineiros-io/terramate/config/filter"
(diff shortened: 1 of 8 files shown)


mariux (approved):
lgtm

PR #798, opened 2023-01-06
feat: allow ordering stacks by tag filters

i4ki:
# Reason for This Change

Introduce the `tag:<query>` syntax in the elements of `stack.after` and `stack.before` attributes.

Example: 

```
stack {
    after = [
        "tag:prod:infra",
        "tag:teamA,teamB",
    ]
}
```

In the example above, the stack must run after all stacks containing both `prod` and `infra` tags, but also after all stacks containing `teamA` or `teamB`.


run/order.go
```diff
@@ -157,9 +158,26 @@ func BuildDAG(
...
 
-	removeWrongPaths := func(fieldname string, paths []string) []string {
-		cleanpaths := []string{}
+	computePaths := func(fieldname string, paths []string) ([]string, error) {
+		uniqPaths := map[string]struct{}{}
 		for _, pathstr := range paths {
+			if strings.HasPrefix(pathstr, "tag:") {
+				if fieldname != "before" && fieldname != "after" {
+					return nil, errors.E(
+						"tag:<query> is not allowed in %q field", fieldname,
```
i4ki, line 167: @mariux this is where we disallow `tag:` filters in `wants` and `wanted_by`.

mariux: nice... :+1:

mariux (approved):
lgtm

PR #800, opened 2023-01-10
chore: fix performance regression introduced by #788

i4ki:
# Reason for This Change

The #788 introduced a subtle change that affected the performance.

Before the project metadata was instantiated by `project.NewMetadata()` which traversed the config tree to fill the `terramate.stacks.list` object, this was done once and the project.Metadata interface was passed over to each stack code generation.

With #788 the `project.Metadata` and `stack.Metadata` interfaces were removed and the runtime objects were constructed when needed, which means config tree was walked over multiple times for the root runtime values.

The root metadata never changes, so it can be safely loaded once when the project config is loaded and then reused.

## Description of Changes

Added a `runtime` field to `config.Root` struct and the method `root.Runtime()` returns a copy of the computed runtime.
So the `root.Stacks()` method is called once during the config loading.


mariux (approved):
lgtm

PR #804, opened 2023-01-11
feat: add --global flag to experimental eval

i4ki:
Introduces the `--global <name>=<val>` flag to the commands `get-config-value`, `eval` and `partial-eval`.

The provided list of global flags can be used to override the values from the configuration or set new ones before handling the command.

Example:

```
$ cat config.tm
globals {
    config = {
        name = "my project"
    }
}
$ terramate experimental eval --global config={} global.config
{}
```

It's possible to set just a specific part of an existent global value, if needed.

```
$ terramate experimental eval --global 'config.name="other value"' global.config
{
    name = "other value"
}
```

It also supports a syntax sugar to extend objects dynamically using a `dot notation`.
So if the object path does not exist, the object is dynamically created.

For the same configuration specified before, the command below works:

```
$ terramate experimental eval --global 'config.firewall.rules=[{allow = "*"}]' global.config
{
  name = "my project"
  firewall = {
    rules = [{
      allow = "*"
    }]
  }
}
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -33,6 +33,7 @@ import (
 	"github.com/mineiros-io/terramate/event"
 	"github.com/mineiros-io/terramate/generate"
 	"github.com/mineiros-io/terramate/globals"
+	"github.com/mineiros-io/terramate/hcl/dynexpr"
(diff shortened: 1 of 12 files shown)


mariux (dismissed):
lgtm ;)

mariux (approved):
lgtm

PR #813, opened 2023-01-20
fix: git worktree was not supported

i4ki:
# Reason for This Change

We were using `git rev-parse --git-dir` to found the `.git` directory and then infer the project root but this approach doesn't work with `git worktree`.

We are not officially announcing support for it yet as we need to better test if all Terramate features work well with `worktree`, but this change in one step in that direction.

## Description of Changes

It was simply changed to use `git rev-parse --show-toplevel`.


mariux (approved):
lgtm

PR #821, opened 2023-01-26
chore: remove token injection hack

i4ki:
This PR removes an ugly hack from Terramate which caused a lot of trouble in the past.
As the HCL library doesn't provide `Expression -> Tokens` method, we loaded the original tokens from the source file again when needed, using the `expr.Range()` info, and for the case of partial-eval, there was no source for dynamically modified token streams then we stored the tokens inside the range.Filename for later processing...

The PR #820 eliminates the need for such workarounds as it walks through the AST and generates the tokens. As the parsed nodes lack some syntactic information, it means the generated tokens cannot give back the exact tokens as they are in the source file.

Examples:

- ~~All strings with interpolation are represented by `hclsyntax.TemplateExpr` but there's no field to differentiate between a quoted string and a HEREDOC string. Then we generate a quoted string always.~~ 
Any multiline string (heredoc or plain strings with `\n`) returns a HEREDOC string. Inline strings always return a plain string (quoted string).

- The syntax `a.0` is supported in HCL (but deprecated) and it's parsed in the same node as `a[0]`, with no difference, in the `hclsyntax.IndexExpr` type. Then we generate `a[0]` always.

- By same reason as above, the splat `a.*.b` will become `a[*].b`.

- Literal numbers with unneeded trailing zeros are stripped. Eg.: `1.000` becomes `1`, `0.100` becomes `0.1`, etc

- As we lack all the user style (newlines, spaces, etc) from original file, then we generate a more opinionated style (formatted). Eg.: inline objects are never generated.

- All cases above are also true for HIL interpolations (check all examples below)

Example of generated code changes below:

Inline objects are not generated anymore:
```
obj = {a = 1, b = 2}
```
Generates:
```
obj = {
  a = 1
  b = 2
}
```

Optional commas in objects are stripped:
```
obj = {
  a = 1,
  b = 2,
}
```

Generates:
```
obj = {
  a = 1
  b = 2
}
```

Multiline string returns a HEREDOC:
```
msg = "Welcome to the XPTO Infrastructure\n\t\t-By Mineiros.io"
```
Generates:
```
msg = <<-EOT
Welcome to the XPTO Infrastructure
        - By Mineiros.io
EOT
```

HIL interpolation also has its code reformatted with same rules:

```
value = "prefix string ${ {a = 1, b = 2}["a"] } suffix string"
```
Generates:
```
value = "prefix string ${ {
  a = 1
  b = 2
}["a"]} suffix string"
```

Multiline strings inside HIL expressions:
```
value = "testing ${"some\nmultiline\nstring"}"
```
Generates:
```
value = "testing ${<<-EOT
some
multiline
string
EOT
}"
```

The same rules apply if the multiline string is defined in a `global/let` variable and referenced in generation.

```
globals {
    str = "some\nmultiline\nstring"
}

generate_hcl "file.hcl" {
  content {
    value = "testing ${global.str}"
  }
}
```
Generates:
```
value = "testing ${<<-EOT
some
multiline
string
EOT
}"
```


cmd/terramate/cli/cli.go
```diff
@@ -1334,33 +1334,29 @@ func (c *cli) checkGenCode() bool {
...
-
 		c.outputEvalResult(val, c.parsedArgs.Experimental.Eval.AsJSON)
 	}
 }
 
 func (c *cli) partialEval() {
 	ctx := c.setupEvalContext(c.parsedArgs.Experimental.PartialEval.Global)
 	for _, exprStr := range c.parsedArgs.Experimental.PartialEval.Exprs {
-		expr, err := dynexpr.ParseExpressionBytes([]byte(exprStr))
+		expr, err := ast.ParseExpression(exprStr, "<cmdline>")
```
mariux, line 1352: nice ;)

mariux (approved):
lgtm

mariux (approved):
lgtm

PR #823, opened 2023-01-29
chore: add benchmark regression check

i4ki:
# Reason for This Change

We should be aware of performance regressions on core parts of Terramate.
This PR adds a [benchcheck](https://github.com/madlambda/benchcheck) job in the CI which fails in the case of performance regressions (at the moment, just for hcl/ast benchmarks).

## Description of Changes

Makefile and GHA workflow setup.


i4ki:
https://github.com/mineiros-io/terramate/actions/runs/4034785125/jobs/6936264247#step:4:12

PR #826, opened 2023-02-04
chore: remove stack.CreateCfg and simplify tests

i4ki:
# Reason for This Change

In the past, the `config.Stack` fields were not exposed and then a separate structure was needed to configure new stacks. This is not needed anymore and the `config.Stack` is an standalone object.

A good result of this change is that now it's impossible to call `stack.Create()` with invalid paths (relative paths, paths outside rootdir, etc).

## Description of Changes

- Remove `stack.CreateCfg` and use the `config.Stack` object directly.
- refactor tests to use the new API.
- Some tests were removed (or moved to another place) because they are impossible to occur at the `stack.Create()` time.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -855,20 +855,19 @@ func (c *cli) createStack() {
 		stackDescription = stackName
 	}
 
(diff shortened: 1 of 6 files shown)


mariux (approved):
lgtm

PR #828, opened 2023-02-05
fix: create command can generate invalid stack definitions

i4ki:
# Reason for This Change

The `terramate create` command was not fully validating the stack configuration.
By the schema, the fields `after`, `before`, `wants` and `wanted_by` are a `set(string)` but the create command allows for creation of stacks with duplicated entries in those fields.


mariux (approved):
lgtm

PR #830, opened 2023-02-06
fix: only generate HEREDOC when string ends with nl

i4ki:
# Reason for This Change

The #820 change incorrectly generated any multiline string as HEREDOC but only strings that ends with newline should be converted to it.

## Description of Changes

Checks if the last character is `\n` and only returns HEREDOC in that case.


mariux (approved):
lgtm

PR #832, opened 2023-02-06
chore: avoid lookups+eval of directories without globals

i4ki:
# Reason for This Change

When evaluating the globals for a directory, we also evaluate the parent directories even when no globals are defined. This PR skips directories with no globals defined.

## Description of Changes

The `*config.Tree.NonEmpttGlobalsParent()` was added to retrieve a parent directory which must have globals defined or nil if none is found.
The `globals.LoadExpr()` API was changed to work with the `config.Tree` instead of having to lookup the path each time.


mariux (approved):
lgtm

PR #822, opened 2023-01-29
chore: add TokensForExpression benchmarks

i4ki:

PR #833, opened 2023-02-06
chore: bench/check against main

i4ki:
# Reason for This Change

When the benchmarks were added, the TokensForExpression benchmark functions were not merged then the bench/check was added to check against the #822 branch.

This PR points the checks against `main`. This fixes the wrong bench results being sent to PRs.


mariux (approved):
lgtm

PR #836, opened 2023-02-07
chore: remove generate benchmarks

i4ki:
# Reason for This Change

The `generate` benchmarks were added a long time ago but are not helpful as they are too high-level, slow and unpredictable.

A new set of localized benchmarks will be added later for benchmarking internal expensive functions of the generate.

## Description of Changes

Removed the `generate/generate_bench_test.go` and set the `bench/check` make a rule to bench the entire project.


i4ki:
The benchmark failed because it tests against `main` and main still has the slow benchmarks and hangs the tests.

mariux (approved):

PR #837, opened 2023-02-07
chore: use ast.TokensForExpression in tests instead of re-scanning files.

i4ki:
# Reason for This Change

The same optimization we did in the core Terramate can now be done in the tests. So file re-scanning is avoided.


mariux (approved):
lgtm

PR #838, opened 2023-02-07
chore: implements partial evaluator at the AST level instead of lexical tokens.

i4ki:
# Reason for This Change

In the benchmark below, the `_v1` refers to the old implementation and `_v2` to this implementation.

Simple expression:
```
goos: linux
goarch: amd64
pkg: github.com/mineiros-io/terramate/hcl/eval
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkPartialEval_v1-12    	   76551	     23795 ns/op	    9688 B/op	     248 allocs/op
BenchmarkPartialEval_v1-12    	   41655	     25689 ns/op	    9688 B/op	     248 allocs/op
BenchmarkPartialEval_v1-12    	   56005	     21671 ns/op	    9688 B/op	     248 allocs/op
BenchmarkPartialEval_v1-12    	   67210	     23182 ns/op	    9688 B/op	     248 allocs/op
BenchmarkPartialEval_v1-12    	   41947	     26917 ns/op	    9688 B/op	     248 allocs/op
BenchmarkPartialEval_v2-12    	  183658	      6549 ns/op	    2728 B/op	      81 allocs/op
BenchmarkPartialEval_v2-12    	  195541	      6851 ns/op	    2728 B/op	      81 allocs/op
BenchmarkPartialEval_v2-12    	  266918	      6157 ns/op	    2728 B/op	      81 allocs/op
BenchmarkPartialEval_v2-12    	  225440	      6894 ns/op	    2728 B/op	      81 allocs/op
BenchmarkPartialEval_v2-12    	  183746	      5606 ns/op	    2728 B/op	      81 allocs/op
PASS
ok  	github.com/mineiros-io/terramate/hcl/eval	17.969s
```

Complex expression:
```
goos: linux
goarch: amd64
pkg: github.com/mineiros-io/terramate/hcl/eval
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkPartialEval/v1-12 	    7845	    185529 ns/op	   69514 B/op	    1162 allocs/op
BenchmarkPartialEval/v1-12 	    7269	    220940 ns/op	   69513 B/op	    1162 allocs/op
BenchmarkPartialEval/v1-12 	    6042	    171614 ns/op	   69514 B/op	    1162 allocs/op
BenchmarkPartialEval/v1-12 	    7035	    175034 ns/op	   69513 B/op	    1162 allocs/op
[...]

PR #839, opened 2023-02-08
chore: add benchmarks for partial evaluator

i4ki:
# Reason for This Change

Those benchmarks are needed to assess if #838 is a good move.


mariux (approved):
lgtm

PR #846, opened 2023-02-20
feat: support tm_dynamic.attributes together with tm_dynamic.content

i4ki:
Add support for both `tm_dynamic.attributes` and `tm_dynamic.content` block in the same `tm_dynamic` block, given that they don't conflict. A conflict would be having a `attributes` field also defined in the `content` block.

The result is a merge of both configuration types.


mariux (dismissed):
lgtm

mariux (approved):

PR #847, opened 2023-02-21
feat: add --no-tags flag

i4ki:
# Reason for This Change

Add `--no-tags` flag which filter by stacks not having the provided list of tags.
It could be combined with `--tags` and in such case a `AND` condition is generated.
Then `--tags prod --no-tags k8s` will return all stacks which have `prod` tag and not `k8s` tag.

## Description of Changes

The internal tag filter was updated to support tag inequality but the syntax is not exposed to users.


mariux (approved):
great stuff :1st_place_medal: ... lgtm

PR #848, opened 2023-02-23
feat: allow multiple definitions of terramate.required_version

i4ki:
draft


i4ki:
another approach was decided

PR #851, opened 2023-02-25
chore: add tests for version constraints

i4ki:
# Reason for This Change

We only had a few e2e tests for version constraints.
The tests added here are not exhaustive but just test the basic behavior we intent to fix in an upcoming feature.

## Description of Changes

Added a separate `terramate.CheckVersionFor()` function which checks the constraints against any version (for testing purposes) and added tests.


version_test.go
```diff
@@ -0,0 +1,175 @@
...
+			constraint: "> 1.2.3-aaa", // match because zzz > aaa
+		},
+		{
+			version:    "1.2.3-zzz",
+			constraint: "> 1, > 1.2.3-aaa", // doesnt match because > 1 never matches pre-releases
+			want:       errors.E(terramate.ErrVersion),
+		},
+		{
+			version:    "1.2.3-zzz",
+			constraint: "> 1.2.3-aaa", // match because aab > aaa
```
i4ki, line 150:
```suggestion
			constraint: "> 1.2.3-aaa", // match because zzz > aaa
```

soerenmartius (approved):
lgtm

version_test.go
```diff
@@ -0,0 +1,175 @@
...
+	type testcase struct {
+		version    string
+		constraint string
+		want       error
+	}
+
+	for _, tc := range []testcase{
+		{
+			version:    "0.0.0",
+			constraint: "0.0.0",
```
i4ki, line 38: @mariux this is currently valid but it's not with `apparentlymart/go-versions` package:
https://github.com/apparentlymart/go-versions#unspecified-versions
So when user provides `terramate.required_version_allow_pre_releases` it will fail.
Is it ok?

PR #855, opened 2023-02-27
feat: introduce tm_version_match function

i4ki:
Introduces the function `tm_version_match(version, constraint, optional_args)` where the `optional_args` is one variadic argument of type object and schema below:
```
{
    allow_prereleases: bool
}
```
If `optional_args.allow_prereleases` is `true` then it also matches prereleases accordingly to Semver specification.


mariux (approved):
super nice ;) thanks

mariux (approved):

PR #867, opened 2023-03-14
fix: global expressions with pending indexing references 

i4ki:


generate/genhcl/genhcl_test.go
```diff
@@ -1576,6 +1576,65 @@ func TestGenerateHCL(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "generate HCL on stack with lets referencing indexing variable",
```
i4ki, line 1580: the issue is not related to `lets` but I added a test for lets anyway because we will handle them the same soon.

mariux (approved):
lgtm ;) amazing

globals/globals_test.go
```diff
@@ -3606,8 +3846,6 @@ func TestLoadGlobalsErrors(t *testing.T) {
 					`,
 				},
 			},
-			// FIXME(katcipis): would be better to have ErrGlobalRedefined
```
i4ki, line 3609: it doesn't make sense to return ErrGlobalRedefined because multiple attributes with the same name is a syntax error.

PR #870, opened 2023-03-15
fix: funcall ellipsis argument (...) is not rendered in code generation

### LINKED ISSUE #869: [BUG] `... ' modifier is lost during generate_hcl, how to pass it to generated HCL
**Describe the bug**

I'm trying to use "generate_hcl" and `...` modifier is lost. Which is expected if the HCL is evaluated but I'm unsure how to pass it down to generated HCL?

I need to use this because I've a nested `for` and I'd like to flatten and merge the final result.

Do I have to use `generate_file` in this case? 

Ideally, I'd love to do the whole preprocessing in `globals` block and pass the final output to generated HCL but `merge` and other functions are not working there. 

**To Reproduce**

1. Given context:

```hcl
generate_hcl "_terramate_deployments.gen.tf" {
  content {
    locals {
      app_env_deployments = global.app_env_deployments
      app_envs            = global.app_envs

      deployments = merge([
          for app_env in local.app_envs: {
            for name in local.app_env_deployments:
              "${name}-${app_env}" => {
                env  = app_env
                name = name
              }
          }
      ]...)
    }
}
```

3. Run command: `terramate generate`

**Expected behavior**

`...` is present in generated output

**Log Output**

-

 - OS: [e.g. Linux] Linux
 - OS Version [e.g. Ubuntu 20.04] Ubuntu 22.04
 - Git Version [e.g. 2.35.1] (git --version): 2.30
 - Terramate Version [e.g. 0.1.0] (terramate --version): 0.2.14

**Additional context**

none

i4ki:
# Reason for This Change

When rendering function calls, the node has the `funcall.ExpandFinal` flag which tells if the last argument has the ellipsis (...) symbol. This was not checked when rendering HCL.

## Description of Changes

Renders the `...` when `fn.ExpandFinal` is `true`.

Closes #869


hcl/ast/expr.go
mariux, line 1: uhhh comments on full files... ;)

PR #871, opened 2023-03-15
fix: render ellipsis in the for-expr

i4ki:
# Reason for This Change

The `...` (ellipsis) is not being rendered in `for-expr`.
Example:
```
{for i, v in ["a", "a", "b"]: v => i...}
```
when generating code it renders:
```
{for i, v in ["a", "a", "b"]: v => i}
``` 

## Description of Changes

Renders the ellipsis if used.


mariux:
relates to #870 and #869

PR #878, opened 2023-03-22
feat: add basic version upgrade check

i4ki:
Introduces upgrade checks in the `terramate version` command.
It uses the same technology used by Terraform.
- backend based on: https://checkpoint.hashicorp.com/
- client based on: https://github.com/hashicorp/go-checkpoint

The check can be disabled by exporting the environment variable `CHECKPOINT_DISABLE=1`.
A completely random signature (UUID v4) is generated and stored in `~/.terramate.d/checkpoint_signature`.

The check is cached and invalidated every 48 hours.

The implementation mimicks what Terraform does, the implementation is very similar.

### DIFF
--- a/docs/upgrade-check.md
+++ b/docs/upgrade-check.md
@@ -1,0 +1,22 @@
+# Upgrade and Security Bulletin Checks
+
+The Terramate CLI commands interacts with the [Mineiros Checkpoint](https://checkpoint-api.mineiros.io/)
+service to check for the availability of new versions and for critical security 
+bulletins about the current version.
+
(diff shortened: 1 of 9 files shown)


docs/upgrade-check.md
```diff
@@ -0,0 +1,22 @@
+# Upgrade and Security Bulletin Checks
```
i4ki, line 1: @soerenmartius the docs is based on [this](https://developer.hashicorp.com/terraform/cli/commands#upgrade-and-security-bulletin-checks).

soerenmartius: looks good to me! thanks!

soerenmartius (approved):
lgt

PR #883, opened 2023-03-24
feat: add cli configuration file for disabling checkpoint

i4ki:
This PR adds CLI flags for disabling checkpoint and introduces the Terramate CLI Configuration file.

On UNIX the file is named `.terramaterc` and must be placed in the user's HOME directory.
On Windows the file is named `terramate.rc` and must be placed in the user's `%APPDATA%` directory (you can use `$env:APPDATA` in the PowerShell to find its location).

The configuration language is plain `HCL`, only literal values could be used in the expressions, and the only options at the moment are:

- disable_checkpoint (type bool)
- disable_checkpoint_signature (type bool)

Example below:

```
# /home/i4k/.terramaterc

disable_checkpoint = false
disable_checkpoint_signature = true
```


soerenmartius (approved):
lgtm

PR #885, opened 2023-03-27
chore: add generate benchmark for referenced globals

i4ki:
# Reason for This Change

Add a generate benchmark for the case where there are a lot of globals defined in a hierarchy of directories but only a few are actually referenced in generate blocks.


soerenmartius (approved):
lgtm! good job!

PR #891, opened 2023-04-01
Restructure the entry README.md

soerenmartius:
# Reason for This Change

Cleaning up the main entry README.md so folks don't get lost when they see Terramate for the first time


i4ki (approved):
lgtm

PR #894, opened 2023-04-03
chore: misleading error message in tm_dynamic.labels

i4ki:
# Reason for This Change

When using `tm_dynamic.labels` with an invalid type, it failed with the message:

```
invalid tm_dynamic.labels: tm_dynamic.labels is not a string list: value must be a set(string) but ...
```
But the `labels` must be a `list(string)` not a `set(string)` as duplicated values *are allowed*. The code do allow for duplication, the issue was just in the error message.
An additional test for duplicated labels was added just in case.

## Description of Changes

The error message was updated and a duplicated labels test was added.


mariux (approved):
lgtm

PR #897, opened 2023-04-04
chore: add stack list tests for git submodules

i4ki:
# Reason for This Change

Add initial tests for `git submodule`.
For now, just ensure stack listing works in a repo with initialized submodules.

@mariux we have a common `listTestcases()` function that returns a list of testcases that **must** work independently if the project is a git repository, a bare directory with a terramate.config defined, etc. I'm using the same list of testcases here to ensure that `list` works also if the files are inside a `git submodule`. So in this case I added a single test function but it's tested a long range of cases.

NOTE: I inadvertently merged the #898 PR into this one, sorry about that.

## Description of Changes

Tests added.


test/sandbox/git.go
```diff
@@ -225,12 +232,19 @@ func (git Git) Pull(branch string) {
 	}
 }
 
-// CommitAll will add all changed files and commit all of them
-func (git Git) CommitAll(msg string) {
+// CommitAll will add all changed files and commit all of them.
+// It requires files to be commited otherwise it fails.
```
i4ki, line 236:
```suggestion
// It requires files to be committed otherwise it fails.
```

test/sandbox/git.go
```diff
@@ -162,6 +162,13 @@ func (git Git) Add(files ...string) {
 	}
 }
 
+func (git Git) AddSubmodule(name string, url string) {
```
i4ki, line 165:
```suggestion
// AddSubmodule adds name as a submodule for the provided url.
func (git Git) AddSubmodule(name string, url string) {
```

mariux (approved):
lgtm

PR #903, opened 2023-04-09
Use Alpine instead of Ubuntu for Docker

soerenmartius:
# Reason for This Change

Refactor build to use Alpine instead of Ubuntu, resulting in a smaller image. @i4ki what do you think about providing both, alpine and ubuntu versions?


i4ki:
> what do you think about providing both, alpine and ubuntu versions?

We will need a naming scheme for that, no? like `terramate:<version>` being the default OS (maybe `alpine`) and `terramate-ubuntu:<version>`?

If you think it's needed, then let's do it.

i4ki:
Just noticed that our current ubuntu Dockerfile is not shipping `git`, then it cannot do change detection. Thanks for fixing that!

But just I tested the Alpine dockerfile you proposed and it was not able to do change detection either.
When providing `--changed` it fails with:

```
2023-04-09T14:08:05Z FTL flag --changed provided but no git repository found
```

Then I tried to execute any git command inside the repo and it gives the error below:
```
$ docker run -it terramate-alpine --log-level trace run -- git branch
... skipped ...
2023-04-09T13:57:08Z TRC loaded stacks run environment variables, running commands action=run.Exec() cmd="git branch"
2023-04-09T13:57:08Z INF running cmd="git branch" stack=/nginx
fatal: detected dubious ownership in repository at '/root/tutorial'
To add an exception for this directory, call:

	git config --global --add safe.directory /root/tutorial
2023-04-09T13:57:08Z TRC got command result cmd="git branch" stack=/nginx
2023-04-09T13:57:08Z ERR running /usr/bin/git branch (at stack /nginx): exit status 128
2023-04-09T13:57:08Z FTL one or more commands failed
```

Maybe you know about this issue already but it's being some time since I last played with docker + host cloned repositories and then I didn't know that.

So other than the solution showed in the error message (which has security implications), the correct way of executing the container (if you need to run git commands in the repositories in the host machine) is:

```
docker run -it -u $(id -u ${USER}):$(id -g ${USER}) \
    -v /some/repo:/some/repo \
    ghcr.io/mineiros-io/terramate:latest -C /some/repo --changed run -- cmd
```

The `-u` is the important one, as it's going to execute the command inside the container using the mapped host user and group provided.

I think we should drop some words related to that in the docs but I can do that in a separate PR if needed.

Maybe we can also make the user's life easier if we support an environment variable that could just set the `safe.directory` globally, for users that just want this as default behavior.

i4ki (approved):
lgtm

mariux:
closing, we can discuss the need at a later stage

PR #905, opened 2023-04-12
Set up new Documentation Website with Vitepress

soerenmartius:
# Reason for This Change

Restructures and adds new documentation page deployed at https://terramate.io/docs/cli/.
After deployment, the page will be available at https://terramate.io/ also


docs/.vitepress/config.ts
```diff
@@ -0,0 +1,168 @@
...
+function getPath(path: string) {
+  const uri = path.replace(/(?:(^|\/)index)?\.md$/, '$1')
+
+  return uri === 'index' ? '' : uri
+}
+
+// https://vitepress.dev/reference/site-config
+export default defineConfig({
+  title: 'Terramate',
+  description: 'A VitePress Site',
```
i4ki, line 29:
```suggestion
  description: 'The Terramate Documentation',
```
It's [overridden](https://vitepress.dev/reference/site-config#description) by frontmatter but having a default makes sense in case it's not defined in some specific markdown.

i4ki (approved):

docs/.vitepress/config.ts
soerenmartius: makes sense will change in a follow up PR

PR #908, opened 2023-04-19
chore: document that map block can be used with lets

i4ki:
# Reason for This Change

The documentation didn't mention that `map` could be used inside `lets` block as well


soerenmartius (approved):
lgtm! thanks @i4ki <3

zied-elouaer (approved):

PR #910, opened 2023-04-24
chore: improve performance by caching the tm_regex() patterns.

i4ki:
# Reason for This Change

We identified that `tm_regex()` (using standard Terraform `regex()` function) is very slow because it compiles the regex pattern twice.
Then in this change we copied the Terraform implementation but cached the compiled patterns.


mariux (approved):
lgtm

PR #915, opened 2023-04-26
fix: stacks incorrectly selected when having same prefix wd prefix

### LINKED ISSUE #914: [BUG] Current directory scope limitation seems to be not working correctly
**Describe the bug**
In the (very nice) new [docs](https://terramate.io/docs/cli/orchestration/), it says:

> Terramate uses the current directory it is being executed to filter out stacks, ie., limit the scope of the execution, so if you execute terramate from the project's root directory, all stacks will be selected, and changing to inner directories in the project structure will select only stacks that are children of the current directory.

However, when I have two stacks named as follows:

* `foo/bar/stack-a`
* `foo/bar/stack-b`

If I change into `foo/bar/stack-a` (or use terramates `-chdir`), and do for example `terramate list`, I also see `stack-b` listed.

So, `stack-b` is being selected for execution, which I don't expect given my context is `stack-a`.

Perhaps I'm missing something?

**To Reproduce**

I've set up a repo with the above structure at https://github.com/brk3/terramate-demo

**Expected behavior**
I expected when I'm in a leaf node directory that terramate will only list that for execution.

**Log Output**
```
$ ls
modules          stacks           terramate.tm.hcl
$ terramate list
stacks/test
stacks/test-foo
$ cd stacks/test
$ terramate list
.
-foo
```

Also note that `test-foo` is truncated for some reason.

 - OS: MacOS
 - OS Version: Ventura
 - Git Version: git version 2.40.0
 - Terramate Version: 0.2.17

i4ki:
When matching stacks inside the working directory, the implementation checked for stack paths with the working directory prefix. The code was something like:

```go
if strings.HasPrefix(stackdir, wd) {
    // include stack in the list
}
```

it works for most cases but not if *working directory* is also the prefix of disjoint directories, see example stacks below:

```
/test
/tests
```
if *working dir* is `/test`, then both `/test` and `/tests` is a match.

# Changes

Append `/` to the directories so the matches are *workingDir* `/test/` against `/test/` and `/tests/` which only returns `/test/`.

Closes #914

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -1718,11 +1718,15 @@ func (c *cli) filterStacks(stacks []stack.Entry) []stack.Entry {
 }
 
 func (c *cli) filterStacksByWorkingDir(stacks []stack.Entry) []stack.Entry {
-	relwd := prj.PrjAbsPath(c.rootdir(), c.wd())
-
+	relwd := prj.PrjAbsPath(c.rootdir(), c.wd()).String()
+	if relwd != "/" {
+		relwd += string(filepath.Separator)
+	}
 	filtered := []stack.Entry{}
 	for _, e := range stacks {
-		if e.Stack.Dir.HasPrefix(relwd.String()) {
+		stackdir := e.Stack.Dir.String() + string(filepath.Separator)
+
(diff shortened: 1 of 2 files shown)


mariux (approved):
lgtm

PR #920, opened 2023-05-05
fix: file descriptor leak in the configuration tree load

i4ki:
# Reason for This Change

While profiling the Terramate execution I noticed a high number of file descriptors were kept open and never released.

The issue was in the `config.LoadTree()` which open a directory file descriptor for listing file entries and it was never closed. 

## Description of Changes

In the loading of configuration trees, the file descriptors are now closed in a defer.


mariux (approved):
lgtm

PR #922, opened 2023-05-05
chore: upgrade golangci-lint to v1.52.2 as it fixes some hangs

i4ki:
# Reason for This Change

The v1.52.2 also catches unused function arguments by default, then the code was also updated to comply.


soerenmartius (approved):
lgtm

mariux (approved):
lgtm

PR #931, opened 2023-05-17
Add more information to README.md

soerenmartius:
# Reason for This Change

Previously we highlighted technical features. We are changing this now to use-cases for our users to understand the value of Terramate. In general, this PR aims toward providing the audience with a better idea of what Terramate is and how what problems it can sovle.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -60,7 +60,7 @@ representative at an online or offline event.
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
-hello@mineiros.io.
+hello@terramate.io.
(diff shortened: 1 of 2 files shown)


kkmlr (approved):
Looks great. My only comment is minor - that I would be in favour of having the installation instructions in the README, maybe in the getting started section. I know they're linked in that guide, but I usually expect to see installation instructions in a README.

mariux (dismissed):

soerenmartius:
> Looks great. My only comment is minor - that I would be in favour of having the installation instructions in the README, maybe in the getting started section. I know they're linked in that guide, but I usually expect to see installation instructions in a README.

good point :) I'll make sure to address this in a follow up PR!

mariux (approved):

PR #939, opened 2023-05-26
chore: change LICENSE to MPL-2.0

i4ki:
License text obtained from: https://www.mozilla.org/media/MPL/2.0/index.txt
Now using only the SPDX license ID in each file.


mariux (approved):
lgtm

PR #947, opened 2023-05-26
Remove slack from releases since we moved to discord

soerenmartius:
We don't use slack anymore since we moved to Discord.


i4ki:
@soerenmartius can you please set [here](https://github.com/terramate-io/terramate/blob/main/.goreleaser.yaml#L59) to `false` as well?

soerenmartius:
> @soerenmartius can you please set [here](https://github.com/terramate-io/terramate/blob/main/.goreleaser.yaml#L59) to `false` as well?

sorry missed that! I just updated this now

i4ki (approved):

PR #955, opened 2023-06-07
docs: reviewed orchestration

Adeniyikayodee:
# Reason for This Change

change in the future).

## Description of Changes

Overview of what was changed.

### DIFF
--- a/docs/orchestration/index.md
+++ b/docs/orchestration/index.md
@@ -13,14 +13,13 @@ next:
 
 # Stacks Execution Orchestration
 
-The orchestration of stack's execution is driven by the `terramate run` command
+The orchestration of stack's execution is driven by the `terramate run` command,
 and it supports different ways of selecting stacks and configuring the order of
 execution.
 
 ## Stacks selection
 
-The **selection** defines which stacks from the whole set must be selected to
-execute and terramate provides three ways of configuring that:
+Choosing the right stacks for execution is a key part of Stacks Execution Orchestration. Terramate offers three ways to select stacks:
 
 1. Change detection
 
@@ -35,10 +34,7 @@ project structure will select only stacks that are children of the current direc
 
 3. Explicit `wants` relationship.
 
-The `wants` attribute of the stack block defines an explicit relationship
-between a stack and its list of wanted stacks, that when provided it says
-that when a stack is selected all the stacks listed on its `wants` list will also be
-selected, always, independent of any other selection criteria.
+The `wants` attribute of the stack block defines an explicit relationship between a stack and its list of wanted stacks. When a stack is selected, all the stacks listed on its `wants` list will also be selected, independent of any other selection criteria.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


docs/orchestration/index.md
```diff
@@ -56,68 +52,39 @@ applied is: `change detection`, `current directory`, `wants`.
 
 ## Stacks ordering
 
-Sometimes stacks are completely independent of each other, but on
-certain occasions it may happen that infrastructure that is created
-by **stack-a** is required by **stack-b**, like using the outputs
```
soerenmartius, line 61: @katcipis @mariux is this actually correct? I thought we never use outputs of one stack in another?

soerenmartius (approved):
lgtm

docs/orchestration/index.md
mariux: we ar enot using the outputs in terramate but people can query then with remote state..

docs/orchestration/index.md
```diff
@@ -56,68 +52,39 @@ applied is: `change detection`, `current directory`, `wants`.
...
 
-**after** ensures the opposite, that the stacks contained in the provided
-directories are executed before the current stack. As the stack, you are saying:
-"I execute after all stacks inside these directories".
+1. Choose the stack directory you want to change it's order of execution
+2. Open your Terramate configuration file `stack.tm.hcl` in a text editor.
+2. Select the stack block you want to allocate a specific order.
+3. Depending on your needs, add either the **before** or **after** field inside the stack block.
+4. Add **string** values representing the directory paths you want to order execution for in the **before** or **after** field.
+5. Save and close the file.
```
i4ki, line 87: Just my opinion, feel free to ignore.

This kind of explanation stands out from the rest of the documentation. For all other configurations, we never say "open the file, add the field, save and close the file" ...
This is a configuration language after all, so users can understand if we just say:

```
This can be configured by adding `before` and `after` fields to the stack block. Something like:

stack {
    before = [
        "/other/stack"
    ]
}
```

docs/orchestration/index.md
```diff
@@ -159,14 +126,14 @@ stack {
...
 ```
 
-For both equivalent configurations, the order of execution will be:
+In both configurations, the execution sequence will be:
 
 * stack-a
 * stack-b
 
-The same order of execution can be defined using **before**:
+You can also use the **before** field to define the execution order. For instance:
```
i4ki, line 134: The word "same" is important here, as the same final execution order is achieved by using `before` instead of `after`.

docs/orchestration/index.md
```diff
@@ -56,68 +52,39 @@ applied is: `change detection`, `current directory`, `wants`.
...
-child one, and trying to make this using explicit **before** and **after** clauses
-will lead to a cycle error.
 
-References to directories can be relative to the stack being configured in the form:
+### Explicit Order Of Execution
 
-```
-../../dir
-```
+Terramate's Explicit Order of Execution feature allows you to designate a specific order for stack execution. This is done by defining **before** and **after** fields within the **stack** block, referring to other directories containing stacks.
```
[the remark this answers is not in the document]
Adeniyikayodee: I will rephrase that.

i4ki (approved):
LGTM but just added a suggestion to drop the "terraform apply" from the phrase, as the Terramate command is just `terramate run`.

soerenmartius (approved):
lgtm

PR #962, opened 2023-06-12
feat: terramate run --cloud-sync-deployment

i4ki:
Introduces the `--cloud-sync-deployment` flag which syncs the Terramate orchestration with Terramate Cloud.

It creates deployments and set the status of the execution of each stack. The supported statuses set by the CLI at the moment are: `pending`, `running`, `failed`, and `ok`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         run: ./bin/terramate experimental cloud info
 
       - name: check cloud sync
-        run: ./bin/terramate run --cloud-sync-deployment -- echo "Hello from Terramate CI"
+        run: ./bin/terramate run --disable-check-git-remote --cloud-sync-deployment -- echo "Hello from Terramate CI"
 
   gh_integration_test:
     name: GHA Integration Test
--- a/cmd/terramate/cli/cloud.go
+++ b/cmd/terramate/cli/cloud.go
@@ -6,12 +6,10 @@ package cli
 import (
 	"context"
 	"net/http"
-	"os"
 	"strings"
 	"time"
 
 	"github.com/golang-jwt/jwt"
-	"github.com/hashicorp/go-uuid"
 	"github.com/rs/zerolog/log"
 	"github.com/terramate-io/terramate/cloud"
 	"github.com/terramate-io/terramate/cmd/terramate/cli/cliconfig"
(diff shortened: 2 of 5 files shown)


cmd/terramate/cli/cloud.go
```diff
@@ -42,25 +64,168 @@ func credentialPrecedence(output out.O, clicfg cliconfig.Config) []credential {
...
+		fatal(
+			errors.E("requires 1 organization associated with the credential but %d found: %s",
+				len(orgs),
+				orgs),
+		)
+	}
+
+	c.cloud.run.meta2id = make(map[string]int)
+
+	if runid := os.Getenv("GITHUB_RUN_ID"); runid != "" {
```
mariux, line 89: Run ID is not an uuid and should not be used for the deployment as it does not change between runs in GitHub..

mariux (changes requested):
lgtm..

cmd/terramate/cli/cloud.go
i4ki: yes true. I was using this env to force a fixed deployment ID when using the fake server. I guessed this GH env was an UUID but yeah, just removed as it's out of the scope now.

.github/workflows/ci.yml
```diff
@@ -56,6 +56,9 @@ jobs:
       - name: check cloud info
         run: ./bin/terramate experimental cloud info
 
+      - name: check cloud sync
```
mariux, line 59: This won't work as trust is temporary and should not be persists as a treat here

mariux (changes requested):

cmd/terramate/cli/cloud.go
```diff
@@ -42,25 +64,168 @@ func credentialPrecedence(output out.O, clicfg cliconfig.Config) []credential {
...
+	// TODO(i4k): convert repoURL to Go-style module name (eg.: github.com/org/reponame)
+
+	var payload cloud.DeploymentStacksPayloadRequest
+	for _, s := range stacks {
+		tags := s.Tags
+		if tags == nil {
+			tags = []string{}
+		}
+		payload.Stacks = append(payload.Stacks, cloud.DeploymentStackRequest{
+			MetaID:          s.ID,
```
mariux, line 152: We will need to make the id caseinsensitive as a next step and breaking change..

.github/workflows/ci.yml
i4ki: As all Terramate tests talk just with the fake server, I just had this idea of validating if it can successfully sync a deployment with the real cloud services, similarly to the ªcheck cloud infoª job. Otherwise, it's easy to introduce a change which passes all the tests but is broken.

It's working now.
Do you have an alternative idea?

mariux: Should be only tested against fake server fmpov as the cloud will be reconfigured at any point

i4ki: by reconfigured, do you mean redeployed?

MordecaiMalignatus (approved):
This is really cool and impressive work, kudos. 

I don't really have any notes or comments, this seems all rock solid. Well done! :D

PR #964, opened 2023-06-13
feat: allow globs on import

### LINKED ISSUE #649: [FEATURE] Allow globs on import
It would be nice to use globs to import modules.

Since my current setup counts with many files on each "stack" that I have, I could easily wrap the file contents into a `generate_hcl` block and just import everything (as is) to my stacks instead of adding many import blocks.

example:
```
modules/
└── my_mod/
    ├── data.tm
    ├── iam.tm
    ├── locals.tm
    └── main.tm
```
```hcl
import {
  source = "/modules/my_mod/*.tm"
}
```

> I know that I can wrap everything into a single file or just add many import blocks, so it's a "nice to have" feature!

brk3:
This patch allows for glob patterns in an `import` block.

It resolves a common pain point of having to maintain a list of imports for every file in a root module, e.g.

```
import { source = "../modules/foo/a.tm.hcl" }
import { source = "../modules/foo/b.tm.hcl" }
import { source = "../modules/foo/c.tm.hcl" }
```

can be replaced with:

```
import { source = "../modules/foo/*.tm.hcl" }
```

# Reason for This Change

When adding files to a module, it can be easy to forget to include this in the stack imports. This patch enables a stack to more easily stay in sync with a module, and removes the burden of maintaining a list of imports.

Closes #649 

## Description of Changes

Adds simple globbing functionality within `handleImports`. The import logic is then moved inside a loop of found matches. (Note, the patch looks larger than it is - if you check the `Hide whitespace` button on the GitHub diff view it helps see the changes more easily 👍 )


mariux:
Hi @brk3, thanks for your contribution..

We are currently working on the contribution guidelines to accept external PRs and will already consider your PR as the first one to be accepted. So your timing was close to perfect ;) 

We want to have it ready for early july (so in 2-3 weeks, hopefully earlier) and we will get back to you asap.

brk3:
@mariux Thanks that sounds awesome! Happy to change this up where needed based on the outcome of the guidelines.

MordecaiMalignatus (approved):

soerenmartius:
Hi @brk3,

We have implemented your changes in https://github.com/terramate-io/terramate/commit/66cc3a7b4d21bb10f9bface75f592769e634a4ff in addition to some more tests. Thanks a lot for contributing to Terramate, we really appreciate it!

PR #966, opened 2023-06-16
Docs/code generation overview

Adeniyikayodee:
# Reason for This Change

change in the future).

## Description of Changes

Overview of what was changed.


docs/code-generation/index.md
```diff
@@ -21,17 +21,26 @@ each integration scenario in the best way possible.
...
+> This strategy generates code that is closely tied to and relative to the stack where it is defined.
+It provides access to a rich set of code generation features, including globals, metadata, functions, lets,
+and assertions. By using these features you can easily manipulate and transform data within stack context.
+>
+
+* [File generation](./generate-file.md) with `root` and `stack` context.
+ > With this strategy, you can generate code at the project root level, outside of specific stacks. It offers 
+ access to project metadata, functions, lets, and assertions. BY using this relatively broader context you 
+ can generate code that spans multiple stacks or interacts with data at the project level.
+ >
```
i4ki, line 34: "With this strategy, you can generate code at the project root level, outside of specific stacks."
This depends on the context, with `context=stack` it actually generates inside stacks (exactly same behavior of `generate_hcl`).
With this approach, it looks like choosing `generate_file` or `generate_hcl` are related to where you'd need the file generated, which is not the case, they are orthogonal concepts.
We have two generate methods (_hcl and _file) and we support 2 execution contexts (`stack` and `root`). It's just a limitation that at the moment the `generate_hcl` does not support `context=root` but this can change in the future.

"... or interacts with data at the project level." 
I don't get what this part refers to.

docs/code-generation/index.md
```diff
@@ -61,26 +69,27 @@ generate_file "/file.txt" {
 
 All code generation blocks use labels to identify the block and define where
 the generated code will be saved but they have different constraints depending
-on the [generation context](#generation-context).
+on the generation context.
```
i4ki, line 72: Isn't the link useful here?

docs/code-generation/index.md
```diff
@@ -61,26 +69,27 @@ generate_file "/file.txt" {
...
+### 1. `Stack` Context Labels Constraints:
 
-For `stack` context, the labels must follow the constraints below:
+* The labels are in relative path and they must adhere to the format `<dir>/<filename>` or simply `<filename>`, that represents a relative path.
+* Regardless of the operating system, labels should be defined using the forward slash `/`.
+* Labels should not contain `../`, as code generation is confined to the stack and cannot navigate outside.
+* Labels should not start with `./`.
+* Labels should not represent symbolic links.
+* Each label must be unique within the stack hierarchy for blocks with `condition=true`.
+* It can’t be the stack.
```
i4ki, line 82: I'm not sure the new text reads better. The labels word is repeated in each sentence, even though we are enumerating the label constraints.

And this one does not read well:
`- It can’t be the stack.`

Which stack?
Maybe "It's not a stack" or "It can't be a stack" reads better?

i4ki (changes requested):
I think the PR still needs to improve the wording accuracy a little bit.

PR #220, opened 2022-02-15
feat: add content block on generate_hcl 

katcipis:
This PRs changes the behavior of code generation using `generate_hcl`. Instead of generating code by doing:

```hcl
generate_hcl {
  terraform {
  }
}
```

All generated HCL must be inside the `content` block, like this:

```hcl
generate_hcl {
  content {
    terraform {
    }
  }
}
```

A full example:

```sh
#!/bin/bash

set -o nounset
set -o errexit

basedir=$(mktemp -d)

cd "${basedir}"

# We need the project root config since it is not git
cat > terramate.tm.hcl <<- EOM
terramate {
  config {
  }
}
EOM

mkdir -p stacks/stack-{1,2}
terramate stacks init stacks/stack-{1,2}

# Defining some code for all stacks
# Globals are defined on root and also on stacks, to showcase possibilities
# Inspired on: https://github.com/mineiros-io/terramate-blueprint-iac-github/tree/main/stacks/repositories/teamA
cat > stacks/terramate.tm.hcl <<- EOM
globals {
    tf_version = "1.1.3"
    gh_owner = "the-awesome-mineiros"
    gh_provider_version = "4.19.1"
}

generate_hcl "provider.tf" {
    content {
        provider "github" {
          owner = global.gh_owner
        }

        terraform {
          required_providers {
            github = {
              source  = "integrations/github"
              version = global.gh_provider_version
            }
          }
        }

        terraform {
          required_version = global.tf_version
        }
    }
}

generate_hcl "main.tf" {
  content {
      module "repositories" {
        source = "../../../modules/terramate-github-repositories"

        repositories        = global.repositories
        repository_defaults = global.repository_defaults
        member_map          = global.member_map
      }
  }
}
EOM

# Now add some per stack globals

cat > stacks/stack-1/terramate.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-1"
            visibility = "private"
        }
    ]
    repository_defaults = {
        config = "stack-1"
    }
    member_map  = {
        "first.lastname" = "powerpuffGirls69"
    }
}
stack {}
EOM

cat > stacks/stack-2/terramate.tm.hcl <<- EOM
globals {
    repositories = [
        {
            name = "stack-2"
            visibility = "public"
        }
    ]
    repository_defaults = {
        config = "stack-2"
    }
    member_map  = {
        "first.lastname" = "onePunchMan666"
    }
}
stack {}
EOM


echo "initial tree"
tree

echo
[...]

PR #976, opened 2023-06-23
fix: crash when having generate_hcl with no content block

i4ki:
I spent some time digging if we lost this test somehow when moving testcases around between packages, but we actually never had tests for a missing `content` block inside the `generate_hcl` block.
The `content` block was introduced here: https://github.com/terramate-io/terramate/pull/220/files

Then this fixes a bug that has existed since [v0.0.11](https://github.com/terramate-io/terramate/releases/tag/v0.0.11).

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -832,6 +832,11 @@ func parseGenerateHCLBlock(block *ast.Block) (GenHCLBlock, error) {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


MordecaiMalignatus (approved):
Huh! Good catch, sometimes the most obvious things are the hardest to think of.

PR #979, opened 2023-06-23
Fix config to always use darkmode as default mode instead of detecting the browsers default

soerenmartius:
# Reason for This Change

Previously the configuration of Vitepress was faulty the theme was always set to the users browser standard instead of setting it to dark mode per default.

## Description of Changes

This change moves `appearance: 'dark',` to the correct level in the configuration so that dark mode is always used as the default mode.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -85,6 +85,8 @@ export default defineConfig({
 
     return head
   },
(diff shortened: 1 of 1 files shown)


i4ki (approved):
lgtm

mariux (approved):
lgtm

PR #984, opened 2023-06-27
fix: only send the commitSHA if the repository is clean

i4ki:
# Reason for This Change

If the repository is not in a clean state, we shall not send the CommitSHA field of the stack object.

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -859,15 +859,15 @@ func debugFiles(files []string, msg string) {
 	}
 }
 
-func (c *cli) gitFileSafeguards(checks stack.RepoChecks, shouldAbort bool) {
+func (c *cli) gitFileSafeguards(shouldAbort bool) {
 	if c.parsedArgs.Run.DryRun {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -859,15 +859,15 @@ func debugFiles(files []string, msg string) {
 	}
 }
 
-func (c *cli) gitFileSafeguards(checks stack.RepoChecks, shouldAbort bool) {
+func (c *cli) gitFileSafeguards(shouldAbort bool) {
```
MordecaiMalignatus, line 862: Nitpick: plain boolean arguments are never really super great. In this case it seems inconsequential (there's two call-sites) so there's no real urgent need to change it. Constants would be nice (`PanicOnError = true` and `WarnOnError=false` or something), but again, it's a nitpick.

MordecaiMalignatus (approved):

cmd/terramate/cli/cli.go
i4ki: hm yes, I think it makes sense! I will try to improve this while doing the other PRs.

PR #994, opened 2023-06-30
feat: add support for selecting the organization using env var

i4ki:
The environment variable `TM_CLOUD_ORGANIZATION` can be used to select the organization where data will be sent to.

PR #995, opened 2023-06-30
feat: send detailed review request information

i4ki:


MordecaiMalignatus (approved):
This looking good, the main mechanisms have already been parcelled out and the rest here is mostly about adding a bit more information we already had. The rest seems to be covered in #994.

MordecaiMalignatus (approved):

PR #1002, opened 2023-07-04
feat: implement experimental init command.

i4ki:
Implements the command `terramate experimental init`.
For now, it requires a `--all-terraform` flag which initialize Terraform stacks.
In the future additional flags would be created to initialize Terragrunt, Kubernetes, etc.


zied-elouaer (approved):
Tested on iac-gcloud and works like a charm. Very nice job!

soerenmartius:
Really cool - this only is worth a product update (will compose something)

PR #1011, opened 2023-07-06
feat:  experimental ensure-stack-id command

i4ki:
This command generates a random UUID for the `stack.id` of each stack which does not define it.


mariux (approved):
lgtm

PR #1012, opened 2023-07-06
feat!: make stack.id case insensitive

i4ki:
**breaking change**: The handling of `stack.id` is now case-insensitive, which means that the ids `my-id`, `My-id`, `MY-id`, `MY-Id`, etc, all represent the same id.
The Terramate CLI is going to treat them as lowercase internally and in the cloud.

### DIFF
--- a/cloud/types.go
+++ b/cloud/types.go
@@ -5,6 +5,7 @@ package cloud
 
 import (
 	"bytes"
+	"strings"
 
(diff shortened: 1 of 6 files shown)


MordecaiMalignatus (approved):

soerenmartius (approved):
lgtm

PR #1014, opened 2023-07-06
fix: incorrect filesystem implicit order due to reloading of objects

i4ki:
# Reason for This Change

change in the future).

## Description of Changes

Overview of what was changed.

### DIFF
--- a/config/stack.go
+++ b/config/stack.go
@@ -292,9 +292,9 @@ func validateWatchPaths(rootdir string, stackpath string, paths []string) (proje
 func StacksFromTrees(root string, trees List[*Tree]) (List[*SortableStack], error) {
 	var stacks List[*SortableStack]
 	for _, tree := range trees {
-		s, err := NewStackFromHCL(root, tree.Node)
+		s, err := tree.Stack()
 		if err != nil {
-			return List[*SortableStack]{}, err
+			return nil, err
 		}
 		stacks = append(stacks, &SortableStack{s})
 	}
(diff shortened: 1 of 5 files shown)


i4ki:
hey @zied-elouaer 
can you check if this fix work?
Thanks for reporting it.

zied-elouaer (dismissed):
Tested on the previously broken use case in `iac-gcloud` and it works just fine now. Nicely done!

MordecaiMalignatus (approved):

config/stack.go
```diff
@@ -279,12 +279,12 @@ func validateWatchPaths(rootdir string, stackpath string, paths []string) (proje
 }
 
 // StacksFromTrees converts a List[*Tree] into a List[*Stack].
-func StacksFromTrees(root string, trees List[*Tree]) (List[*SortableStack], error) {
+func StacksFromTrees(trees List[*Tree]) (List[*SortableStack], error) {
 	var stacks List[*SortableStack]
 	for _, tree := range trees {
-		s, err := NewStackFromHCL(root, tree.Node)
+		s, err := tree.Stack()
```
i4ki, line 285: @terramate-io/cli-tooling 

Note the `NewStackFromHCL()` in the line above.
The problem is that while building the DAG, stacks can be reloaded from disk and then they loose the implicit "before" clauses added for the filesystem ordering.

wmalik (approved):
Looks good to me

wmalik (approved):

PR #1016, opened 2023-07-07
fix: handle OIDC trusted status

i4ki:
In the case of `Github OIDC` the organization status is `trusted` and it was not handled.


zied-elouaer (approved):
Tested and works well.

PR #1049, opened 2023-07-21
fix: benchmarks not running for some branch names

i4ki:
The benchmark workflow was using `GITHUB_HEAD_REF` and using it with the `benchcheck` tool for the diff.
The problem is that `benchcheck` invokes `go mod -json download github.com/terramate-io/terramate@<ref>` where ref is whatever ref Go supports... but Go does not support slashes in the ref.. then if someone creates a branch like `myname/myfeature` then the Go tools fails.


soerenmartius (approved):
Lgtm

PR #1071, opened 2023-07-24
Add documentationt or the run-env command

soerenmartius:
# Reason for This Change

Previously we didn't have any documentation for the `run env` command.

## Description of Changes

This PR adds a new documentation for the `run env` command.

### DIFF
--- a/docs/cmdline/run-env.md
+++ b/docs/cmdline/run-env.md
@@ -1,0 +1,31 @@
+---
+title: terramate run-env - Command
+description: With the terramate run-env command see all environment variables configured for stacks.
+
+# prev:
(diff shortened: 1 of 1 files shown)


docs/cmdline/run-env.md
```diff
@@ -0,0 +1,31 @@
+---
+title: terramate run-env - Command
```
MordecaiMalignatus, line 2: shouldn't this be different for experimental commands, to reflect the subcommand tree?

MordecaiMalignatus: fair enough yep

MordecaiMalignatus (approved):

PR #1075, opened 2023-07-26
Split import and glob import into two different examples

soerenmartius:
# Reason for This Change

The current example of import and glob imports is missleading and not clearly separated.

## Description of Changes

This splits up the example into two different examples which should be more clear to the audience.


mariux (approved):
lgtm

PR #1077, opened 2023-07-31
fix: allow comma in the --tags of the create command

i4ki:
# Reason for This Change

The `--tags` was missing support for separating values with comma.
Now both ways are supported:
```
terramate create --tags a,b ...
```
and
```
terramate create --tags a --tags b
```

similarly to the other list flags (`--import`, `--after`, `--before`).

The docs were updated accordingly.


mariux (approved):
lgtm

PR #1078, opened 2023-08-02
chore: improve git url normalization

i4ki:


cloud/git_test.go
```diff
@@ -66,14 +66,14 @@ func TestNormalizeGitURL(t *testing.T) {
 			normalized: "something else",
 		},
 		{
-			name:       "unrecognized ssh url",
+			name:       "unrecognized ssh url - THIS IS A BUG IN THE go-gh library",
```
i4ki, line 69: @mariux this is a bug that should never happen in the real world (I guess). Only if someone place an incorrect git URL in the remote configuration. 
The problem is that `git://` scheme does not support port numbers in the URL, in then the go-gh library parses it incorrectly. In order to use port number, the user must provide the `ssh://` protocol (or the `ssh+git://` protocol).

lennyterra (dismissed):
LGTM :+1:

cloud/git_test.go
mariux: hmmm can the go-gh library parse gitlab and co?

mariux (changes requested):
does this break all non GH urls?

cloud/git_test.go
i4ki: yes, just added tests covering that! Thanks

PR #1079, opened 2023-08-02
feat: sending of pull_request information when running locally

i4ki:
- This implementation does not rely on the `GITHUB_REPOSITORY` environment variable but parses the git remote URL.

- GITHUB_TOKEN is now optional (if not set, uses GH default rate limiting)


cmd/terramate/cli/cloud.go
```diff
@@ -220,14 +221,29 @@ func (c *cli) createCloudDeployment(stacks config.List[*config.SortableStack], c
 			logger.Debug().Msg("commit SHA is not being synced because the repository is dirty")
 		}
```
lennyterra, line 223: side note: this func is basically impossible for me to keep in my head and grasp

lennyterra (dismissed):
LGTM :+1:

mariux (approved):
lgtm

cmd/terramate/cli/cloud.go
i4ki: yes, I agree with you. The creation of deployment now has a lot of nuances and just quickly splitting this function would require the functions to receive lots of arguments and return multiple values. Also it has a small optimization to avoid re-parsing/canonicalizing the repository.
It definitely needs polishing. Thanks for pointing it out. I will try to refactor it later on.

MordecaiMalignatus (approved):

PR #1081, opened 2023-08-03
fix: use stack environ for the cmd lookpath

i4ki:
When executing the commands provided to `terramate run`, the actual path of the program was being resolved using Terramate's own `PATH` environment variable instead of the stack's environment (that could be changed using `terramate.config.run.env.PATH` configuration.

The fix ensures that the program to be invoked is resolved using the stack's environment.

### DIFF
--- a/cmd/terramate/e2etests/cmd/test/test.go
+++ b/cmd/terramate/e2etests/cmd/test/test.go
@@ -21,6 +21,11 @@ func main() {
 	}
 
 	switch os.Args[1] {
+	case "echo":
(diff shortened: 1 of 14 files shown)


mariux (approved):
lgtm

i4ki:
@soerenmartius I improve the docs regarding this behavior. Can you have a look?

i4ki:
@soerenmartius what to do here regarding the LICENSE?

mariux (approved):
lgtm

PR #1082, opened 2023-08-04
chore: make cloud host and IDP configurable

i4ki:


soerenmartius (commented):
Can we add this to the documentation as well please?

cmd/terramate/cli/cloud_localhost.go
```diff
@@ -10,7 +10,7 @@ import (
 	"os"
 )
 
-const cloudBaseURL = "http://localhost:3001"
+const cloudDefaultBaseURL = "http://localhost:3001"
```
mariux, line 13: do we still need this special case now we can set it?

mariux (approved):
lgtm

cmd/terramate/cli/cloud_localhost.go
i4ki: yes but when running the e2etests I provide the `-tags localhostEndpoints` to build terramate pointing to localhost in all kinds of integration (atm just checkpoint API and TMC). We can do the same exporting the ENV vars in the tests... but I wasn't sure which approach is better. Wdyt?

i4ki: because still for checkpoint API, the build  tag is needed as we don't have other undocumented env var for it.

MordecaiMalignatus (approved):

PR #1089, opened 2023-08-07
fix: send pull request HEAD sha instead of current HEAD.

i4ki:
# Reason for This Change

When using the `--cloud-sync-deployment`, the `review_request` field was being populated with the current HEAD `sha` instead of the associated pull request HEAD sha.


mariux (dismissed):
lgtm

PR #1096, opened 2023-08-08
chore: remove reference to old terramate-ls repository

i4ki:


soerenmartius (approved):
lgtm

@i4ki can we add the language server to the documentation? (if we don't have that already)

PR #1117, opened 2023-08-23
Fix(create/all-terraform): Add in Name/Description in new stacks.

MordecaiMalignatus:
# Reason for This Change

When we run `terramate create --all-terraform`, we do not fill the `name` and
`description` fields. This is different behaviour than `terramate create
/path/to/stack`, which will use `stack` as value for both name and description.

## Description of Changes

When running `terramate create --all-terraform`, created stacks now have the
same derived names and descriptions as stacks created without the
`--all-terraform` flag.


mariux (dismissed):
lgtm please add change to changelog.md before release

mariux (dismissed):

mariux (approved):

PR #1127, opened 2023-09-07
Fix(cloud): typo: Context-Type → Content-Type

MordecaiMalignatus:
We used to not check for content type in the backend, we now do, therefore
requests fail now because of this. We're adding a middleware to the backend that
translates `Context-Type` to `Content-Type` for backwards compatability.


i4ki (approved):
LGTM

:man_facepalming: 

thanks

PR #1135, opened 2023-09-13
chore(tests): disable signal tests on MacOS

i4ki:
I removed the runner `runFixture` dependency from the test tables of tests that do not involve signals, then because of that some testcases had to be adjusted.


MordecaiMalignatus (dismissed):

lennyterra (approved):
reapproving after rebase

PR #1158, opened 2023-09-22
chore(tmc): send additional github integration data

i4ki:
These fields were already sent in the `review_request` object with the exception of the `github_pull_request_state`. But now that we also sync drifts, they were missing in the *drift metadata* because review_request is not sent in this case.


lennyterra (approved):
LGTM :+1:

kaynetik (approved):

PR #1166, opened 2023-09-26
fix(tmc): avoid creating a deployment if no stacks were selected.

i4ki:
# Reason for This Change

If no stacks are selected, the TMC deployment should not be created.
The `/v1/deployments` was returning `400 Bad Request` because a payload containing `{"stacks": null}` is rejected by the API.

This issue has existed since v0.3.1 (the first release with `--cloud-sync-deployment`).

## Description of Changes

Return early if no stack is selected for deployment.

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -104,6 +104,11 @@ func (c *cli) runOnStacks() {
 		return
 	}
 
+	if len(orderedStacks) == 0 {
+		logger.Debug().Msg("no selected stacks")
+		return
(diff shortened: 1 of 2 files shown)


soerenmartius (approved):
lgtm

soerenmartius:
Should this be added to the `CHANGELOG.md` also?

zied-elouaer (approved):

i4ki:
@soerenmartius closing as a non-issue. This was actually a backend mishandling.

PR #1177, opened 2023-10-12
chore(tmc): use real terraform in e2e tests

i4ki:
## Reasons for this change

Before we used a fake terraform binary that mimicked some Terraform behavior and it supported fully customization of the program's output.
The problem is that the fake terraform misses a lot of important details:
- `terraform show` gives different error messages depending on different requirements not met.
- `terraform show` requires an initialized terraform module and then mimicking this behavior is hard.
- `terraform show` checks the lock files
- etc

if we ignore those details, we miss the opportunity to test those cases and improve Terramate error messages for such errors.
_The error message improvements are not addressed in this PR_. 

## Changes

- Now during the e2e test setup, the Terraform is installed in a temp directory using the hashicorp official `hc-install` library.
- Tests were updated to use a real project.
- Sandbox library updated to support `copy` and `run` commands.


test/sandbox/sandbox.go
```diff
@@ -641,34 +658,82 @@ func buildTree(t testing.TB, root *config.Root, layout []string) {
...
+			genStackFile(param1, param2)
+		case "f:", "file:":
+			test.WriteFile(t, rootdir, param1, param2)
+		case "copy:":
+			assert.NoError(t, fs.CopyAll(
+				filepath.Join(rootdir, param1),
+				param2,
+			))
+		case "run:":
+			cmd := exec.Command("sh", "-c", param2)
```
snakster, line 693: Do we assume sh is always present, even on Windows? If not, could make this more portable in the future.

snakster (approved):

test/sandbox/sandbox.go
i4ki: nice catch! no we should not assume that! I forgot to improve this. Thanks

snakster (approved):

PR #1181, opened 2023-10-13
feat: add config setting for cloud organization

snakster:
# Reason for This Change

When being member of multiple cloud organizations, the user had to set the environment variable TM_CLOUD_ORGANIZATION to the name of a specific one when syncing cloud data.
In many cases it would be more convenient to set this automatically per project via configuration.

## Description of Changes

A new config setting has been added to the terramate config block. Usage example:
```
terramate {
  config {
    cloud {
      organization = "my-org-name"
    }
  }
}
```
TM_CLOUD_ORGANIZATION still works as before and overrides the config value if set.

### DIFF
--- a/cmd/terramate/e2etests/run_cloud_config_test.go
+++ b/cmd/terramate/e2etests/run_cloud_config_test.go
@@ -1,0 +1,200 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package e2etest
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/terramate-io/terramate/cloud"
+	"github.com/terramate-io/terramate/cloud/testserver"
+	"github.com/terramate-io/terramate/cmd/terramate/cli/clitest"
+	"github.com/terramate-io/terramate/test/sandbox"
+)
+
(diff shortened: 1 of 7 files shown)


cmd/terramate/e2etests/run_cloud_config_test.go
```diff
@@ -0,0 +1,203 @@
...
+			s.Git().CommitAll("created stacks")
+			env := removeEnv(os.Environ(), "CI")
+
+			for k, v := range tc.customEnv {
+				env = append(env, fmt.Sprintf("%v=%v", k, v))
+			}
+
+			tm := newCLI(t, s.RootDir(), env...)
+
+			tm.loglevel = zerolog.WarnLevel.String()
```
i4ki, line 193: Is this line still needed? We usually do not test on the `WRN` level and looking at the test cases, it seems we are only testing the `fatal` cases.

i4ki (approved):
LGTM! Well done!!!

Are you going to update the changelog and docs in a follow-up PR?

cmd/terramate/e2etests/run_cloud_config_test.go
snakster: You're right, I copied it over from another test case, but it doesn't seem to be needed here.

i4ki (approved):
LGTM

PR #1182, opened 2023-10-16
feat(tmc): send logs of the user-supplied command to Terramate Cloud.

i4ki:
# Reasons for This Change

Terramate Cloud supports visualizing the deployment logs in real-time.

## Description of the changes

A log syncer was created that proxies stdout/stderr writes using an in-memory pipe. The syncer creates 3 goroutines (1 for stdout, 1 for stderr and 1 for synchronizing logs) that *MUST* be shutdown/terminated after execution of each stack.
The syncer supports a customizable `batchSize` and `maxLineSize` but the requirement of the latter will be fixed in a separate PR.

### DIFF
--- a/cloud/log_syncer.go
+++ b/cloud/log_syncer.go
@@ -1,0 +1,203 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package cloud
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"sync"
+	"time"
+
+	"github.com/rs/zerolog/log"
+	"github.com/terramate-io/terramate/errors"
+)
+
+type (
+	// LogSyncer is the log syncer controller type.
+	LogSyncer struct {
+		pending      DeploymentLogs
+		fds          []io.Closer
+		in           chan *DeploymentLog
+		lastEnqueued time.Time
+		syncfn       Syncer
+		wg           sync.WaitGroup
+		shutdown     chan struct{}
+
+		maxLineSize  int
+		batchSize    int
+		idleDuration time.Duration
+	}
+
+	// Syncer is the actual synchronizer callback.
+	Syncer func(l DeploymentLogs)
+)
+
+// DefaultLogMaxLineSize is the default maximum line.
+// TODO(i4k): to be removed.
+const DefaultLogMaxLineSize = 4096
+
+// DefaultLogBatchSize is the default batch size.
+const DefaultLogBatchSize = 256
+
+// DefaultLogIdleDuration is the maximum idle duration before a sync could happen.
+const DefaultLogIdleDuration = 1 * time.Second
+
+// NewLogSyncer creates a new log syncer.
+func NewLogSyncer(syncfn Syncer) *LogSyncer {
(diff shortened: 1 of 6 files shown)


cloud/log_syncer.go
```diff
@@ -0,0 +1,203 @@
...
+			s.pending = s.pending[rest:]
+		}
+		s.shutdown <- struct{}{}
+	}()
+}
+
+func (s *LogSyncer) enqueue(l *DeploymentLog) {
+	s.pending = append(s.pending, l)
+	if len(s.pending) >= s.batchSize ||
+		(len(s.pending) > 0 && time.Since(s.lastEnqueued) > s.idleDuration) {
```
snakster, line 153: Do I understanding correctly that the buffered logs are only sent after either idle time or batch size are exceeded? If yes, that means at worst case there could be a ~256s delay if a command keeps adding output ~every second. Even without the worst case, for commands that keep producing output it would probably mean that syncing is delayed until they're done.
If we want to provide a "live" view of the output in the cloud UI, we could consider using a fixed interval instead.

cloud/log_syncer.go
```diff
@@ -0,0 +1,203 @@
...
+	syncfn Syncer,
+	maxLineSize int,
+	batchSize int,
+	idleDuration time.Duration,
+) *LogSyncer {
+	if maxLineSize == 0 {
+		panic("max line size must be set")
+	}
+	l := &LogSyncer{
+		in:       make(chan *DeploymentLog, batchSize),
```
snakster, line 63: I think the buffer size of this channel should be independent of the batch size? The consumer takes values out as soon as they are available and adds them to pending, so there isn't really any relation to the batch size.

cloud/log_syncer.go
snakster: As a follow-up, here's an idea for an implementation that syncs in a fixed interval:
```
func (s *LogSyncer) start() {
	go func() {
		ticker := time.NewTicker(500 * time.Millisecond)

		defer ticker.Stop()
		defer func() { s.shutdown <- struct{}{} }()

		for range ticker.C {
			next, done := s.fetchNext()
			if len(next) > 0 {
				s.syncfn(next)
			}
			if done {
				return
			}
		}
	}()
}

func (s *LogSyncer) fetchNext() (next DeploymentLogs, done bool) {
	for {
		select {
		case m, ok := <-s.in:
			// Keep taking elements from the queue ...
			if ok {
				next = append(next, m)

				// ... until batch size is exceeded
				if len(next) >= s.batchSize {
					return next, false
				}
			} else {
				// ... or until the queue has been closed
				return next, true
			}

		default:
			// ... or until there are no more elements for now
			return next, false
		}
	}
}
```

i4ki: As aligned in other channels, it definitely makes sense to sync at regular intervals so the user have a nicer UI feedback to the command output, but we will improve this behavior in a separate PR in next release.

Thanks @snakster

cloud/log_syncer.go
i4ki: yes but I feel that there's a good relationship between the batch size and the channel throughput, but just subjective at this point. As discussed before, this value only affects performance, as it could be even zero (unbuffered) without any issues other than performance.
I will come back to it later, in upcoming PRs.

Thanks @snakster

snakster (approved):

PR #1183, opened 2023-10-17
chore: add make target to run tests within docker

snakster:
# Reason for This Change

On Macos and Windows, running the tests is slower by a factor of 3-4 when compared to Linux, which has a negative impact on the development workflow on these systems.
When running the tests in a Docker container on the other hand, the performance is comparable to native Linux.

## Description of Changes

A Dockerfile and make target are added to run the tests with Docker build. Usage:
`make docker/test`

Currently, only the e2etests are executed, because when running terramate as root within the container, some filesystem related tests break. The one e2etest that breaks this way has been adapted


i4ki (approved):
LGTM

PR #1185, opened 2023-10-18
 chore(tests): allow running tests with different version of Terraform.

i4ki:
# Reason for This Change

Add support for running e2e tests with a different version of Terraform when exporting the `TM_TEST_TERRAFORM_REQUIRED_VERSION` environment variable.

Additionally, the tests now detect if Terraform is already installed and then the detected version is used. If the detected version does not match the version provided in the environment variable, then it's installed as before.

Usage example:

```
$ which terraform
/usr/bin/terraform
$ terraform version
Terraform v1.5.0

# Run tests with existing Terraform
$ go test -v ./cmd/terramate/e2etests

# Run tests with a different version of Terraform
$ TM_TEST_TERRAFORM_REQUIRED_VERSION=1.3.0 go test ./cmd/terramate/e2etests
```

## Description of Changes

The e2etest setup was updated to support a configurable Terraform version and tests were updated to be independent of the Terraform version.


cmd/terramate/e2etests/main_test.go
```diff
@@ -146,12 +143,33 @@ func buildTerramate(goBin, projectRoot, binDir string) (string, error) {
 	return outBinPath, nil
 }
 
-func installTerraform() (string, *install.Installer, error) {
+func installTerraform() (string, func(), error) {
+	requireVersion := os.Getenv("TM_TEST_TERRAFORM_REQUIRED_VERSION")
+	tfExecPath, err := exec.LookPath("terraform")
+	if err == nil {
+		cmd := exec.Command("terraform", "version")
```
snakster, line 150: Could use `terraform version -json` here, since the human readable output might change at some point.

snakster (approved):

PR #1187, opened 2023-10-19
 chore(tmc): remove LogSyncer line length limit.

i4ki:
## Reasons for the changes

The first implementation of the LogSyncer had a line length limit
which amounted to the size of the scanner buffer.

### Description of changes

A custom implementation of `readLines()` was implemented without any
line length limit (other than available memory).

Signed-off-by: Tiago Natel <t.nateldemoura@gmail.com>


cloud/log_syncer.go
```diff
@@ -158,27 +151,58 @@ func (s *LogSyncer) enqueue(l *DeploymentLog) {
...
+		}
+
+		// line ending not found, continue reading.
 	}
-	// If we're at EOF, we have a final, non-terminated line. Return it.
-	if atEOF {
-		return len(data), data, nil
+}
+
+func readLine(r io.Reader) (line []byte, err error) {
```
snakster, line 188: I think this could be removed later, the readLines implementation looks universally better to me and it's not much more complicated, so I don't think we will often need this benchmark in the future.

snakster (approved):

PR #1189, opened 2023-10-20
fix(tmc): panic due to index out of range in logsyncer

i4ki:
## Reasons For This Change

When providing an stdout/stderr containing an empty line, the log syncer crashes:

```
panic: runtime error: index out of range [-1]

goroutine 213 [running]:
github.com/terramate-io/terramate/cloud.dropCRLN(...)
        /vagrant/terramate/cloud/log_syncer.go:192
github.com/terramate-io/terramate/cloud.(*LogSyncer).NewBuffer.func1()
        /vagrant/terramate/cloud/log_syncer.go:103 +0x6fd
created by github.com/terramate-io/terramate/cloud.(*LogSyncer).NewBuffer
        /vagrant/terramate/cloud/log_syncer.go:80 +0x2a5
FAIL    github.com/terramate-io/terramate/cloud 0.055s
FAIL
```

### Description of Changes

A bounds check is added before removing the CRLN from the line.

### DIFF
--- a/cloud/log_syncer.go
+++ b/cloud/log_syncer.go
@@ -181,15 +181,18 @@ func scanLines(data []byte, atEOF bool) (advance int, token []byte, err error) {
 	return 0, nil, nil
 }
 
-// dropCRLN drops a terminal \r from the data.
+// dropCRLN drops a terminating \n and \r from the data.
 func dropCRLN(data []byte) []byte {
+	data = dropByte(data, '\n')
+	data = dropByte(data, '\r')
+	return data
+}
+
+func dropByte(data []byte, b byte) []byte {
 	if len(data) == 0 {
 		return data
 	}
-	if data[len(data)-1] == '\n' {
-		data = data[0 : len(data)-1]
-	}
-	if data[len(data)-1] == '\r' {
+	if data[len(data)-1] == b {
 		data = data[0 : len(data)-1]
 	}
 	return data
--- a/cloud/log_syncer_test.go
+++ b/cloud/log_syncer_test.go
@@ -114,6 +114,26 @@ func TestCloudLogSyncer(t *testing.T) {
 			},
 		},
 		{
+			name: "empty line -- regression check",
+			writes: []write{
+				{channel: cloud.StdoutLogChannel, data: []byte("\n")},
+			},
+			want: want{
+				output: map[cloud.LogChannel][]byte{
+					cloud.StdoutLogChannel: []byte("\n"),
+				},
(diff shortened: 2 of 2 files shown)


cloud/log_syncer.go
```diff
@@ -181,15 +181,18 @@ func scanLines(data []byte, atEOF bool) (advance int, token []byte, err error) {
 	return 0, nil, nil
 }
 
-// dropCRLN drops a terminal \r from the data.
+// dropCRLN drops a terminating \n and \r from the data.
 func dropCRLN(data []byte) []byte {
```
snakster, line 185: I think it would be better to write the output to `out` immediately so that it's independent of the line splitting:
```
func (s *LogSyncer) NewBuffer(channel LogChannel, out io.Writer) io.Writer {
	r, w := io.Pipe()
	mw := io.MultiWriter(out, w)

	s.fds = append(s.fds, w)
	s.wg.Add(1)
	go func() {
		defer s.wg.Done()
		line := int64(1)
		scanner := bufio.NewScanner(r)

		errs := errors.L()
		for scanner.Scan() {
			message := scanner.Text()

			t := time.Now().UTC()
			s.in <- &DeploymentLog{
				Channel:   channel,
				Line:      line,
				Message:   message,
				Timestamp: &t,
			}
			line++
		}
		if err := scanner.Err(); err != nil {
			errs.Append(errors.E(err, "scanning output lines"))
		}

		errs.Append(r.Close())
		errs.Append(w.Close())
		if err := errs.AsError(); err != nil {
			log.Error().Err(err).Msg("synchronizing command output")
		}
	}()
	return mw
}
```
This way you don't need the custom scanLines and dropCRLN, which simplifies the code and reduces the potential for bugs.

At least the tests (including the regression added in this PR) still work when I tried it like this.

i4ki: This PR targets `main` which still uses the `scanner` package, which calls `dropCRLN()` internally.
Anyway, in my other PR that get rid of `scanner`, it's also necessary to drop CRLN even if using the `io.MultiWriter()`.

I have some concerns regarding the `io.MultiWriter()` because the documentation states:

```
If a listed writer returns an error, that overall write operation stops and returns the error; it does not continue down the list.
```
So if it fails in the first write (`out` in your example) then it will not write to `w` (our pipe).

snakster: Fair enough, since the scanner approach is already deprecated anyway I'll mark it resolved.

snakster (approved):

PR #1198, opened 2023-10-23
feat(tmc): add git commit info to deployment metadata

snakster:
# Reason for This Change

The deployment metadata currently only queries the GitHub API. This change adds commit metadata for non-GitHub users as well.

## Description of Changes

The deployment metadata is extended with attributes for commit author, email, time, title and description. These attributes are queried from git directly, so they are always set. GitHub-specific metadata is still set as before.

For now, there are no e2etests, they will be added in a later PR.

### DIFF
--- a/cmd/terramate/cli/cloud.go
+++ b/cmd/terramate/cli/cloud.go
@@ -21,6 +21,7 @@ import (
 	"github.com/terramate-io/terramate/cmd/terramate/cli/out"
 	"github.com/terramate-io/terramate/config"
 	"github.com/terramate-io/terramate/errors"
+	"github.com/terramate-io/terramate/git"
 )
 
 const (
@@ -239,22 +240,16 @@ func (c *cli) detectCloudMetadata() {
 		Str("head_commit", c.prj.headCommit()).
 		Logger()
 
-	if c.prj.prettyRepo() == "local" {
-		logger.Debug().Msg("skipping review_request and metadata for local repository")
+	prettyRepo := c.prj.prettyRepo()
+	if prettyRepo == "local" {
+		logger.Debug().Msg("skipping review_request and remote metadata for local repository")
 		return
 	}
 
-	r, err := repository.Parse(c.prj.prettyRepo())
-	if err != nil {
-		logger.Debug().
-			Msg("repository cannot be normalized: skipping pull request retrievals for commit")
+	headCommit := c.prj.headCommit()
 
-		return
-	}
-
-	if r.Host != github.Domain {
-		return
-	}
+	c.cloud.run.metadata = &cloud.DeploymentMetadata{GitCommitSHA: headCommit}
+	md := c.cloud.run.metadata
 
 	defer func() {
 		if c.cloud.run.metadata != nil {
@@ -268,6 +263,28 @@ func (c *cli) detectCloudMetadata() {
 			logger.Debug().Msg("no provider metadata detected")
 		}
 	}()
+
+	if commit, err := c.prj.git.wrapper.ShowCommitMetadata("HEAD"); err == nil {
+		setDefaultGitMetadata(md, commit)
+	} else {
+		logger.Warn().
+			Err(err).
+			Msg("failed to retrieve commit information from GitHub API")
+	}
+
+	r, err := repository.Parse(prettyRepo)
+	if err != nil {
+		logger.Debug().
+			Msg("repository cannot be normalized: skipping pull request retrievals for commit")
+
+		return
+	}
+
+	if r.Host != github.Domain {
+		return
+	}
+
+	setGithubActionsMetadata(md)
 
 	ghRepo := r.Owner + "/" + r.Name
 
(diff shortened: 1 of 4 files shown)


cmd/terramate/cli/cloud.go
```diff
@@ -239,12 +240,26 @@ func (c *cli) detectCloudMetadata() {
...
+
+	if commit, err := c.prj.git.wrapper.ShowCommitMetadata("HEAD"); err == nil {
+		setDefaultGitMetadata(md, commit)
+	} else {
+		logger.Warn().
+			Err(err).
+			Msg("failed to retrieve commit information from GitHub API")
+	}
+
+	if prettyRepo == "local" {
```
i4ki, line 257: I think this check must be moved to the beginning of the function, where it was before, because AFAIK we are not interested in metadata for local commits as we cannot show them or make any sense out of them in the TMC frontend, because they reference a local bare upstream directory.

cmd/terramate/cli/cloud.go
```diff
@@ -287,138 +302,165 @@ func (c *cli) detectCloudMetadata() {
...
 
-	headCommit := c.prj.headCommit()
-	pulls, err := ghClient.PullsForCommit(ctx, ghRepo, headCommit)
-	if err != nil {
+	if pulls, err := getGithubPR(&ghClient, ghRepo, headCommit); err == nil {
+		if len(pulls) == 0 {
+			logger.Warn().
+				Msg("no pull request associated with HEAD commit")
+
+			return
```
i4ki, line 318: returning here will not sync *GithubActions* metadata when no pull request is associated with the commit.

git/git.go
```diff
@@ -727,6 +738,54 @@ func (git *Git) ListUncommitted(dirs ...string) ([]string, error) {
...
+	} else {
+		commitTime = nil
+	}
+
+	email := strings.TrimSpace(lines[2])
+	subject := strings.TrimSpace(lines[3])
+
+	body := ""
+	if len(lines) > 4 {
+		body = strings.TrimSpace(lines[4])
```
i4ki, line 777: Isn't this just getting the first line of the commit body?

git/git_test.go
```diff
@@ -1,13 +1,14 @@
 // Copyright 2023 Terramate GmbH
 // SPDX-License-Identifier: MPL-2.0
 
-//go:build linux
+//go:build linux || darwin
```
i4ki, line 4: I'm surprised this works :-)
Our tests generates a repository with a HEAD commit SHA that's known beforehand and this used to be an issue to work in non-linux envs.

git/git_test.go
```diff
@@ -325,6 +326,76 @@ func TestListRemoteWithMultipleBranches(t *testing.T) {
...
+		{
+			name:        "title-only",
+			title:       "add feature x",
+			description: "",
+		},
+		{
+			name:        "title with description",
+			title:       "add feature y",
+			description: "this is the latest feature",
+		},
```
i4ki, line 346: I think makes sense to add a case here with a multiline body (I think this is not covered).

i4ki (changes requested):
I left some comments for small improvements, but it looks good overall.

cmd/terramate/cli/cloud.go
snakster: Ah okay, I see. I actually thought about where to put this check and my thought was that metadata like commit message might still be useful, but only if local data from local repositories is relevant at all.

git/git.go
snakster: I'll add a test for this as suggested, but this works since I used strings.SplitN, not .Split, so lines[4] contains the remaining unsplit body.

git/git_test.go
snakster: Was not intentional, I had to change this locally to debug the tests and forgot to change it back... But we can leave it in I suppose if it now works on the CI? On my Mac all tests in this file work as well.

cmd/terramate/cli/cloud.go
snakster: Good catch, I looked through the previous version and I had actually not fully understood the intention behind the specific return points before. I changed it so the behaviour should be the same again.

cmd/terramate/cli/cloud.go
```diff
@@ -269,6 +264,28 @@ func (c *cli) detectCloudMetadata() {
...
+			Msg("repository cannot be normalized: skipping pull request retrievals for commit")
+
+		return
+	}
+
+	if r.Host != github.Domain {
+		return
+	}
+
+	setGithubActionsMetadata(md)
```
i4ki, line 287: This line still needs to be moved up there, maybe after line 252.

git/git_test.go
i4ki: yes we can test on OSX as well, cool.

i4ki (changes requested):
I think it's mostly done but I left a small improvement that will make it sync GithubActions metadata even if other integrations don't work.

i4ki (approved):
LGTM

i4ki:
@snakster please update the commit message because this PR apply to both drift and deployment metadata.

i4ki (approved):
LGTM

PR #1203, opened 2023-10-26
chore(docs): Improve documentation structure.

i4ki:
- Overall configuration docs was moved up in the navbar.
- Avoid links in the collapsable navbar item
  - Use a child "Overview" page when needed.
- Split "Sharing data" file into "overview", "globals" and "metadata" docs.
- Fix functions navbar and add an Overview item link
- **Lazy Evaluation** doc was improved.


soerenmartius (approved):
makes sense, thanks!

PR #1208, opened 2023-10-30
chore: improve running tests in docker.

i4ki:
The changes below were implemented:

- `make docker/test` -> `make test/docker`.

Reasoning: All test commands begin with `test`.

- Run `make test` in the container build instead of `go test ... cmd/terramate/e2etests`.

Reasoning: Convenience, so we can use a single command to check if everything works.

- Make rules sorted alphabetically.

Then it's easier to spot the rules when running just `make` or `make help`.

- In the e2e tests, builds terramate and helper with -buildvcs=off

Reasoning: In order to embed git meta information, Go needs a configured `git` client and file permissions must be correct. The VCS embedding is not needed in the e2e binaries, then disabling it is easier than making it work in the docker env.

### DIFF
--- a/cmd/terramate/e2etests/main_test.go
+++ b/cmd/terramate/e2etests/main_test.go
@@ -109,6 +109,7 @@ func buildTestHelper(goBin, testCmdPath, binDir string) (string, error) {
 	cmd := exec.Command(
 		goBin,
 		"build",
+		"-buildvcs=false",
 		"-o",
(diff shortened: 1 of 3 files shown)


i4ki:
@snakster if the changes downgrade your workflow, please let me know of the issues that we can try to improve. If not possible, I'm happy to use a separate make rule.

snakster (approved):

PR #1211, opened 2023-10-31
chore: make release-dry-run job optional.

i4ki:
Reasoning: it's a very slow job that runs in each branch push but unlikely to fail if `make build` works. Still useful to keep so we can receive alerts if it fails but it never did.


i4ki:
@wmalik @snakster

snakster (approved):

PR #1214, opened 2023-11-01
chore: fix linting issues in OSX.

i4ki:
Some testing helper methods are unused because signal tests are skipped in OSX.


i4ki:
@wmalik @snakster

PR #1219, opened 2023-11-07
feat: add clone support for child stacks

snakster:
# Reason for This Change

The experimental clone command did not work correctly with nested stacks. Before, it just treated them as normal files so the cloned directory would contain duplicated stack ids.

## Description of Changes

* `clone` now detects nested stacks and updates their stack ids if required.
* An flag `--skip-child-stacks` has been added to `clone` to ignore nested stacks. Files/directories within the parent stack that are _not_ nested stacks will still be cloned recursively.
* Since nested stacks are considered, clone can also be called on directories that are not stacks. 
* Output messages have been adapted to the new semantics.
* Docs have been updated.

Internal changes:
* `sandbox.AssertTree` helper has been added. It supports specs similar to `BuildTree` to assert that certain files or directories are present. This is less verbose than checking for this by hand and enables a unified pattern of `BuildTree(spec)`, `<modify tree with commands>`, `AssertTree(newSpec)`. This function can be extended in the future to support more specification directives.

### DIFF
--- a/cmd/terramate/e2etests/exp_clone_test.go
+++ b/cmd/terramate/e2etests/exp_clone_test.go
@@ -61,7 +61,7 @@ generate_hcl "test2.hcl" {
 	res := tmcli.run("experimental", "clone", srcStack, destStack)
 
 	assertRunResult(t, res, runExpected{
-		StdoutRegex: fmt.Sprintf("Cloned stack %s to %s with success\n", srcStack, destStack),
+		StdoutRegex: cloneSuccessMsg(1, srcStack, destStack),
 	})
 
 	destdir := filepath.Join(s.RootDir(), destStack)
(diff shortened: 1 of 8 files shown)


test/os.go
```diff
@@ -27,6 +28,46 @@ func TempDir(t testing.TB, base string) string {
 	return CanonPath(t, dir)
 }
 
+// IsNotExist calls os.Stat and asserts that the entry does not exist
+func IsNotExist(t testing.TB, dir, fname string) {
```
wmalik, line 32: nitpick:

```suggestion
// DoesNotExist calls os.Stat and asserts that the entry does not exist
func DoesNotExist(t testing.TB, dir, fname string) {
```

stack/clone.go
```diff
@@ -27,62 +28,115 @@ const (
...
+			if _, found := srcset[abspath]; found {
+				return false
+			}
+
+			return true
+		}
+
+		logger.Trace().Msgf("copying stack files from %q to %q", st.Srcdir, st.Destdir)
+
+		if err := fs.CopyDir(st.Destdir, st.Srcdir, filter); err != nil {
```
wmalik, line 124: I am wondering about the following edge case:

1. User starts the clone operation
2. The clone operations gets interrupted e.g. due to a file system error, Ctrl+C, crashing process etc.


In that case, would the user have a partially copied directory? 🤔 

I am wondering if we could copy the files to a temporary staging directory first and then execute a "rename" to make this whole process atomic. What do you think?

cmd/terramate/e2etests/exp_clone_test.go
```diff
@@ -93,3 +96,166 @@ generate_hcl "test2.hcl" {
...
+		tc := tc
+
+		t.Run(tc.Name, func(t *testing.T) {
+			layout := []string{}
+			for _, st := range tc.Stacks {
+				layout = append(layout, fmt.Sprintf("s:%s", st))
+			}
+			layout = append(layout, tc.Files...)
+
+			s := sandbox.New(t)
```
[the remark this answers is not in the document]
snakster: Normally, yes, since it doesn't require git. However, since I didn't add a terramate config the root, terramate will fail to run:
> FTL Project root not found. If you invoke Terramate inside a Git repository, Terramate will automatically assume the top level of your repository as the project root. If you use Terramate in a directory that isn't a Git repository, you must configure the project root by creating a terramate.tm.hcl configuration in the directory you wish to be the top-level of your Terramate project.```

I would leave it like this for now, but in general it would be nice to have a non-Git sandbox that works without having to manually add a config file to each testcase.

test/os.go
snakster: I iterated though a couple of names on this one during development :) `IsNotExist` is not proper English, but it's in line with Go's `os.IsNotExist()` and `os.ErrNotExist`.

wmalik: Ah I was not aware of that, however the stdlib `os.IsNotExist(err error)` has a different use case i.e. it checks if the provided `err` is an `ErrNotExist` see https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/os/error.go;l=90

test/os.go
```diff
@@ -27,6 +28,46 @@ func TempDir(t testing.TB, base string) string {
...
+	isDirOrFile(t, dir, fname, true)
+}
+
+// IsFile calls os.Stat and asserts that the entry is a file
+func IsFile(t testing.TB, dir, fname string) {
+	t.Helper()
+	isDirOrFile(t, dir, fname, false)
+}
+
+func isDirOrFile(t testing.TB, dir, fname string, isDir bool) {
```
wmalik, line 55: I am wondering if its worthwhile to import `stretchr/testify` considering it already implements these functions (and other common assertions), then we wouldn't need to maintain these helpers.

For example:
* https://github.com/stretchr/testify/blob/331c520966e81450aef3b50956ea4a5db5e0ccdf/require/require.go#L68
* https://github.com/stretchr/testify/blob/331c520966e81450aef3b50956ea4a5db5e0ccdf/require/require.go#L572
* https://github.com/stretchr/testify/blob/331c520966e81450aef3b50956ea4a5db5e0ccdf/require/require.go#L1378

However I understand that it's out of scope of this PR and would need refactoring of other tests.

snakster: Agreed, the less of this kind of code we have to write/maintain, the better.

stack/clone.go
snakster: I'm now just deleting the dest dir, since the code is defined right next to the check that it doesn't exist, IMO this is obvious enough.

cmd/terramate/e2etests/exp_clone_test.go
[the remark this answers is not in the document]
i4ki: The `sandbox.NoGit(t, true)`, creates the root config.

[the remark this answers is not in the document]
snakster: Ah, I see it now - hadn't rebased to latest main yet at the time of the comment.

i4ki (commented):
Amazing!!! 
just some suggestions for now. I will proper test it later today!

donovanrost:
On this, there's a comment:
> // - All files and directories are copied  (except dotfiles/dirs)
Would it be possible to copy dotfiles (not dirs)?

I use `.rtx.toml` to set environment variables according to the stack. It would be great if this file was copied as well. It's not a huge burden to manually create the file, but it would be a minor QoL improvement

i4ki:
@snakster please have a look at the sequence of commands below:
```
➜  test-stacks ~/go/bin/terramate list
➜  test-stacks ~/go/bin/terramate create stack-A
Created stack /stack-A
➜  test-stacks ~/go/bin/terramate create stack-A/child-stack
Created stack /stack-A/child-stack
➜  test-stacks ~/go/bin/terramate create stack-A/child-stack/nested
Created stack /stack-A/child-stack/nested
➜  test-stacks ~/go/bin/terramate list
stack-A
stack-A/child-stack
stack-A/child-stack/nested
➜  test-stacks ~/go/bin/terramate experimental clone stack-A stack-A/stack-A-copied
Cloned 3 stack(s) from stack-A to stack-A/stack-A-copied with success
Generating code on the new cloned stack(s)
Nothing to do, generated code is up to date
➜  test-stacks ~/go/bin/terramate list
stack-A
stack-A/child-stack
stack-A/child-stack/nested
stack-A/stack-A-copied
stack-A/stack-A-copied/child-stack
stack-A/stack-A-copied/child-stack/nested
➜  test-stacks ~/go/bin/terramate experimental clone stack-A stack-A/stack-A-copied/copied-again
Cloned 6 stack(s) from stack-A to stack-A/stack-A-copied/copied-again with success
Generating code on the new cloned stack(s)
Fatal failure preparing for code generation.
Error details: duplicated ID found on stacks: stack "/stack-A/stack-A-copied/copied-again/stack-A-copied/copied-again" and "/stack-A/stack-A-copied/copied-again" have same ID "7446cd5b-a354-40e9-af30-f6e8d68a03e6"
```

snakster:
> @snakster please have a look at the sequence of commands below:
> 
> ```
> ➜  test-stacks ~/go/bin/terramate list
> ➜  test-stacks ~/go/bin/terramate create stack-A
> Created stack /stack-A
> ➜  test-stacks ~/go/bin/terramate create stack-A/child-stack
> Created stack /stack-A/child-stack
> ➜  test-stacks ~/go/bin/terramate create stack-A/child-stack/nested
> Created stack /stack-A/child-stack/nested
> ➜  test-stacks ~/go/bin/terramate list
> stack-A
> stack-A/child-stack
> stack-A/child-stack/nested
> ➜  test-stacks ~/go/bin/terramate experimental clone stack-A stack-A/stack-A-copied
> Cloned 3 stack(s) from stack-A to stack-A/stack-A-copied with success
> Generating code on the new cloned stack(s)
> Nothing to do, generated code is up to date
> ➜  test-stacks ~/go/bin/terramate list
> stack-A
> stack-A/child-stack
> stack-A/child-stack/nested
> stack-A/stack-A-copied
> stack-A/stack-A-copied/child-stack
> stack-A/stack-A-copied/child-stack/nested
> ➜  test-stacks ~/go/bin/terramate experimental clone stack-A stack-A/stack-A-copied/copied-again
> Cloned 6 stack(s) from stack-A to stack-A/stack-A-copied/copied-again with success
> Generating code on the new cloned stack(s)
> Fatal failure preparing for code generation.
> Error details: duplicated ID found on stacks: stack "/stack-A/stack-A-copied/copied-again/stack-A-copied/copied-again" and "/stack-A/stack-A-copied/copied-again" have same ID "7446cd5b-a354-40e9-af30-f6e8d68a03e6"
> ```
Good catch! If the destination dir is within the source, it might happen that a later step copies files that were created during an earlier step of the same clone operation. It should be fixed now, though I was not able to reproduce this _exact_ error, since the test case is non-deterministic, probably since the results order of `ListDir()` isn't.

stack/clone.go
```diff
@@ -27,62 +28,133 @@ const (
...
+			if strings.HasPrefix(entry.Name(), ".") {
+				return false
+			}
+
+			abspath := filepath.Join(dir, entry.Name())
+			if _, found := stackset[abspath]; found {
+				return false
+			}
+
+			return true
```
i4ki, line 136:
```suggestion
			_, found := stackset[abspath]
			return !found

```

stack/clone.go
i4ki: I like @wmalik suggestion but in practice, there are no atomic renames/moves that crosses filesystem boundaries (on linux and other OSs). We implemented like this in vendoring feature and we got bug reports that it didn't work in lots of cases. See here:
https://github.com/terramate-io/terramate/pull/520

test/sandbox/sandbox.go
```diff
@@ -192,6 +194,46 @@ func (s S) BuildTree(layout []string) {
...
+func (s S) AssertTree(layout []string, opts ...AssertTreeOption) {
+	s.t.Helper()
+
+	o := &assertTreeOptions{}
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	assertTree(s.t, s.Config(), layout, o)
+}
```
i4ki, line 219: amazing :-)

i4ki (dismissed):
LGTM

I left a small suggestion but it's up to you, just a minor.

i4ki (approved):
LGTM

soerenmartius (approved):
LGTM!! Thank you

PR #1220, opened 2023-11-09
chore(tmc): send workflow name and ref in the metadata

i4ki:
Add 2 additional metadata when synchronising deployments or drifts with the _Terramate Cloud_.
- `metadata.github_actions_workflow_name`
- `metadata.github_actions_workflow_ref`

### DIFF
--- a/cloud/types.go
+++ b/cloud/types.go
@@ -202,6 +202,8 @@ type (
 		GithubActionsDeploymentTriggeredBy string `json:"github_actions_triggered_by,omitempty"`
 		GithubActionsRunID                 string `json:"github_actions_run_id,omitempty"`
 		GithubActionsRunAttempt            string `json:"github_actions_run_attempt,omitempty"`
+		GithubActionsWorkflowName          string `json:"github_actions_workflow_name,omitempty"`
+		GithubActionsWorkflowRef           string `json:"github_actions_workflow_ref,omitempty"`
 	}
 
 	// DeploymentReviewRequest is the review_request object.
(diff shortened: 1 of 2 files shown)


snakster (dismissed):

cloud/types.go
```diff
@@ -202,6 +202,8 @@ type (
 		GithubActionsDeploymentTriggeredBy string `json:"github_actions_triggered_by,omitempty"`
 		GithubActionsRunID                 string `json:"github_actions_run_id,omitempty"`
 		GithubActionsRunAttempt            string `json:"github_actions_run_attempt,omitempty"`
+		GithubActionsWorkflowName          string `json:"github_actions_workflow_name,omitempty"`
+		GithubActionsWorkflowPath          string `json:"github_actions_workflow_path,omitempty"`
```
mariux, line 206: This should be called ref as it is not a path but a complete reference.

mariux (dismissed):

cloud/types.go
i4ki: Thanks!

cloud/types.go
```diff
@@ -202,6 +202,8 @@ type (
 		GithubActionsDeploymentTriggeredBy string `json:"github_actions_triggered_by,omitempty"`
 		GithubActionsRunID                 string `json:"github_actions_run_id,omitempty"`
 		GithubActionsRunAttempt            string `json:"github_actions_run_attempt,omitempty"`
+		GithubActionsWorkflowName          string `json:"github_actions_workflow_name,omitempty"`
+		GithubActionsWorkflowRef           string `json:"github_actions_workflow_path,omitempty"`
```
snakster, line 206: Just to be sure, is using `github_actions_workflow_path` still intended here?

i4ki: what? hahaha no no thanks!

snakster (approved):

PR #1221, opened 2023-11-09
chore(tmc): refactor the testserver to support state between different features

i4ki:
## Reason for This Change

The e2e testserver didn't support keeping state between different endpoints.

## Description of Changes

A new `cloudstore` package was created with a API for updating an in-memory store representing the TMC data state.


cloud/testserver/cloudstore/store_test.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package cloudstore_test
```
snakster, line 4: I don't think we should spent too much time writing tests for tests, but it could be useful to have just one test that uses a fully pre-initialized cloudstore with data for every type and then validates that the values in all cloud.*Response types match.

i4ki: yes, I agree, just don't know yet the best of implementing this.

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
-
-	params := httprouter.ParamsFromContext(r.Context())
-	orguuid := params.ByName("orguuid")
-	deployuuid := params.ByName("deployuuid")
-
-	if !strings.HasPrefix(r.Header.Get("User-Agent"), "terramate/") {
-		w.WriteHeader(http.StatusBadRequest)
+// CreateDeployment is the POST /deployments handler.
+func CreateDeployment(store *cloudstore.Data, w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	defer func() { _ = r.Body.Close() }()
```
wmalik, line 40: This is not needed, see https://github.com/golang/go/blob/master/src/net/http/request.go#L180-L181

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
+	data, _ := io.ReadAll(r.Body)
```
wmalik, line 41: We could handler the `error` here?

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
+	var rPayload cloud.DeploymentStacksPayloadRequest
+	err := json.Unmarshal(data, &rPayload)
+	if err != nil {
```
wmalik, line 44: nitpick: I think this has the added benefit of not mistakenly reusing the `err` variable later in the function.
 
```suggestion
	if err := json.Unmarshal(data, &rPayload); err != nil {
```

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
 
-	if dhandler.deployments[orguuid] == nil {
-		dhandler.deployments[orguuid] = make(map[string]map[int64]cloud.DeploymentStackRequest)
-		dhandler.events[orguuid] = make(map[string]map[string][]string)
-	}
-	if dhandler.deployments[orguuid][deployuuid] == nil {
-		dhandler.deployments[orguuid][deployuuid] = make(map[int64]cloud.DeploymentStackRequest)
-		dhandler.events[orguuid][deployuuid] = make(map[string][]string)
+	err = rPayload.Validate()
+	if err != nil {
```
wmalik, line 51: nitpick
```suggestion
	if err := rPayload.Validate(); err != nil {
```

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
 
-	if strings.HasSuffix(r.URL.Path, "/events") {
-		if r.Method != "GET" {
-			w.WriteHeader(http.StatusMethodNotAllowed)
+	// deployment commit_sha is not required but must be present in all test cases.
+	// TODO(i4k): review this!!!
+	for _, st := range rPayload.Stacks {
+		if st.CommitSHA == "" {
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = w.Write([]byte(`commit_sha is missing`))
```
wmalik, line 64: It would be nice to handle the error here in case this ever fails, otherwise it would make the failing test hard to debug.

cloud/testserver/deployments.go
```diff
@@ -7,160 +7,151 @@ import (
...
 			w.WriteHeader(http.StatusInternalServerError)
 			writeErr(w, err)
 			return
 		}
-
-		err = p.Validate()
-		if err != nil {
-			w.WriteHeader(http.StatusBadRequest)
-			writeErr(w, err)
+		st, _ := store.GetStack(org, s.StackID)
```
wmalik, line 141: Would it make sense to handle the error here? 🤔

cloud/testserver/helpers.go
```diff
@@ -3,20 +3,38 @@
...
+}
+
+func marshalWrite(w io.Writer, obj interface{}) {
+	data, err := json.Marshal(obj)
+	if err != nil {
+		log.Printf("error: %v", err)
+		return
+	}
+	write(w, data)
 }
```
wmalik, line 40: Since this function only writes JSON, I guess it's safe to assume that all callers also want to set the content-type header, we could do this here instead of at the call-site.
```suggestion
func marshalWrite(w http.ResponseWriter, obj interface{}) {
	data, err := json.Marshal(obj)
	if err != nil {
		log.Printf("error: %v", err)
		return
	}
	w.Header().Set("Content-type", "application/json")
	write(w, data)
}
```

cloud/testserver/stacks.go
```diff
@@ -49,232 +76,234 @@ func (handler *stackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
 
-		handler.stacks[orguuid][stackid] = st
-		handler.statuses[orguuid][stackid] = st.Status
-		w.WriteHeader(http.StatusNoContent)
+	err = store.InsertDeploymentLogs(orguuid, stack.Stack.MetaID, deploymentUUID, logs)
+	if err != nil {
```
wmalik, line 289:
```suggestion
	if err := store.InsertDeploymentLogs(orguuid, stack.Stack.MetaID, deploymentUUID, logs); err != nil {
```

cloud/testserver/stacks.go
```diff
@@ -49,232 +76,234 @@ func (handler *stackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-		if _, ok := handler.stacks[orguuid]; !ok {
-			handler.stacks[orguuid] = make(map[int]cloud.StackResponse)
-		}
-		if _, ok := handler.statuses[orguuid]; !ok {
-			handler.statuses[orguuid] = make(map[int]stack.Status)
-		}
+	var logs cloud.DeploymentLogs
+	err = json.Unmarshal(bodyData, &logs)
+	if err != nil {
```
wmalik, line 283:
```suggestion
	if err := json.Unmarshal(bodyData, &logs); err != nil {
```

cloud/testserver/users.go
```diff
@@ -3,20 +3,60 @@
...
+	}
+	w.Header().Add("Content-Type", "application/json")
+	data, err := json.Marshal(user)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		write(w, data)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	write(w, data)
```
wmalik, line 34: We could use the `marshalWrite` helper here? 🤔

cmd/terramate/e2etests/run_cloud_config_test.go
```diff
@@ -118,45 +113,30 @@ func TestCloudConfig(t *testing.T) {
...
 			t.Parallel()
-			router := testserver.RouterWith(map[string]bool{
-				cloud.UsersPath:       true,
-				cloud.MembershipsPath: false,
-				cloud.DeploymentsPath: true,
-				cloud.DriftsPath:      true,
+			store, err := cloudstore.LoadDatastore(testserverJSONFile)
+			assert.NoError(t, err)
+
+			store.UpsertOrg(cloudstore.Org{
```
wmalik, line 119: TODO: Should we handle errors? 🤔

makefiles/unix.mk
```diff
@@ -38,8 +36,8 @@ test/helper:
 ## test code
 .PHONY: test
 tempdir=$(shell ./bin/helper tempdir)
-test: test/helper
-	TM_TEST_ROOT_TEMPDIR=$(tempdir) go test -race -count=1 ./...
+test: test/helper build
+	TM_TEST_ROOT_TEMPDIR=$(tempdir) ./bin/terramate run --no-recursive -- go test -race -count=1 ./...
```
wmalik, line 40: Could you please elaborate the user of `./bin/terramate` here? 🤔

testdata/testserver/test-cloud-orgs.tm
```diff
@@ -0,0 +1,34 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+# You can define multiple organizations in this file.
+#
+# The `name` field of the organizatio MUST match the `global.testserver.orgs.<name>`.
```
wmalik, line 6:
```suggestion
# The `name` field of the organization MUST match the `global.testserver.orgs.<name>`.
```

wmalik (commented):
Good work!

I only have minor comments and some questions.

testdata/example-stack/out.plan
wmalik, line 1: I noticed that we have `*.plan` in the `.gitignore` in a subsequent PR and not sure if this file was intentionally added here.

i4ki: good catch!!

i4ki:
Thanks @wmalik 

As explained, I will merge and address your concerns in a new PR.
I really appreciate the comments.

cloud/testserver/deployments.go
i4ki: Oh true. I had forgotten that!

wmalik:
> Thanks @wmalik 
> 
> As explained, I will merge and address your concerns in a new PR.
> I really appreciate the comments.

@i4ki sounds good

cloud/testserver/helpers.go
i4ki: I follow the practice of only setting headers and the status code from the main handler function, because then it's clear from the Handler func what comes out of the API in each returning branch. Then you don't need to memorize or enter each function to know if things are set, etc.

cloud/testserver/stacks.go
i4ki: sometimes inline error handling like that makes the code hard to read, when the funcall takes lots of arguments, and it trades readability for a small correctness benefit.
But I will see how it looks this way, thanks!

wmalik (approved):

cmd/terramate/e2etests/run_cloud_config_test.go
i4ki: the UpsertOrg() method does not fail (in-memory store).

PR #1223, opened 2023-11-10
feat(tmc): add support for additional stack status filters

wmalik:
# Reason for This Change

Terramate Cloud supports filtering stacks by various statuses. The CLI so far has only supported the `unhealthy` filter.

## Description of Changes

This PR adds support for the following filters (`--experimental-status`) when using `terramate list` and `terramate experimental trigger` commands:

* `ok`
* `drifted`
* `failed`
* `healthy`

### Notable changes

* Refactor type `FilterStatus` to decouple it from from the stack status


cloud/stack/filterstatus.go
```diff
@@ -0,0 +1,90 @@
...
+	"golang.org/x/exp/slices"
+)
+
+const (
+	// ErrInvalidFilterStatus represents an invalid filter stack status error.
+	ErrInvalidFilterStatus errors.Kind = "invalid filter stack status"
+)
+
+type (
+	// FilterStatus represents the query paramater used for filtering stacks via the API
```
i4ki, line 17:
```suggestion
	// FilterStatus represents the query parameter used for filtering stacks via the API
```

i4ki (commented):
I feel like with this implementation there's no need to have `stack.Status`, `drift.Status` and `deployment.Status` as integers anymore. So making them all strings will enable us to reuse the `stack.Status` code in the filter implementation. 

Example:

```
func NewFilterStatus(str string) (FilterStatus, error) {
    if str == "unhealthy" {
        return FilterStatusUnhealthy, nil
    }
    if str == "healthy" {
        return FilterStatusHealthy, nil
    }
    s, err := NewStatus(str)
    if err != nil {
        return "", errors.E(err, "unrecognized stack filter")
    }
    return s, nil
}
```

All of them were integers so they could be filtered in the testserver (and potentially in the cli implementation) using bitwise operations, avoid switch-cases, ifs, etc.

cloud/stack/filterstatus.go
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1223/commits/074f235065a2efea2d5f284c89912e7e5076b5ad will rebase when review is complete

wmalik:
> I feel like with this implementation there's no need to have `stack.Status`, `drift.Status` and `deployment.Status` as integers anymore. So making them all strings will enable us to reuse the `stack.Status` code in the filter implementation.
> 
> Example:
> 
> ```
> func NewFilterStatus(str string) (FilterStatus, error) {
>     if str == "unhealthy" {
>         return FilterStatusUnhealthy, nil
>     }
>     if str == "healthy" {
>         return FilterStatusHealthy, nil
>     }
>     s, err := NewStatus(str)
>     if err != nil {
>         return "", errors.E(err, "unrecognized stack filter")
>     }
>     return s, nil
> }
> ```
> 
> All of them were integers so they could be filtered in the testserver (and potentially in the cli implementation) using bitwise operations, avoid switch-cases, ifs, etc.

@i4ki  do you suggest to address that in this PR?

I am wondering about the advantages of reusing code between the different status types as compared to keeping them decoupled, mainly:
- is there a possibility that list of values for these status types will diverge in the future?
- `FilterStatus` currently only implements functionality around parsing/marshalling the `--experimental-status` flag,  I am wondering what would be the benefit of coupling it with other `*.Status` types that might have different validation/marshalling/unmarshalling constraints?

i4ki:
I didn't ask for reusing code between different status types.

> I feel like with this implementation there's no need to have stack.Status, drift.Status and deployment.Status as integers anymore. So making them all strings will enable us to reuse the stack.Status code in the filter implementation.

The stack filter implementation would reuse the stack statuses.

The other status types (deployment.Status and drift.Status) were designed to support the same filtering strategy (even thought it's not implemented/required yet) but I wanted to keep the consistency. So if a deployments filter is implemented in the future, the same approach would be used. But if now we switch to string comparison in the stack filter, then there's no need to have them handled as integers anymore.

So either keep the current design or go all into using a new type backed by an underlying string type (as you did before for `stack.Status`, if I'm not mistaken).

wmalik:
@i4ki 

> I didn't ask for reusing code between different status types.

Ah sorry about misunderstanding.

> 
> > I feel like with this implementation there's no need to have stack.Status, drift.Status and deployment.Status as integers anymore. So making them all strings will enable us to reuse the stack.Status code in the filter implementation.
> 
> The stack filter implementation would reuse the stack statuses.

There are some `stack.Status values` (`failed`,`canceled` and `unknown`) that are currently not allowed in the `GET /v1/stacks/:org_id?status` query parameter i.e. the list of possible `stack.Status` values is different than the list of possible `stack.FilterStatus` values. I am not sure how to address that with your suggested approach.

In short I am arguing for even keeping the `stack.Status` and `stack.FilterStatus` types+impl decoupled because they both seem to have slightly different semantics. If you agree to that, then I can also refactor the `{stack,drift,deployment}.Status` enums to use string based enums. What do you think?
 

> 
> The other status types (deployment.Status and drift.Status) were designed to support the same filtering strategy (even thought it's not implemented/required yet) but I wanted to keep the consistency. So if a deployments filter is implemented in the future, the same approach would be used. But if now we switch to string comparison in the stack filter, then there's no need to have them handled as integers anymore.
> 
> So either keep the current design or go all into using a new type backed by an underlying string type (as you did before for `stack.Status`, if I'm not mistaken).

i4ki:
> There are some stack.Status values (failed,canceled and unknown) that are currently not allowed in the GET /v1/stacks/:org_id?status query parameter i.e. the list of possible stack.Status values is different than the list of possible stack.FilterStatus values. I am not sure how to address that with your suggested approach.

The `canceled` is not a stack status anymore and can be removed. The `unknown` is valid only in the backend database, not a public status anymore, then it can also be removed from the cli.

In the end, the only valid stack statuses are: `ok`, `failed` and `drifted`.
The `failed` status is valid in the API and it can be used in the `GET /v1/stacks/:org?status=failed`.

I still think the stack statuses filters are a superset of the stack statuses.

wmalik:
See https://github.com/terramate-io/terramate/pull/1256

PR #1232, opened 2023-11-14
Add UUID behaviour to the stacks properties documentation

soerenmartius:
# Reason for This Change

When users read the documentation and stumble across the exposed properties of a stack, it's not clear that this defaults to auto-generated UUIDs v4.

## Description of Changes

Add this to the documentation and reference the commands also that use UUID generation.


i4ki (approved):
LGTM

PR #1235, opened 2023-11-15
fix: git change detection works on any commit

snakster:
# Reason for This Change

change in the future).

## Description of Changes

Overview of what was changed.


cmd/terramate/cli/project.go
```diff
@@ -119,35 +100,76 @@ func (p *project) remoteDefaultCommit() string {
...
+	//
+	//   - Historic changes are all other non-deployed and non-pending, i.e. commits from an already merged and deployed branch.
+	//     They should be compared to the fork point with origin/main.
+	//
+	//     Case 5: H has a fork point with origin/main -- a merged branch commit
+	gitcfg := p.gitcfg()
+	gw := p.git.wrapper
+
+	remoteDefaultBranchRef := p.remoteDefaultBranchRef()
+	headRev, _ := gw.RevParse("HEAD")
```
i4ki, line 130: I appreciate the intent of keeping the logic simple here!
Maybe you can put the error handling in a helper (memoized) function.
The error handling is required because there are cases where `HEAD` is not set.

i4ki: same for the other rev-parses

i4ki: The `rev-parse` outputs in the stdout even when an error happens.
Example:
````
➜  terramate git:(snk-git-change-detection) ✗ git rev-parse HEAD:README.md 2>/dev/null
0de873a28b4a0dfe3ef61d369db8085a11b48837
➜  terramate git:(snk-git-change-detection) ✗ echo $?
0
➜  terramate git:(snk-git-change-detection) ✗ git rev-parse HEAD:INVALID.md 2>/dev/null
HEAD:INVALID.md
➜  terramate git:(snk-git-change-detection) ✗ echo $?
128
```

I would do this as a defensive programming because we depend on the git version installed in the user end.

snakster: All of this error-ignoring relies on `Git.exec()` always returning an empty string in the error case:
```
stdout, err := cmd.Output()
if err != nil {
  stderr := []byte{}
  var exitError *exec.ExitError
  if errors.As(err, &exitError) {
	  stderr = exitError.Stderr
  }
  
  return "", NewCmdError(cmd.String(), stdout, stderr)
}
```
So while the executing rev-parse directly does return the error on stdout, `gw.RevParse` should not.

mariux (changes requested):
finally tested the PR, please check the notion document for comments on a case this PR now creates wrong results when calculating changes.

soerenmartius:
@mariux @snakster this needs to well documented also

PR #1241, opened 2023-11-16
fix(generate): blocks with context=root were ignored if defined in stacks.

i4ki:
# Reason for This Change

The `generate_file` blocks with `context=root` were ignored when defined inside stack directories. By the spec, they could be defined anywhere in the project.

## Description of Changes

Removed a check for stack directories when ignoring dirs.


wmalik (approved):
Looks good!

wmalik (approved):

PR #1248, opened 2023-11-20
feat(terramate-ls): implements terramate.createStack command.

i4ki:
The command supports the arguments below:

- `uri=`                target directory.
- `genid=`              any value, if set, generates an ID.
- `name=`               name of the stack.
- `description=`        description of the stack.


ls/commands.go
```diff
@@ -0,0 +1,104 @@
...
+			if err != nil {
+				return errors.E(ErrCreateStackFailed, err, "generating ID")
+			}
+			stackConfig.ID = id.String()
+		case "name":
+			stackConfig.Name = argVal
+		case "description":
+			stackConfig.Description = argVal
+		default:
+			panic("unreachable")
```
wmalik, line 94: Would this crash the language server? Would it make sense to return an `ErrUnrecognizedAttr` instead?

ls/ls.go
```diff
@@ -129,6 +135,9 @@ func (s *Server) handleInitialize(
 					IncludeText: false,
 				},
 			},
+			// ExecuteCommandProvider: &lsp.ExecuteCommandOptions{
```
wmalik, line 138: Is it intentionally commented out?

wmalik (commented):
Looks great, some minor comments/questions.

ls/commands.go
i4ki: yes, I added the panic while testing and forgot to improve. Thanks!

wmalik (approved):
Looks good!

PR #1249, opened 2023-11-20
feat(cloud): Give support to optional parameters in Content Type from tmc

sixstone-qq:
# Reason for This Change

With this change, the client is able to receive responses from the
cloud that contain optional parameters like `charset` in the
`Content-Type` header value.

## Description of Changes

Accepting optional parameters using `mime` package from standard library.

Change all test servers to return an optional parameter to mimic real behaviour.

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"net/http/httputil"
 	"os"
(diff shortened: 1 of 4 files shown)


sixstone-qq:
> The change makes sense but can you undo the testserver changes? I have big PRs refactoring the entire testserver, then the change touche files that will be gone once they are merged. I will update in my PRs to set the charset.

Thanks! I just removed that files from the changeset.

i4ki (approved):
LGTM

PR #1255, opened 2023-11-22
feat(scripts): evaluate script blocks

wmalik:
# Reason for This Change

The attributes of a script block need to be evaluated at runtime in order to resolve any variables/globals/functions etc.

## Description of Changes

* Added an `EvalScript` function that evaluates a script based on the provided context
* Added tests for validating types + evaluating functions/globals

### DIFF
--- a/config/script.go
+++ b/config/script.go
@@ -1,0 +1,187 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package config
+
+import (
+	hhcl "github.com/hashicorp/hcl/v2"
+	"github.com/terramate-io/terramate/errors"
+	"github.com/terramate-io/terramate/hcl"
+	"github.com/terramate-io/terramate/hcl/eval"
+	"github.com/terramate-io/terramate/hcl/info"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Errors for indicating invalid script schema
+const (
+	ErrScriptSchema              errors.Kind = "script config has an invalid schema"
+	ErrScriptInvalidTypeDesc     errors.Kind = "invalid type for script.description"
+	ErrScriptInvalidTypeCommand  errors.Kind = "invalid type for script.command"
+	ErrScriptInvalidTypeCommands errors.Kind = "invalid type for script.commands"
+)
+
+// ScriptJob represents an evaluated job block
+type ScriptJob struct {
+	Cmd  []string
+	Cmds [][]string
+}
+
+// Script represents an evaluated script block
+type Script struct {
+	Range       info.Range
+	Labels      []string
+	Description string
+	Jobs        []ScriptJob
+}
+
+// Commands is a convenience method for callers who don't specifically
+// care about which command attr was set e.g. command or commands. This method
+// returns a list of commands irrespective of whether they were set through
+// job.command or job.commands
+func (es ScriptJob) Commands() [][]string {
+	if es.Cmd != nil {
+		return [][]string{es.Cmd}
+	}
+	return es.Cmds
+}
+
+// EvalScript evaluates a script block using the provided evaluation context
+func EvalScript(evalctx *eval.Context, script hcl.Script) (Script, error) {
+	evaluatedScript := Script{
+		Range:  script.Range,
+		Labels: script.Labels,
+	}
+	errs := errors.L()
+
+	desc, err := evalScriptDesc(evalctx, script.Description.Expr, "script.description")
+	errs.Append(err)
+
+	evaluatedScript.Description = desc
+
+	for _, job := range script.Jobs {
+		evaluatedJob := ScriptJob{}
+
+		if job.Command != nil {
+			command, err := evalScriptJobCommand(evalctx, job.Command.Expr, "command")
+			if err != nil {
+				errs.Append(err)
+				continue
+			}
+			evaluatedJob.Cmd = command
+		}
+
+		if job.Commands != nil {
+			commands, err := evalScriptJobCommands(evalctx, job.Commands.Expr, "commands")
+			if err != nil {
+				errs.Append(err)
+				continue
+			}
+			evaluatedJob.Cmds = commands
+		}
+
+		evaluatedScript.Jobs = append(evaluatedScript.Jobs, evaluatedJob)
+	}
+
+	if err := errs.AsError(); err != nil {
+		return Script{}, err
+	}
+
+	return evaluatedScript, nil
+}
+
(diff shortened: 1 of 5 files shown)


config/script.go
```diff
@@ -0,0 +1,192 @@
...
+	Labels      []string
+	Description string
+	Jobs        []EvaluatedScriptJob
+}
+
+// CommandOrCommands is a convenience method for callers who don't specifically
+// care about which command attr was set e.g. command or commands. This method
+// returns a list of commands irrespective of whether they were set through
+// job.command or job.commands
+func (es EvaluatedScriptJob) CommandOrCommands() [][]string {
```
i4ki, line 39: Just call it `Commands()`. The evaluated data structure does not need to reflect the source code data structures. In practice, only the list of commands make sense.

config/script.go
```diff
@@ -0,0 +1,192 @@
...
+	evaluatedScript := EvaluatedScript{
+		Labels: script.Labels,
+	}
+	errs := errors.L()
+
+	desc, err := evalScriptDesc(evalctx, script.Description.Expr, "script.description")
+	if err != nil {
+		errs.Append(err)
+	}
+
```
i4ki, line 57:
```suggestion
	errs.Append(err)
```

config/script.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// EvalScript evaluates a script block using the provided evaluation context
+func EvalScript(evalctx *eval.Context, script hcl.Script) (EvaluatedScript, error) {
+	evaluatedScript := EvaluatedScript{
+		Labels: script.Labels,
+	}
+	errs := errors.L()
+
+	desc, err := evalScriptDesc(evalctx, script.Description.Expr, "script.description")
```
i4ki, line 53: we have `evalString()` in this same package.
https://github.com/terramate-io/terramate/blob/main/config/assert.go#L74-L86

config/script.go
```diff
@@ -0,0 +1,192 @@
...
+func evalScriptJobCommands(evalctx *eval.Context, expr hhcl.Expression, name string) ([][]string, error) {
+	val, err := evalctx.Eval(expr)
+	if err != nil {
+		return nil, errors.E(ErrScriptSchema, expr.Range(),
+			err, "evaluating %s", name)
+	}
+
+	if !val.Type().IsTupleType() {
+		return nil, errors.E(ErrScriptInvalidTypeCommands, expr.Range(),
+			"%s should be a list type", name)
```
i4ki, line 151:
```suggestion
			"%s should be a list of list(string) type", name)
```

config/script.go
```diff
@@ -0,0 +1,192 @@
...
+
+		evaluatedCommand := []string{}
+		indexCommand := -1
+		itCommand := elem.ElementIterator()
+		for itCommand.Next() {
+			indexCommand++
+			_, elem := itCommand.Element()
+			if elem.Type() != cty.String {
+				errs.Append(errors.E(ErrScriptInvalidTypeCommands, expr.Range(),
+					"field %s must be a list(string) but element %d has type %q",
```
i4ki, line 177:
```suggestion
					"field %s must be a string but element %d has type %q",
```

config/script_test.go
```diff
@@ -0,0 +1,308 @@
...
+					assert.EqualInts(t, len(wantJob.Commands), len(gotJob.Commands), "len(commands) mismatch")
+					for j, gotCommand := range gotJob.Commands {
+						wantCommand := wantJob.Commands[j]
+						assert.IsTrue(t,
+							slices.Equal(wantCommand, gotCommand),
+							fmt.Sprintf("command value mismatch: want[%#v], got [%#v]", wantCommand, gotCommand))
+					}
+				}
+
+			}
```
i4ki, line 304: As `EvaluatedScript` and `EvaluatedScriptJob` are simple types (structs with basic Go types), what do you think of just using `cmp.Diff()` (go-cmp):

```
if diff := cmp.Diff(expectedScript, gotScript); diff != "" {
    t.Fatal(diff)
}
```

i4ki (commented):
It looks very good!!

I left some comments for discussion.

config/script.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+// EvaluatedScriptJob represents an evaluated job block
+type EvaluatedScriptJob struct {
+	Command  []string
+	Commands [][]string
+}
+
+// EvaluatedScript represents an evaluated script block
+type EvaluatedScript struct {
```
snakster, line 29: I think the naming should align with similar types of this package, i.e.
```
// Stack represents an evaluated stack.
type Stack struct { ...

// Assert represents evaluated assert block configuration.
type Assert struct { ...
```

config/script.go
wmalik: Yes, I considered using it however I'd like the eval* functions to return specific errors (e.g. not a general error e.g.`ErrSchema`) that are relevant to the attributes e.g. description

config/script_test.go
wmalik: Good idea, that would save us some lines of code and also the diff output is much more easier to read.

config/script.go
i4ki: Very good point. I missed that.

config/script.go
i4ki: you can wrap the errors with `errs.AppendWrap()`, or `err.Append(errors.E(OtherErr, err))`.

wmalik: I see, I'll use the `evalString()` function then. Would it be possible to print only the top level error when we later emit this to the user's stdout? 🤔

wmalik:
@snakster @i4ki thanks for the review. I have addressed your feedback in these commits:

* https://github.com/terramate-io/terramate/pull/1255/commits/19193918a24ed71a2cc01ec3932942b574fe4930
* https://github.com/terramate-io/terramate/pull/1255/commits/17a7bbba647426a9580a745c5cc31de2f38c1155

Let me know if you are fine with the changes, then I will squash+push and ask for an approval.

snakster:
Looks good 👍 

One request, could you add the `hcl.Range` to the `config.Script` type, similar to how it's done in `config.Assert`?

It would be convenient for printing out something like `location: /terramate.tm:200,20` when listing the scripts.

config/script.go
i4ki: I think the wrapping makes sense (as you did now).

```
terramate schema error: invalid type for script.description
```

Do you think it should be still improved?

wmalik: Ah sorry, I forgot to embed the underlying error, fixed now.
image

snakster (approved):

i4ki (approved):
LGTM

i4ki (approved):
LGTM 🎉

PR #1258, opened 2023-11-23
chore: improve the released docker image.

dejanu:
# Reason for This Change

Update Dockerfile according to best practices and reduce the size of the image 

## Description of Changes

- Cleaning up the apt cache and removing /var/lib/apt/lists
- Pin git version in apt install

PS: Update the base image to 23.04 might be beneficial also since  is smaller by 2.5 MB,  contains 17 fewer packages, introduces no new vulnerability but removes 10


i4ki:
Hi @dejanu 

Thanks for your contribution. The changes makes sense.
Unfortunately, we can't run CI on external contributors changes at the moment.
But I can cherry-pick your commit in a separate PR and then you get the credits.
We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), then if you can `git commit --amend` your commit and make it compliant, we can have it merged. Just make the title similar to the (updated) PR title here.

soerenmartius:
Hey, thanks again for your contribution. I cherry picked your commit in #1357 and updated the base image also.

PR #1261, opened 2023-11-24
chore(tests): fix flaky e2e test.

i4ki:
# Reason for This Change

The cloudstore.Data was reused between testcases and then it hit some cases not implemented in the testserver (handling drifts in stacks still pending).

## Description of Changes

The `tc.cloudData` was serialized and then loaded again in each subtest.

### DIFF
--- a/cloud/testserver/cloudstore/store.go
+++ b/cloud/testserver/cloudstore/store.go
@@ -37,7 +37,7 @@ type (
 		Members     []Member                   `json:"members"`
 		Stacks      []Stack                    `json:"stacks"`
 		Deployments map[cloud.UUID]*Deployment `json:"deployments"`
-		Drifts      []*Drift                   `json:"drifts"`
+		Drifts      []Drift                    `json:"drifts"`
 	}
 	// Member represents the organization member.
 	Member struct {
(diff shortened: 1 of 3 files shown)


wmalik (approved):

cloud/testserver/cloudstore/store.go
```diff
@@ -208,15 +208,15 @@ outer:
...
 
 // GetStackByMetaID returns the given stack.
-func (d *Data) GetStackByMetaID(org Org, id string) (*Stack, int, bool) {
+func (d *Data) GetStackByMetaID(org Org, id string) (Stack, int, bool) {
 	d.mu.RLock()
 	defer d.mu.RUnlock()
 	for i, st := range org.Stacks {
 		if st.Stack.MetaID == id {
-			return &st, i, true
+			return st, i, true
```
i4ki, line 216: as we are returning here, it wasn't actually buggy because the `st` scoped variable was escaping with the last value (the right one). But it's indeed a bad practice getting the pointer of a range variable, then it was best to just use a value instead of a pointer.
Thanks @wmalik for pointing this out.

PR #1262, opened 2023-11-26
chore(perf): improve generate performance.

i4ki:
# Reason for This Change

change in the future).

## Description of Changes

Overview of what was changed.


i4ki:
closing this for now because the Evaluator v2 ensures the evalctx is always reused.

PR #1275, opened 2023-12-05
chore(scripts): add debug commands for scripts

snakster:

Three new experimental commands are added, related to the new scripts feature. For now, they are intended for debugging as they allow inspecting the script hierarchy.

#### script-list

Shows a list of all available scripts in the current directory. If there are multiple definitions, it uses the most parent definition, or the first sibling in case all definitions are on the same level.

Usage:
`terramate experimental script list`

Output:
```
deploy
  some description
  Defined at /
    (+1 more)

stagify
  do something
  Defined at /stacks/stage

validate
  some description
  Defined at /
    (+1 more)
```

#### script info

Shows information about _a specific script_. For each definition, it lists the jobs and affected stacks.

Usage: 
`terramate experimental script info -- validate`

Output:
```
Definition: /scripts.tm.hcl:11,1-24,2
Description: some description
Stacks:
  /stacks/prod
  /stacks/prod/backend
  /stacks/prod/frontend
  /stacks/stage
  /stacks/stage/backend
  /stacks/stage/frontend
Jobs:
  * ["echo","hello"]
    ["echo","bye"]
  * ["nono","nono"]

Definition: /stacks/dev/scripts.tm.hcl:1,1-9,2
Description: overriding validate
Stacks:
  /stacks/dev
Jobs:
  * ["echo","hello"]
    ["echo","bye"]
```

#### script tree

Shows a tree of all available scripts in the current directory and also which stacks are in the current selection scope.

Usage:
`terramate experimental script tree`

Output:
```
/
│ * deploy: some description
│ * validate: some description
└── stacks
    └── stage
        │ # stage
        │ * deploy: overriding deploy
        │ * stagify: do something
        │ ~ validate
        ├── frontend
        │     # frontend
        │     ~ deploy
        │     ~ stagify
        │     ~ validate
        └── backend
              # backend
              ~ deploy
              ~ stagify
              ~ validate
```

* I did not add any tests for `script tree` since this is really just intended for debugging for now.

No. It adds three new commands, but they are all categorized as experimental.

### DIFF
--- a/cmd/terramate/cli/script_info.go
+++ b/cmd/terramate/cli/script_info.go
@@ -1,0 +1,160 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package cli
+
+import (
+	"sort"
+
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/terramate-io/terramate/config"
+	"github.com/terramate-io/terramate/hcl"
+	"github.com/terramate-io/terramate/hcl/ast"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+func (c *cli) printScriptInfo() {
+	labels := c.parsedArgs.Experimental.Script.Info.Labels
+
+	stacks, err := c.computeSelectedStacks(false)
+	if err != nil {
+		fatal(err, "computing selected stacks")
+	}
+
+	m := newScriptsMatcher(labels)
+	m.Search(c.cfg(), stacks)
+
+	for _, x := range m.Results {
+		c.output.MsgStdOut("Definition: %v", x.ScriptCfg.Range)
+		c.output.MsgStdOut("Description: %v", formatScriptDescription(x.ScriptCfg))
+
+		if len(x.Stacks) > 0 {
+			c.output.MsgStdOut("Stacks:")
+			for _, st := range x.Stacks {
+				c.output.MsgStdOut("  %v", st.Dir())
+			}
+		} else {
+			c.output.MsgStdOut("Stacks: (none)")
+		}
+
+		c.output.MsgStdOut("Jobs:")
+		for _, job := range x.ScriptCfg.Jobs {
+			for cmdIdx, cmd := range formatScriptJob(job) {
+				if cmdIdx == 0 {
+					c.output.MsgStdOut("  * %v", cmd)
+				} else {
+					c.output.MsgStdOut("    %v", cmd)
+				}
+			}
+		}
+
+		c.output.MsgStdOut("")
+	}
+}
+
+type scriptInfoEntry struct {
+	ScriptCfg *hcl.Script
+	Stacks    config.List[*config.SortableStack]
+}
+
+type scriptsMatcher struct {
+	labels []string
+
+	Results []*scriptInfoEntry
+}
+
+func newScriptsMatcher(labels []string) *scriptsMatcher {
+	return &scriptsMatcher{
+		labels:  labels,
+		Results: nil,
+	}
+}
+
+func (m *scriptsMatcher) Search(cfg *config.Root, stacks config.List[*config.SortableStack]) {
+	rootScope := make(config.List[*config.SortableStack], len(stacks))
+	copy(rootScope, stacks)
+
+	m.searchRecursive(cfg.Tree(), &rootScope)
+}
+
+func tryGetScriptByName(cfg *config.Tree, labels []string) *hcl.Script {
+	for _, s := range cfg.Node.Scripts {
+		if slices.Equal(s.Labels, labels) {
+			return s
+		}
+	}
+	return nil
+}
+
+func sortedKeys[V any](m map[string]V) []string {
+	ks := maps.Keys(m)
+	sort.Strings(ks)
+	return ks
+}
+
+func (m *scriptsMatcher) searchRecursive(cfg *config.Tree, scope *config.List[*config.SortableStack]) {
+	scriptCfg := tryGetScriptByName(cfg, m.labels)
+	if scriptCfg != nil {
+		m.addResult(cfg, scriptCfg, scope)
+	} else {
+		for _, k := range sortedKeys(cfg.Children) {
+			m.searchRecursive(cfg.Children[k], scope)
+		}
+	}
+}
+
+func (m *scriptsMatcher) addResult(cfg *config.Tree, scriptCfg *hcl.Script, scope *config.List[*config.SortableStack]) {
+	outerScope := config.List[*config.SortableStack]{}
+	innerScope := config.List[*config.SortableStack]{}
+
+	cfgdir := cfg.Dir()
+	for _, st := range *scope {
+		stackdir := st.Dir()
+
+		if stackdir.HasDirPrefix(cfgdir.String()) {
+			innerScope = append(innerScope, st)
+		} else {
+			outerScope = append(outerScope, st)
+		}
+	}
+
+	// Modify the current scope
+	*scope = outerScope
+
+	newEntry := &scriptInfoEntry{
+		ScriptCfg: scriptCfg,
+		// This may be updated at a deeper level as we move stacks to children
+		Stacks: innerScope,
+	}
+	m.Results = append(m.Results, newEntry)
+
+	for _, k := range sortedKeys(cfg.Children) {
+		m.searchRecursive(cfg.Children[k], &newEntry.Stacks)
+	}
+}
+
+func formatScriptDescription(cfg *hcl.Script) string {
+	return string(ast.TokensForExpression(cfg.Description.Expr).Bytes())
+}
+
+func formatScriptJob(job *hcl.ScriptJob) []string {
+	if job.Commands != nil {
+		switch e := job.Commands.Expr.(type) {
+		case *hclsyntax.TupleConsExpr:
+			commands := []string{}
+			for _, cmd := range e.Exprs {
+				commands = append(commands, string(ast.TokensForExpression(cmd).Bytes()))
+			}
+			return commands
+
+		default:
+			return []string{string(ast.TokensForExpression(e).Bytes())}
+		}
+
+	} else if job.Command != nil {
+		return []string{string(ast.TokensForExpression(job.Command.Expr).Bytes())}
+	}
+
+	return []string{}
+}
--- a/cmd/terramate/cli/script_list.go
+++ b/cmd/terramate/cli/script_list.go
@@ -1,0 +1,93 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package cli
+
+import (
+	"github.com/terramate-io/terramate/config"
+	"github.com/terramate-io/terramate/hcl"
+	prj "github.com/terramate-io/terramate/project"
+)
+
+type scriptListEntry struct {
+	ScriptCfg *hcl.Script
+	Dir       string
+	DefCount  int
+}
+
+type scriptListMap map[string]*scriptListEntry
+
+func (c *cli) printScriptList() {
+	srcpath := prj.PrjAbsPath(c.rootdir(), c.wd())
+
+	cfg, found := c.cfg().Lookup(srcpath)
+	if !found {
+		return
+	}
+
+	entries := scriptListMap{}
+
+	addParentScriptListEntries(cfg, entries)
+	addChildScriptListEntries(cfg, entries)
+
+	for _, name := range sortedKeys(entries) {
+		entry := entries[name]
+
+		c.output.MsgStdOut("%v", name)
+		c.output.MsgStdOut("  %v", formatScriptDescription(entry.ScriptCfg))
+
+		c.output.MsgStdOut("  Defined at %v", entry.Dir)
+
+		if entry.DefCount > 1 {
+			c.output.MsgStdOut("    (+%v more)", entry.DefCount-1)
+		}
+
+		c.output.MsgStdOut("")
+	}
+}
+
+func addParentScriptListEntries(cfg *config.Tree, entries scriptListMap) {
+	for _, sc := range cfg.Node.Scripts {
+		scriptname := sc.Name()
+		defcount := 1
+		olddef := entries[scriptname]
+		if olddef != nil {
+			defcount = olddef.DefCount + 1
+		}
+
+		// Towards parents we replace child definitions
+		entries[scriptname] = &scriptListEntry{
+			ScriptCfg: sc,
+			Dir:       cfg.Dir().String(),
+			DefCount:  defcount,
+		}
+	}
+
+	if cfg.Parent != nil {
+		addParentScriptListEntries(cfg.Parent, entries)
+	}
+}
+
+func addChildScriptListEntries(cfg *config.Tree, entries scriptListMap) {
+	for _, k := range sortedKeys(cfg.Children) {
+		childCfg := cfg.Children[k]
+		for _, sc := range childCfg.Node.Scripts {
+			scriptname := sc.Name()
+
+			olddef := entries[scriptname]
+			if olddef != nil {
+				// Towards children we keep parent definitions
+				olddef.DefCount++
+				continue
+			}
+
+			entries[scriptname] = &scriptListEntry{
+				ScriptCfg: sc,
+				Dir:       childCfg.Dir().String(),
+				DefCount:  1,
+			}
+		}
+
+		addChildScriptListEntries(childCfg, entries)
+	}
+}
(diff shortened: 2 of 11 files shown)


cmd/terramate/cli/script_tree.go
```diff
@@ -0,0 +1,164 @@
...
+
+	blockPrefix := prefix
+	if len(node.Children) > 0 {
+		blockPrefix += "│ "
+	} else {
+		blockPrefix += "  "
+	}
+
+	if node.StackName != "" {
+		writeln(blockPrefix + "# " + stackColor(node.StackName))
```
wmalik, line 71: we could use:
```
fmt.Fprintln(w, blockPrefix + "# " + stackColor(node.StackName))
```

and `fmt.Fprint(w, ...)` later in this file.

Then we could remove the `writeln` + `write` functions. What do you think?

wmalik (commented):
Looks great, just one minor comment.

cmd/terramate/cli/script_tree.go
snakster: Yes, that's nicer, thanks!

wmalik (approved):

wmalik (approved):

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,158 @@
...
+	myStacks := config.List[*config.SortableStack]{}
+
+	cfgdir := cfg.Dir()
+	for _, st := range *stacks {
+		stackdir := st.Dir()
+
+		if stackdir.HasPrefix(cfgdir.String()) {
+			myStacks = append(myStacks, st)
+		} else {
+			parentStacks = append(parentStacks, st)
```
i4ki, line 112: `parentStacks` is a misleading name as it will store all other stacks not children of this cfg tree.

```suggestion
			otherStacks = append(parentStacks, st)
```

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,158 @@
...
+		m.searchRecursive(cfg.Children[k], &newEntry.Stacks)
+	}
+}
+
+func formatScriptDescription(cfg *hcl.Script) string {
+	toks := ast.TokensForExpression(cfg.Description.Expr)
+	s := string(toks.Bytes())
+	s = strings.Trim(s, "\"")
+	return s
+}
```
i4ki, line 136:
```suggestion
func formatScriptDescription(cfg *hcl.Script) string {
	return string(ast.TokensForExpression(cfg.Description.Expr).Bytes())
}
```

This special treatment for string could have strange consequences depending on the variable interpolations in place.

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	} else if job.Command != nil {
+		toks := ast.TokensForExpression(job.Command.Expr)
+		commands = append(commands, string(toks.Bytes()))
+	} else {
+		return commands
+	}
+
+	return commands
+}
```
i4ki, line 158: Do we need to loop over the expression items and call `ast.TokensForExpression()` for each one of them?
What about this?
```
func formatScriptJob(job *hcl.ScriptJob) string {
	if job.Commands != nil {
		return string(ast.TokensForExpression(job.Commands))
	} else {
		return string(ast.TokensForExpression(job.Command))
	}
}
```

The code was just handling tuples, but the command can be the result of of a funcall, or for loop, etc.

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,158 @@
...
+func (c *cli) printScriptInfo() {
+	labels := c.parsedArgs.Experimental.Script.Info.Labels
+
+	stacks, err := c.computeSelectedStacks(false)
+	if err != nil {
+		fatal(err, "computing selected stacks")
+	}
+
+	m := newScriptsMatcher(labels)
+	m.Search(c.cfg().Tree(), &stacks)
```
i4ki, line 26: I wonder if this pointer usage needs to be leaked to the caller.
Because the scoped stacks computation is complex and depends on flags, etc, so ideally we must do it once per Terramate invocation, but this method is destructive and `stacks` cannot be used afterward.

Maybe you can make `Search()` receive the list (not a pointer) and then inside it yo add:
```
copied := make(..., len(stacks))
copy(copied, stacks)
m.searchRecursive(cfg, &copied)
```
then just the `copied` is mutated.

Additionally, it looks like only the stack paths are used, then maybe just passing `stacks.Paths()` would suffice?

i4ki (changes requested):
The implementation looks great and works in most scenarios but I found the bug below:

```
✗ cat cmd/terramate/script.tm
script "a" "b" {
    description = "test"
    job {
        command = ["echo", "terramate"]
    }
}

stack {}

✗ cat cmd/terramate-ls/script.tm
script "a" "b" {
    description = "!"
    job {
        command = ["echo", "terramate-ls"]
    }
}

stack {}
```

and then:

```
./bin/terramate experimental script run a b
(skipping) provided criteria does not allow script execution
Found a b defined at /cmd/terramate/script.tm:1,1-6,2 having 1 job(s)

/cmd/terramate (job:0.0)> echo terramate
terramate

/cmd/terramate-ls (job:0.0)> echo terramate
terramate

/cmd/terramate/e2etests/cloud/interop/testdata/interop-stacks/basic-drift (job:0.0)> echo terramate
terramate

/cmd/terramate/e2etests/cloud/interop/testdata/interop-stacks/empty (job:0.0)> echo terramate
terramate
Found a b defined at /cmd/terramate-ls/script.tm:1,1-6,2 having 1 job(s)
(skipping) provided criteria does not allow script execution
```

Notice that `cmd/terramate-ls` script definition was not used when invoking its job. It printed "terramate" for the `cmd/terramate-ls" script.

Additionally, it prints that selected criteria does not select stacks twice, but why?

By the code, it looks like for each configuration tree node, it's computing the visible configs from unrelated dirs (not from the parent branch).

@wmalik have a look.

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,158 @@
...
+		c.output.MsgStdOut("Description: %v", formatScriptDescription(x.ScriptCfg))
+		c.output.MsgStdOut("Stacks:")
+		for _, st := range x.Stacks {
+			c.output.MsgStdOut("  %v", st.Dir())
+		}
+		c.output.MsgStdOut("Jobs:")
+		for _, job := range x.ScriptCfg.Jobs {
+			for cmdIdx, cmd := range formatScriptJob(job) {
+				if cmdIdx == 0 {
+					c.output.MsgStdOut("  * %v", cmd)
```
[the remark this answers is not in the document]
snakster: It prints the leading `*` at the first command of each job:
```
* job0:cmd0
  job0:cmd1
* job1:cmd0
  job1:cmd1
  ...
```
```
* job0:cmd0
* job1:cmd0
  job1:cmd1
* job2:cmd0
  ...
```
I used this as a minimalistic way format lists of lists.

cmd/terramate/cli/script_info.go
snakster: Marking as resolved after our call. Here's a small visual example of how the algorithm works:
https://docs.google.com/drawings/d/1xTFA5FXE9YKa_eb2zyHJIq-sPW1dmwiZm2Y5Xozoe8Y/edit?usp=sharing

cmd/terramate/cli/script_info.go
snakster: Good point, I changed it as you suggested. I couldnt just use the paths because it looks to me like script run needs  them.

snakster:
> The implementation looks great and works in most scenarios but I found the bug below:
>  ...

Good catch! 👍 
There was a bug in how sub-directories are tested for, so `/cmd/terramate-ls` would mistakenly be detected as a sub-directory of `/cmd/terramate`. It has been fixed and a test was added to the script info tests.

cmd/terramate/cli/script_info.go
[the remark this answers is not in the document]
i4ki: Gotcha.

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,161 @@
...
+	cfgdir := cfg.Dir()
+	for _, st := range *parentStacks {
+		stackdir := st.Dir()
+
+		if stackdir.HasDirPrefix(cfgdir.String()) {
+			myStacks = append(myStacks, st)
+		} else {
+			newParentStacks = append(newParentStacks, st)
+		}
+	}
```
i4ki, line 117: I think the names `parentStacks`, `newParentStacks` and `myStacks` are misleading here.
We start with "all stacks" (in the first call, but later it's the "pending stacks", and for the given `cfg` node, we split it into "child stacks" (or stacks in current script scope) and "unrelated stacks".

The "unrelated stacks" are the stacks that the current script (defined at `cfg`) does not apply, then you update the "pending stacks" pointer to become the rest.

Then down there you call `m.searchRecursive()` again for each cfg children node, providing a pointer to the "child stacks" of this node.

Then, IMO, the names should be:

```
parentStacks    = pendingStacks
newParentStacks = unrelatedStacks
myStacks        = scopedStacks
```

wdyt?

cmd/terramate/cli/script_list.go
```diff
@@ -0,0 +1,102 @@
...
+type scriptListEntry struct {
+	ScriptCfg *hcl.Script
+	Dir       string
+	DefCount  int
+}
+
+type scriptListMap map[string]*scriptListEntry
+
+func (c *cli) printScriptList() {
+	srcpath := project2.PrjAbsPath(c.rootdir(), c.wd())
```
i4ki, line 23: this same directory imports "project" as "prj". Can you change here to use the same for now?

We can fix this in a separate PR renaming the "project" type so this conflict does not happen again.

cmd/terramate/cli/script_tree.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	for _, sc := range node.Scripts {
+		desc := formatScriptDescription(sc)
+		fmt.Fprintln(w, blockPrefix+"* "+scriptColor(sc.Name()+": "+desc))
+	}
+
+	if node.IsStack {
+		for _, p := range parentScripts {
+			_, found := slices.BinarySearchFunc(node.Scripts, p,
```
i4ki, line 84: I don't expect more than 10 scripts per directory, so I think `slices.ContainsFunc()` could even perform better. I like to follow Rob Pike style rules for complexity defined [here](https://www.lysator.liu.se/c/pikestyle.html), check `rule 3`.

cmd/terramate/cli/script_tree.go
```diff
@@ -0,0 +1,172 @@
...
+					return strings.Compare(a.Name(), b)
+				})
+			if !found {
+				fmt.Fprintln(w, blockPrefix+parentScriptColor("~ "+p))
+			}
+		}
+	}
+
+	for _, e := range node.Scripts {
+		_, found := slices.BinarySearch(parentScripts, e.Name())
```
i4ki, line 95: the serial search usually performs better when `n` is small due to better use of CPU cache.

cmd/terramate/cli/script_tree.go
```diff
@@ -0,0 +1,172 @@
...
+
+		childNode := &scriptsTreeNode{
+			DirName: dirname,
+			IsStack: isStack,
+			Visible: isStack,
+			Parent:  cur,
+		}
+
+		childNode.Scripts = append(childNode.Scripts, childCfg.Node.Scripts...)
+
```
i4ki, line 167:
```suggestion
```

cmd/terramate/e2etests/core/script_info_test.go
```diff
@@ -0,0 +1,183 @@
...
+)
+
+func TestScriptInfo(t *testing.T) {
+	t.Parallel()
+
+	layout := []string{"d:stacks"}
+
+	addStackWithScripts := func(path string, scripts []string) {
+		layout = append(layout,
+			fmt.Sprintf("s:%s", path))
```
i4ki, line 24:
```suggestion
		layout = append(layout, fmt.Sprintf("s:%s", path))
```

cmd/terramate/e2etests/core/script_info_test.go
```diff
@@ -0,0 +1,183 @@
...
+				}
+				job {
+					command = ["echo", "2"]
+				}
+			}
+			`, name, name, path)
+		}
+
+		layout = append(layout,
+			fmt.Sprintf(`f:%s/script.tm:%s`, path, content))
```
i4ki, line 45:
```suggestion
		layout = append(layout, fmt.Sprintf(`f:%s/script.tm:%s`, path, content))
```

cmd/terramate/e2etests/core/script_info_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	addStackWithScripts("stacks/a", []string{"deploy", "other"})
+	addStackWithScripts("stacks/a/a1", []string{"other2"})
+	addStackWithScripts("stacks/a/a1/a2", []string{"deploy"})
+	addStackWithScripts("stacks/b", []string{"deploy", "other"})
+	addStackWithScripts("stacks/bb", []string{"other3"})
+
+	s := sandbox.New(t)
+	s.BuildTree(layout)
```
i4ki, line 72: it's minor but I feel like generating the buildTree instructions makes the test hard to read. I think it's a good abstraction for table test, where the programmer have a small, terse language for building dirs and files, etc.
But when you can do it programmatically (not a table test), then I feel like using:
```go
s := sandbox.New(t)
test.WriteFile(t, s.RootDir(), "stacks/a/script.tm", scriptAContent)
test.WriteFile(t, s.RootDir(), "stacks/b/script.tm", scriptBContent)
```
makes it easier to understand the test.

We can improve this later though. Wdyt?

cmd/terramate/e2etests/core/script_info_test.go
```diff
@@ -0,0 +1,183 @@
...
+			dir:    "",
+			want: RunExpected{
+				Stdout: "",
+			},
+		},
+		{
+			script: "deploy",
+			dir:    "",
+			want: RunExpected{
+				Stdout: mkExpected("deploy", "stacks/a", "1,1-12,5", []string{
```
i4ki, line 103: I think the location should be tested here, it makes it very cryptic and brittle. Maybe using `StdoutRegexes` would be better and then you just match what makes sense.

cmd/terramate/e2etests/core/script_info_test.go
```diff
@@ -0,0 +1,183 @@
...
+			script: "deploy",
+			dir:    "stacks/a/a1",
+			want: RunExpected{
+				Stdout: mkExpected("deploy", "stacks/a", "1,1-12,5", []string{
+					"/stacks/a/a1",
+				}) + mkExpected("deploy", "stacks/a/a1/a2", "1,1-12,5", []string{
+					"/stacks/a/a1/a2",
+				}) + mkExpected("deploy", "stacks/b", "1,1-12,5", []string{}),
+			},
+		},
```
i4ki, line 123: I really don't know what's being tested here 😂

i4ki (commented):
I added a bunch of new comments, but they are all minor or things to discuss. We can improve them later, if you agree with them, of course.

i4ki:
hey @snakster, please update the CHANGELOG.md

cmd/terramate/cli/script_info.go
```diff
@@ -0,0 +1,161 @@
...
+	m := newScriptsMatcher(labels)
+	m.Search(c.cfg().Tree(), stacks)
+
+	for _, x := range m.Results {
+		c.output.MsgStdOut("Definition: %v", x.ScriptCfg.Range)
+		c.output.MsgStdOut("Description: %v", formatScriptDescription(x.ScriptCfg))
+		c.output.MsgStdOut("Stacks:")
+		for _, st := range x.Stacks {
+			c.output.MsgStdOut("  %v", st.Dir())
+		}
```
i4ki, line 34: When no stacks are involved, it prints:
```
Definition: /cmd/cmd.tm:1,1-6,2
Description: slkljsk
Stacks:
Jobs:
  * ["echo","a"]
```

maybe we can make it:
```
Definition: /cmd/cmd.tm:1,1-6,2
Description: slkljsk
Stacks: (empty)
Jobs:
  * ["echo","a"]
```

wdyt?

i4ki: or `(none)`

cmd/terramate/cli/script_info.go
snakster: I changed it to use the term scope (rootScope, innerScope, outerScope), which should be a better description of the what it actually is.

cmd/terramate/cli/script_info.go
snakster: I'm still returning a list for better formatting, but as discussed I added a default case to properly handle anything that is not a list.

snakster:
> hey @snakster, please update the CHANGELOG.md

Done ✅ 

Regarding the tests, I agree that they should be improved, but we can do this in a later PR, since it will require some more work. I'm resolving the conversations for now.

i4ki (approved):
LGTM

Amazing :) Well done man!

PR #1276, opened 2023-12-05
fix(stdlib): tm_ternary() does not track expression scoped vars.

i4ki:

Some variables are scoped by their expression declaration, that's the case for `[for ...]`, `{for ...}` and `tm_dynamic`. In this case, the `tm_ternary()` implementation must use the provided closure evaluator variables instead of the outer Terramate evaluator.

This fixes an issue where `tm_ternary()` reported the error below:
```
eval expression: evaluating global["val"] from /stack scope: Call to function "tm_ternary" failed: /sandbox/stack/terramate.tm.hcl:5,14-15: partial evaluation failed: evaluating tm_ternary branch: eval expression: There is no variable named "a".
```

This is a fix for PR #909 

```
no
```


zied-elouaer (approved):

snakster (approved):
Noticed a typo in a comment, looks goods otherwise 👍

PR #1287, opened 2023-12-11
docs: remove non-existing functions

kkmlr:

`tm_strcontains` and `tm_timecmp` are documented but not implemented. This removes them from the docs. I've confirmed these are the only two functions documented but not implemented.

no


i4ki (approved):
LGTM

Thanks!

PR #1298, opened 2023-12-13
docs: document script commands

kkmlr:
Documents the upcoming `script` commands

```
no
```


docs/cli/cmdline/script/run.md
```diff
@@ -0,0 +1,51 @@
...
+- the script is available (scripts follow the same inheritance rules as globals)
+- any filters match. `script run` currently supports `--changed` and `--tags` filters.
+
+## Usage
+
+`terramate experimental script run [options] NAME`
+
+## Examples
+
+Run a script called "deploy" on all stacks where it is availablei:
```
wmalik, line 29:
```suggestion
Run a script called "deploy" on all stacks where it is available:
```

wmalik (commented):
Looks great, just a small typo.

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+# Script Info
+
+**Note:** This is an experimental command that is likely subject to change in the future.
+
+The `script info` command shows details of all scripts matching the command line argument. The details include:
+
+| Name        | Meaning                                |
+| ----------- | -------------------------------------- |
+| Definition  | Where the script is defined            |
+| Description | The description label in the script    |
```
i4ki, line 23:
```suggestion
| Description | The description attribute in the script    |
```

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+| Stacks      | The stacks where the script can be run |
+| Jobs        | The commands that comprise the script  |
+
+This information is always relative to the current directory (or the value of `-C`).
+
+**Note: ** Scripts with the same name that are overwritten in child directories are considered separate scripts.
```
i4ki, line 29:
```suggestion
**Note: ** Scripts with the same name that are overridden in child directories are considered separate scripts.
```
reasoning: overwritten usually means mutation but in this case, they are just overridden/shadowed.

docs/cli/cmdline/script/run.md
```diff
@@ -0,0 +1,51 @@
...
+next:
+  text: 'Script Tree'
+  link: '/cli/cmdline/script/tree'
+---
+
+# Script Run
+
+**Note:** This is an experimental command that is likely subject to change in the future.
+
+The `script run SCRIPT-NAME` command will run a Terramate script over a set of stacks. The script will run on all stacks under the current working directory where:
```
i4ki, line 18:
```suggestion
The `script run SCRIPT-NAME...` command will run a Terramate script over a set of stacks. The `SCRIPT-NAME` needs to match the labels defined in the `script` block.

The script will run on all stacks under the current working directory where:
```

docs/cli/cmdline/script/run.md
```diff
@@ -0,0 +1,51 @@
...
+
+- the script is available (scripts follow the same inheritance rules as globals)
+- any filters match. `script run` currently supports `--changed` and `--tags` filters.
+
+## Usage
+
+`terramate experimental script run [options] NAME`
```
i4ki, line 25:
```suggestion
`terramate experimental script run [options] NAME...`
```

docs/cli/cmdline/script/tree.md
```diff
@@ -0,0 +1,56 @@
...
+Shows a tree-view of all scripts relative to the current directory (or a specific directory with the -C flag). The tree expands all sub-directories, and the parent path back to the project root, showing script definitions per directory.
+
+## Usage
+
+`terramate experimental script tree`
+
+## Example
+
+```bash
+terramate experimental script tree
```
i4ki, line 25:
```suggestion
$ terramate experimental script tree
```

docs/cli/cmdline/script/tree.md
```diff
@@ -0,0 +1,56 @@
...
+└── stg
+    └── ec2
+          # ec2
+          ~ deploy
+```
+
+Same setup from a leaf directory
+
+```bash
+ terramate -C dev/ec2 experimental script tree
```
i4ki, line 47:
```suggestion
$ terramate -C dev/ec2 experimental script tree
```

i4ki (commented):
LGTM but added minor debatable improvement ideas.

docs/cli/cmdline/script/info.md
soerenmartius: in that case I would say `overridden (shadowed)` to be really explicit (e.g. I understand shadowed but never heart overridden)

mariux: shadowed seems wrong, i would go with @i4ki suggestion here.

@soerenmartius this is a common concept even in terraform https://developer.hashicorp.com/terraform/language/files/override

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+## Usage
+
+`terramate experimental script info NAME`
+
+## Examples
+
+Show information about a script called "deploy" defined at /scripts.tm.hcl
+
+```bash
+$ terramate-script experimental script-info deploy
```
mariux, line 40:
```suggestion
$ terramate-script experimental script info deploy
```

docs/cli/cmdline/script/run.md
```diff
@@ -0,0 +1,51 @@
...
+Do a dry run of running the deploy script:
+
+```bash
+terramate experimental script run --dry-run deploy
+```
+
+Run a script called "deploy" in a specific stack without recursing into subdirectories:
+
+```bash
+terramate -C path/to/stack experimental script --no-recursive run deploy
```
mariux, line 50:
```suggestion
terramate -C path/to/stack experimental script run --no-recursive deploy
```

no recursive should not be an option of script but of run. cc @i4ki

docs/cli/cmdline/script/tree.md
```diff
@@ -0,0 +1,56 @@
...
+/
+│ * deploy: deploy the infra
+└── dev
+    │ * destroy: destroy the infra
```
mariux, line 51: fmpov this is a bad example as we do not yet support reverse option or destruction triggers.

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+Definition: /scripts.tm.hcl:1,1-8,2
+Description: dummy deploy
+Stacks:
+  /dev/ec2
+  /prd/ec2
+  /stg/ec2
+Jobs:
+  * ["echo","deploying ${global.tags.env}"]
```
mariux, line 48:
```suggestion
  * ["echo","deploying ${global.env}"]
```

mariux: @i4ki do we already support interpolation here?

wmalik: @mariux yes we already support interpolation in commands.

docs/cli/cmdline/script/run.md
kkmlr: what's the meaning of the "..."? normally that indicates multiple words, but as far as I can see `script run` only supports one script. If the intention was that SCRIPT-NAME can have spaces, I think "..." will only be confusing.

If the intention is that we will support labels other than name, then perhaps it should be changed to `script run SCRIPT-LABEL...`

docs/cli/cmdline/script/run.md
kkmlr: same as above - is this NAME or LABEL? if NAME, we shouldn't use "..." just because we support multi word names

docs/cli/cmdline/script/tree.md
kkmlr: good point. I was intending to show a real world example of how context changes and "destroy" only allowed for dev made sense, but I've removed the destroy and the additional example for now

docs/cli/cmdline/script/run.md
i4ki: > If the intention is that we will support labels other than name, then perhaps it should be changed to script run SCRIPT-LABEL...

we do support labels and this is how it's supposed to be used.

The script definition below:
```
script "project" "deploy" {
...
}
```

can only be executed by:
```
$ terramate run script project deploy
```

The labels in the script definition declare a command, similar to `terramate experimental metadata` and `terramate experimental globals`.

So in my opinion, some hint must be put in the docs to make this clear.

i4ki: Another example:

```
script "multi word name" "another multi word name" {
...
}
```

this script can only be executed by:
```
$ terramate script run "multi word name" "another multi word name"
```
if you remove the quotes, it will NOT work... because we refer the `script` labels in the command-line to lookup the script.

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+## Usage
+
+`terramate experimental script info NAME`
+
+## Examples
+
+Show information about a script called "deploy" defined at /scripts.tm.hcl
+
+```bash
+$ terramate-script experimental script info deploy
```
i4ki, line 40: What's that `terramate-script`?

docs/cli/cmdline/script/run.md
i4ki: that's not the case, see my other comment.

docs/cli/cmdline/script/run.md
i4ki: yes, it's a `script run` option. It does not work if provided to the `script` part.

i4ki: https://github.com/terramate-io/terramate/blob/main/cmd/terramate/cli/cli.go#L233

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+next:
+  text: 'Script List'
+  link: '/cli/cmdline/script/list'
+---
+
+# Script Info
+
+**Note:** This is an experimental command that is likely subject to change in the future.
+
+The `script info` command shows details of all scripts matching the command line argument. The details include:
```
snakster, line 18: An alternative way to describe this would be to say that `script info` lists details about all script definitions with the given name.

The `Stacks` below I would describe as the stacks within the scope of the script, i.e. those stacks it can run on.

Even if you don't introduce the multiple definitions concept, I think it would be more precise to say "The stacks the script can run on"

docs/cli/cmdline/script/tree.md
```diff
@@ -0,0 +1,40 @@
...
+Shows a tree-view of all scripts relative to the current directory (or a specific directory with the -C flag). The tree expands all sub-directories, and the parent path back to the project root, showing script definitions per directory.
+
+## Usage
+
+`terramate experimental script tree`
+
+## Example
+
+```bash
+$ terramate experimental script tree
```
snakster, line 25: The output was changed a bit before the merge, now it's more compact:
```│ * deploy: "deploy the infra"
├── dev
│   └── #ec2
│         ~ deploy
├── prd
│   └── #ec2
│         ~ deploy
└── stg
    └── #ec2
          ~ deploy

**Review by snakster — COMMENTED** [review]
Looks good! 👍  I added a few small suggestions.

File: docs/cli/cmdline/script/tree.md (hunk 14 above)
Comment (kkmlr) [reply docs/cli/cmdline/script/tree.md:25]: resloved, thanks

File: docs/cli/cmdline/script/info.md (hunk 12 above)
Comment (kkmlr) [reply docs/cli/cmdline/script/info.md]: local alias. apologies, removed (and confirmed that's the only instance where it slipped through)

File: docs/cli/cmdline/script/info.md (hunk 13 above)
Comment (kkmlr) [reply docs/cli/cmdline/script/info.md]: Fixed. much better, thanks

File: docs/cli/cmdline/script/info.md [hunk 15]
```diff
@@ -0,0 +1,49 @@
...
+next:
+  text: 'Script List'
+  link: '/cli/cmdline/script/list'
+---
+
+# Script Info
+
+**Note:** This is an experimental command that is likely subject to change in the future.
+
+The `script info NAME` command lists details about all script definitions with the given NAME. The details include:
```
i4ki, line 18:
```suggestion
The `script info SCRIPT-NAME...` command lists details about all script definitions with the labels matching the given SCRIPT-NAME. The details include:
```

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+| Stacks      | The stacks within the scope of the script (i.e. those stacks it can run on) |
+| Jobs        | The commands that comprise the script                                       |
+
+This information is always relative to the current directory (or the value of `-C`).
+
+**Note: ** Scripts with the same name that are overridden in child directories are considered separate scripts.
+
+## Usage
+
+`terramate experimental script info NAME`
```
i4ki, line 33:
```suggestion
`terramate experimental script info SCRIPT-NAME...`
```

i4ki (changes requested):
I think the `script info` needs to be updated to match the `script run` docs regarding the name matching the script labels.

docs/cli/cmdline/script/run.md
kkmlr: I've changed it to what we agreed on in Slack, thanks

docs/cli/cmdline/script/info.md
kkmlr: I've updated this to be in line with `run`, but I have chosen to link to the run page rather than duplicate the same info, since in future we might have other ways of targetting labels (e.g. groupings) and IMO it should only be defined in one place

docs/cli/cmdline/script/info.md
kkmlr: resolved

docs/cli/cmdline/script/run.md
kkmlr: resolved

docs/cli/cmdline/script/info.md
```diff
@@ -0,0 +1,49 @@
...
+next:
+  text: 'Script List'
+  link: '/cli/cmdline/script/list'
+---
+
+# Script Info
+
+**Note:** This is an experimental command that is likely subject to change in the future.
+
+The `script info LABEL...` command lists details about all script definitions matching the given LABELs (see [./run](run) command for details about matching labels). The information provided by `script info` includes:
```
i4ki, line 18:
```suggestion
The `script info LABEL...` command lists details about all script definitions matching the given LABELs (see [run](./run) command for details about matching labels). The information provided by `script info` includes:
```
the link is malformed.

Additionally, I think the link text should be "script run" to not confuse with `terramate run`.

i4ki (changes requested):
small link issue

i4ki (approved):
LGTM

PR #1301, opened 2023-12-15
fix(scripts): dir prefix check works with root dir

snakster:

The `HasDirPrefix` function used for scripts matching doesn't handle checking for the `/` prefix properly. This PR fixes it and adds a test case for it.

```
no
```


wmalik (approved):

i4ki (approved):
lgtm

PR #1304, opened 2023-12-18
chore(tmc): add pagination support to /v1/stacks.

i4ki:

Uses pagination when requesting resources from `v1/stacks/...` endpoints.

```
no
```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -13,6 +13,7 @@ import (
 	"mime"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"os"
 	"path"
 
@@ -30,6 +31,8 @@ const Host = "api.terramate.io"
 
 // BaseURL is the default cloud.terramate.io base API URL.
 const BaseURL = "https://" + Host
+
+const defaultPageSize = 50
 
 const (
 	// WellKnownCLIPath is the well-known base path.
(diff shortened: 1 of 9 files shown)


cloud/cloud.go
```diff
@@ -80,12 +83,21 @@ type (
...
+	pageSize         int64 = defaultPageSize
+	debugAPIRequests bool
+)
 
 func init() {
 	if d := os.Getenv("TMC_API_DEBUG"); d == "1" || d == "true" {
 		debugAPIRequests = true
 	}
+	if sizeStr := os.Getenv("TMC_API_PAGESIZE"); sizeStr != "" {
+		size, _ := strconv.Atoi64(sizeStr)
```
wmalik, line 96: Should we handle the error here and print a warning?

strconv/format.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "strconv"
+
+// Itoa64 returns the string representation of the provided int64 using base 10.
+func Itoa64(i int64) string { return strconv.FormatInt(i, 10) }
+
+// Atoi64 returns the int64 represented in the given string.
+func Atoi64(a string) (int64, error) {
+	i, err := strconv.ParseInt(a, 10, 64)
+	return i, err
```
wmalik, line 14: nitpick: 
```suggestion
	return strconv.ParseInt(a, 10, 64)
```

wmalik (commented):
Added some minor inline comments.

Regarding the changes in this PR, I see three kinds of changes:

* stack.ID type change (int -> int64)
* new helper method `(c *Client) URL(...)`
* The pagination implementation

It's not clear to me if the first two changes are related to the pagination implementation, it would be nice to mention that in the commit message if that's the case. Otherwise we should split the commits if possible.

i4ki:
@wmalik I created a separated PR with the int64 changes: #1308 

About the `c.URL()` it was created as a URL builder supporting the query parameters for the pagination and other fields.

cloud/cloud.go
i4ki: This is just for debugging purposes.

wmalik (approved):

PR #1307, opened 2023-12-19
feat(generate): add stack_filter block

snakster:

A new `stack_filter` block type is supported in `generate_hcl` blocks. This allows for improved path-based conditional generation. It can significantly improve performance when using path-based rules with lots of stacks.

```
A new block type is added to the configuration schema.
```

### DIFF
--- a/generate/genhcl/genhcl.go
+++ b/generate/genhcl/genhcl.go
@@ -9,9 +9,11 @@ import (
 	"path"
 	"sort"
 
+	"github.com/gobwas/glob"
 	hhcl "github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/rs/zerolog/log"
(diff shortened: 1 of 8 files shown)


mariux:
tested locally, works like expected on my machine ;) this is great ;)

generate/genhcl/genhcl.go
```diff
@@ -147,6 +149,36 @@ func Load(
...
+
+		matchedAnyStackFilter := len(hclBlock.StackFilters) == 0
+		for _, cond := range hclBlock.StackFilters {
+			matched := true
+
+			for n, g := range map[string]glob.Glob{
+				"project path":    cond.ProjectPaths,
+				"repository path": cond.RepositoryPaths,
+			} {
+				if g != nil {
```
i4ki, line 161:
```suggestion
				if g != nil && !g.Match(st.Dir.String()) {
```
you can inline the ifs.

i4ki (dismissed):
LGTM but left some small suggestions.

i4ki (approved):
LGTM

PR #1308, opened 2023-12-19
chore: consistently uses int64 for TMC ids.

i4ki:

Some TMC endpoints have fields declared as `int64` integers (at least in theory because JSON can only store a maximum 2^53 integer value without loss) but CLI was unmarshaling into `int` (32 bits wide depending on the arch). The real motivation for this change is sticking to a size and using it for all ID-like fields. The new Pagination feature also has new fields defined as `int64`.
This is not a real problem at the moment but strictly following the OpenAPI spec is a good practice and some payloads are defined as `int64`. 
Additionally, by using int64 we are more future-proof in the case IDs are changed to not be serial anymore (unlikely, but who knows).

The #1304 had those changes because implementing pagination in the testserver led to the case of computing if a provided `stack_id` was inside a int64 range (because pagination.total and pagination.page were int64), not a big problem just doing a conversion but I felt like this should be better handled.
Additionally, the weird case of using `strconv.Atoi()` for some IDs (like `stack_id`) and then `Atoi64()` helper for the pagination `total` and other fields.

```
no
```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -15,13 +15,13 @@ import (
 	"net/http/httputil"
 	"os"
 	"path"
-	"strconv"
 
 	hversion "github.com/apparentlymart/go-versions/versions"
 	"github.com/rs/zerolog"
 	"github.com/terramate-io/terramate"
 	"github.com/terramate-io/terramate/cloud/stack"
 	"github.com/terramate-io/terramate/errors"
+	"github.com/terramate-io/terramate/strconv"
 	"github.com/terramate-io/terramate/versions"
 )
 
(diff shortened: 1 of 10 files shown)


cloud/cloud.go
```diff
@@ -144,15 +144,15 @@ func (c *Client) GetStack(ctx context.Context, orgUUID UUID, repo, metaID string
 }
 
 // StackDrifts returns the drifts of the given stack.
-func (c *Client) StackDrifts(ctx context.Context, orgUUID UUID, stackID int, page, perPage int) (DriftsStackPayloadResponse, error) {
-	path := path.Join(StacksPath, string(orgUUID), strconv.Itoa(stackID), "drifts")
+func (c *Client) StackDrifts(ctx context.Context, orgUUID UUID, stackID int64, page, perPage int) (DriftsStackPayloadResponse, error) {
```
snakster, line 147: > `page, perPage int`

Missed or intended? `page` and `perPage` are also int64 in the openapi spec.

snakster: Resolved per call: Intended, as this is changed in a later PR.

snakster (approved):

PR #1324, opened 2024-01-03
chore: auto-enable remote check only with remotes

snakster:

To enable using Terramate without a configured remote, the automatic default of `terramate.config.git.check_remote` needs to be adjusted. Before, it was always enabled by default. Now, it's only enabled by default if the user has any remotes configured in the Git repository, or if a default_remote is configured.

If the safeguard is enabled/disabled explicitly, this value still takes precedence as before.

* This PR also tweaks the error messages to get rid of the "this repo has no commits" message, because it is already handled by "this repo needs two commits".
* The `OptionalCheck` could be handled more generic in the future. I didn't use a `*Bool` because many assertions rely on == comparison between config blocks and I didn't want to change that.

* The behavior of the remote check safeguard is less strict, but this should only affect local, non-production cases (when no remotes are configured).


cmd/terramate/e2etests/core/general_test.go
```diff
@@ -492,6 +491,60 @@ func TestChangedWorksWithoutRemote(t *testing.T) {
...
+	}
+
+	path := test.WriteFile(t, git.BaseDir(), "file1", "# generated by terramate")
+	git.Add(path)
+	git.Commit("first commit")
+	path = test.WriteFile(t, git.BaseDir(), "file2", "# generated by terramate")
+	git.Add(path)
+	git.Commit("second commit")
+
+	assertSuccess(t)
```
wmalik, line 525: nitpick, we could inline the `assertSuccess` function body and remove that function?

```suggestion
	AssertRun(t, cli.ListChangedStacks())
	AssertRun(t, cli.Run("run", "--changed", HelperPath, "cat", "whatever"))
```

git/git.go
```diff
@@ -249,6 +249,16 @@ func (git *Git) RemoteAdd(name string, url string) error {
 	return err
 }
 
+// HasRemotes returns if a there are any remotes configured.
+func (git *Git) HasRemotes() (bool, error) {
+	res, err := git.exec("config", "--get-regexp", "remote\\.")
```
wmalik, line 254: Did you consider using `git remote` instead of `git config --get-regexp remote \\.`?

hcl/hcl.go
```diff
@@ -39,6 +39,34 @@ const (
...
+}
+
+// ToOptionalCheck creates an OptionalCheck value from a bool.
+func ToOptionalCheck(v bool) OptionalCheck {
+	if v {
+		return CheckIsTrue
+	}
+	return CheckIsFalse
+}
+
```
wmalik, line 69: That's a nice abstraction, I have the following questions:
* Would it make sense to rename the type to `OptionalBool`?
* Could we move it to a new package to make this reusable?
* Should we add unit tests for it?

wmalik (commented):
Looks good, I left some comments/questions.

git/git.go
snakster: Yes, I was googling a bit for a Git plumbing command that does this as I figured `git remote` is porcelain. I ended up using `git config`, not sure if this is any better - open to suggestions what the most stable way to detect this would be.

hcl/hcl.go
snakster: I also went through a couple of names, including `OptionalBool`. I settled for something less generic so I don't overpromise, because I didn't want to spend too much time on it for now.

Many other languages have a generic type for this, I considered using https://github.com/markphelps/optional, but this one also cannot be compared with `==`.

git/git.go
wmalik: Ok got it, I guess the git config command is less likely to change over time so we can keep it as it is.

i4ki (approved):
LGTM

PR #1328, opened 2024-01-04
chore: Create default printers

wmalik:

The default printers can be used by other packages without haaving to
manage printer instances e.g.:

```
printer.StdErr.Errorln("")
printer.StdErr.Successln("")
```

This approach is similar to the standard library `log` package where
a default logger instance is pre-initialized.

See https://github.com/golang/go/blob/master/src/log/log.go

no

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -6,6 +6,7 @@ package printer
 import (
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


i4ki (approved):
LGTM

PR #1329, opened 2024-01-04
fix: experimental eval output interpreted as fmt.

### LINKED ISSUE #1327: [BUG] `terramate eval` tries to decode url like encoded sequences
**Describe the bug**
`terramate eval` tries to substitute some url encoded like characters (eg: `%3D` ) in the output and replaces it with (MISSING) if not found

**To Reproduce**

1. Run command `terramate experimental eval --global 'x="E%2BkKhZA%3D"' 'tm_try(global.x, null)'`
2. See `E%!B(MISSING)kKhZA%!D(MISSING)` instead of `E%2BkKhZA%3D`

**Expected behavior**
I expect eval to return the output as is without any modification to the string.

**Log Output**
N/A

 - OS `linux`
 - OS Version `ubuntu-23.04`
 - Git Version `2.39.2`
 - Terramate Version `0.4.4-rc1`

i4ki:

The output of `eval`, `partial-eval` and `get-config-value` was being interpreted as Go's printf formatter.

Fixes #1327 

There are more places with incorrect usage of the function `MsgStdOut()` but they are harmless and will be fixed in a separate PR.

```
yes but in the output of the experimental commands (eval, partial-eval and get-config-value).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 ### Fixed
 
 - fix(generate): blocks with context=root were ignored if defined in stacks.
+- fix: experimental eval/partial-eval/get-config-value wrongly interprets the output as a formatter.
 
 ## 0.4.3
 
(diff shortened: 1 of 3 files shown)


wmalik (approved):
🚀

PR #1336, opened 2024-01-05
chore(scripts): add support for command options

snakster:
This PR adds support for an optional options object at the end of the command list. This is part of supporting cloud deployment sync for scripts.

* The options don't do anything yet, this will be done in a future PR. They are just parsed and validated for now.

```
no
```

### DIFF
--- a/config/script.go
+++ b/config/script.go
@@ -4,6 +4,9 @@
 package config
 
 import (
+	"fmt"
+	"strings"
+
 	hhcl "github.com/hashicorp/hcl/v2"
 	"github.com/terramate-io/terramate/errors"
 	"github.com/terramate-io/terramate/hcl"
@@ -19,12 +22,25 @@ const (
 	ErrScriptInvalidTypeCommand  errors.Kind = "invalid type for script.command"
 	ErrScriptInvalidTypeCommands errors.Kind = "invalid type for script.commands"
 	ErrScriptEmptyCmds           errors.Kind = "job command or commands evaluated to empty list"
+	ErrScriptInvalidCmdOptions   errors.Kind = "invalid options for script command"
 )
+
+// ScriptCmdOptions represents optional parameters for a script command
+type ScriptCmdOptions struct {
+	CloudSyncDeployment    bool
+	CloudSyncTerraformPlan string
+}
+
+// ScriptCmd represents an evaluated script command
+type ScriptCmd struct {
+	Args    []string
+	Options *ScriptCmdOptions
+}
 
 // ScriptJob represents an evaluated job block
 type ScriptJob struct {
-	Cmd  []string
-	Cmds [][]string
+	Cmd  *ScriptCmd
+	Cmds []*ScriptCmd
 }
 
 // Script represents an evaluated script block
@@ -39,9 +55,9 @@ type Script struct {
 // care about which command attr was set e.g. command or commands. This method
 // returns a list of commands irrespective of whether they were set through
 // job.command or job.commands
-func (es ScriptJob) Commands() [][]string {
+func (es ScriptJob) Commands() []*ScriptCmd {
 	if es.Cmd != nil {
-		return [][]string{es.Cmd}
+		return []*ScriptCmd{es.Cmd}
 	}
 	return es.Cmds
 }
@@ -52,6 +68,7 @@ func EvalScript(evalctx *eval.Context, script hcl.Script) (Script, error) {
 		Range:  script.Range,
 		Labels: script.Labels,
 	}
+
 	errs := errors.L()
 
 	desc, err := evalScriptDesc(evalctx, script.Description.Expr, "script.description")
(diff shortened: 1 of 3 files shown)


config/script.go
```diff
@@ -99,102 +145,140 @@ func evalScriptDesc(evalctx *eval.Context, expr hhcl.Expression, name string) (s
...
 		_, elem := it.Element()
-		if !elem.Type().IsTupleType() {
-			errs.Append(errors.E(ErrScriptInvalidTypeCommands, expr.Range(),
-				"field %s must be a list of list(string) but element %d has type %q",
-				name, index, elem.Type().FriendlyName()))
-			continue
+		if elem.Type() == cty.String {
+			r.Args = append(r.Args, elem.AsString())
+		} else if index == lastIndex {
+			if elem.Type().IsObjectType() {
```
wmalik, line 214: Considering a scenario where the options object is provided as a first element (or any position other than the last position), the resulting error will be the one on L224. 
We could catch that case with something like:

```suggestion
		} else if elem.Type().IsObjectType() {
		        if index != lastIndex {
		          errs.Append(errors.E(ErrScriptInvalidOptionsPosition, ..)
		          continue
		        }
```

wmalik (commented):
Looks great, added some minor suggestions.

config/script.go
snakster: I decided not to give any specific error for this, because I think it would be very unlikely to accidentally put the option anywhere but at the end, i.e. something like this would not happen:
```
[
  ["echo", "hello", {
      cloud_sync_deployment = "false"
    }, "world"],
]
```
What's more likely is that maybe a function is used to return one of the arguments, and this function returns the wrong type, in which case the "must be a list of strings" error is better than "options must be last element", which could be really confusing in particular if you didn't even use any options, which will probably be true for most of the script commands.

wmalik: Makes sense, thanks for the clarification.

config/script.go
```diff
@@ -99,102 +146,140 @@ func evalScriptDesc(evalctx *eval.Context, expr hhcl.Expression, name string) (s
...
+			} else {
+				errs.Append(errors.E(ErrScriptInvalidTypeCommand, expr.Range(),
+					"command options must be an object, but last element has type %s",
+					elem.Type().FriendlyName()))
+			}
+		} else {
+			errs.Append(errors.E(ErrScriptInvalidTypeCommand, expr.Range(),
+				"command must be a list(string), but element %d has type %s",
+				index, elem.Type().FriendlyName()))
 		}
```
wmalik, line 228: This if-else chain is a bit hard to read because of the nested clauses. We could use a switch statement here to get rid of the nested clauses? (and also make it easier to add more cases in future) e.g.:

```golang
	for it.Next() {
		_, elem := it.Element()
		switch {
		case elem.Type() == cty.String:
			r.Args = append(r.Args, elem.AsString())
		case (index == lastIndex) && elem.Type().IsObjectType():
			var err error
			r.Options, err = unmarshalScriptCommandOptions(elem, expr)
			errs.Append(err)
		case (index == lastIndex) && !elem.Type().IsObjectType():
			errs.Append(errors.E(ErrScriptInvalidTypeCommand, expr.Range(),
				"command options must be an object, but last element has type %s",
				elem.Type().FriendlyName()))
		default:
			errs.Append(errors.E(ErrScriptInvalidTypeCommand, expr.Range(),
				"command must be a list(string), but element %d has type %s",
				index, elem.Type().FriendlyName()))
		}
		index++
	}
```

What do you think?

config/script.go
```diff
@@ -63,7 +80,14 @@ func EvalScript(evalctx *eval.Context, script hcl.Script) (Script, error) {
...
 		if job.Command != nil {
-			command, err := evalScriptJobCommand(evalctx, job.Command.Expr, "command")
+			expr := job.Command.Expr
+			v, err := evalctx.Eval(expr)
+			if err != nil {
+				errs.Append(errors.E(ErrScriptSchema, expr.Range(), err, "evaluating command"))
+				continue
+			}
+
+			command, err := unmarshalScriptJobCommand(v, expr)
```
i4ki, line 90: nitpick: the "unmarshaling" name sounds a bit strange for me.

i4ki (approved):
LGTM

I agree with @wmalik that some places would benefit of a switch-case instead of if-elses but not a big problem IMO.

config/script.go
snakster: I also thought about this for a while, but I couldn't come up with anything better. Technically, I would say it's a type cast from cty.Value to a concrete type, but `castScriptJobCommand` also sounds strange. If you have any suggestions let me know.

config/script.go
snakster: I disagree on this one, I prefer the nested version because the decision tree is clearer like that.

For example, when I look at this, it's obvious in which branch I end up:
```
if a {
  if b {
  } else {
  }
} else {
}
```
As opposed to
```
if a && b {
} else if a && !b {
} else {
}
```
I look at this, I will wonder why the conditions are formulated like this and whether I'm missing any case.

config/script.go
snakster: Marking this as resolved for now, can refactor if we come up with a better name.

PR #1339, opened 2024-01-08
feat: promote experimental commands

wmalik:

The following commands are being promoted out of experimental:

* `experimental metadata` -> `debug show metadata`
* `experimental globals` -> `debug show globals`
* `experimental generate debug` -> `debug show generate-origins`
* `experimental run-env` -> `debug show runtime-env`
* `experimental run-graph` -> `debug render run-graph`

```
Yes, users must switch to the new command path for the above mentioned commands.
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -171,6 +171,22 @@ type cliSpec struct {
 
 	InstallCompletions kongplete.InstallCompletions `cmd:"" help:"Install shell completions"`
 
+	Debug struct {
+		Show struct {
+			Metadata        struct{} `cmd:"" help:"Shows metadata available on the project"`
+			Globals         struct{} `cmd:"" help:"List globals for all stacks"`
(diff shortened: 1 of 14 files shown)


soerenmartius:
@wmalik I just looked at the docs and we are referencing some of the comments in written text in the docs. We should make sure to update those also to not cause any outdated docs.

wmalik:
> @wmalik I just looked at the docs and we are referencing some of the comments in written text in the docs. We should make sure to update those also to not cause any outdated docs.

@soerenmartius oh good catch, will update that as part of this PR

soerenmartius:
> > @wmalik I just looked at the docs and we are referencing some of the comments in written text in the docs. We should make sure to update those also to not cause any outdated docs.
> 
> @soerenmartius oh good catch, will update that as part of this PR

awesome thank you!

i4ki (approved):
LGTM

PR #1345, opened 2024-01-09
feat: add terramate list --run-order

wmalik:

The `experimental run-order` command is being promoted to `list --run-order`.

* Add new flag --run-order that works similar to `experimental
  run-order`
* Mark `experimenal run-order` command as "hidden"
* Remove docs for `experimental run-order`

```
* `experimental run-order` will not be shown in `terramate experimental --help`
* A new flag `--run-order` command is now available in the `list` command

```


cmd/terramate/cli/cli.go
```diff
@@ -1331,6 +1332,11 @@ func (c *cli) printStacks() {
 
 	c.gitFileSafeguards(false)
 
+	if c.parsedArgs.List.RunOrder {
+		c.printRunOrder(true)
+		return
+	}
```
i4ki, line 1338: AFAIK the `list --run-order` must use the `list` output but sorted by the topological order. The existent run-order will be kept there as before but made hidden.

It just an additional flag to list, then you can even do `terramate list --changed --run-order --why`.

i4ki (changes requested):

cmd/terramate/cli/cli.go
wmalik: That makes sense, will fix.

cmd/terramate/cli/cli.go
```diff
@@ -1328,23 +1329,76 @@ func (c *cli) printStacks() {
...
 	}
-
 	c.gitFileSafeguards(false)
 
-	for _, entry := range c.filterStacks(report.Stacks) {
-		stack := entry.Stack
+	filteredStacks := c.filterStacks(report.Stacks)
+
+	if c.parsedArgs.List.RunOrder {
+		c.printListRunOrder(filteredStacks, c.parsedArgs.List.Why)
```
i4ki, line 1337: The list algorithm should be:

```
status := parseStatusFilter(c.parsedArgs.List.ExperimentalStatus)
report, _ := c.listStacks(mgr, c.parsedArgs.Changed, status)
filteredStacks := c.filterStacks(report.Stacks)

// convert into config.List
stacks := make(config.List[*config.SortableStack], len(entries))
for i, e := range entries {
    stacks[i] = e.Stack.Sortable()
}

// sort
orderedStacks, _, _ := run.Sort(c.cfg(), selectedStacks)

// print as before but ranging over config.List
```

The `list` should not call `c.computeSelectedStacks()` because this pulls more stacks if wants/wanted_by is used.

The `debug show order` command does not need to be changed.

i4ki (changes requested):

cmd/terramate/cli/cli.go
```diff
@@ -1328,23 +1329,55 @@ func (c *cli) printStacks() {
...
 
-	for _, entry := range c.filterStacks(report.Stacks) {
-		stack := entry.Stack
+	c.printStacksList(report.Stacks, c.parsedArgs.List.Why, c.parsedArgs.List.RunOrder)
+}
+
+func (c *cli) printStacksList(allStacks []stack.Entry, why bool, runOrder bool) {
+	filteredStacks := c.filterStacks(allStacks)
+
+	find := func(stackDir string) (stack.Entry, bool) {
```
i4ki, line 1340: I don't get why this `find` is needed 🤔

wmalik: the `--why` handling needs a `reason`. The `reason` is only present in the `stack.Entry` type, not in the  `*config.Stack` type.

i4ki: A `find` is strange because you have the list at hand already and the elements are the same, just a different data structure.
What do you think of making a map of `stack.Dir -> reason` when building the config.List?

```
reasons = map[project.Path]string{}
stacks := make(config.List[*config.SortableStack], len(filteredStacks))
for i, entry := range filteredStacks {
	stacks[i] = entry.Stack.Sortable()
	reasons[entry.Stack.Dir] = entry.Reason
}
```

then later you do:
```
if why {
    printer.Stdout.Println(stdfmt.Sprintf("%s - %s", friendlyDir, reasons[st.Dir])
} else {
    ...
}
```

i4ki (approved):
LGTM but I left an improvement idea, not sure if it's better.

soerenmartius:
@wmalik could you ping me when we get this merged? I will create a redirect from the old URL https://terramate.io/docs/cli/cmdline/run-order to https://terramate.io/docs/cli/cmdline/list so folks don't get confused whenever they are trying to access the old URL that might be referenced in older blog posts etc.

wmalik:
> @wmalik could you ping me when we get this merged? I will create a redirect from the old URL https://terramate.io/docs/cli/cmdline/run-order to https://terramate.io/docs/cli/cmdline/list so folks don't get confused whenever they are trying to access the old URL that might be referenced in older blog posts etc.

@soerenmartius sure

cmd/terramate/cli/cli.go
wmalik: That sounds great, just fixed it.

soerenmartius:
> > @wmalik could you ping me when we get this merged? I will create a redirect from the old URL https://terramate.io/docs/cli/cmdline/run-order to https://terramate.io/docs/cli/cmdline/list so folks don't get confused whenever they are trying to access the old URL that might be referenced in older blog posts etc.
> 
> @soerenmartius sure

awesome thank you

snakster (approved):

PR #1350, opened 2024-01-10
feat: add --detailed-exit-code to generate

wmalik:

The `terramate generate` command must return different exit codes depending on
whether it generated files or not. This is necessary for CI/CD pipelines to be
able to fail when files are not formatted.

```
The terramate generate command now supports a --detailed-exit-code flag.
```


i4ki (approved):
LGTM

PR #1351, opened 2024-01-10
feat: introduce --disable-safeguards=<options>

i4ki:

Introduces the flag `--disable-safeguards=<list of options>` and `terramate.config.disable_safeguards` configuration option. The options are:

| keyword            | meaning     |
| ------------------ | ----------- |
| git-untracked      | Disables the check for untracked files       |
| git-uncommitted    | Disables the check for uncommitted files     |
| git-out-of-sync    | Disables the check for git remote out of sync |
| outdated-code      | Disables the check for outdated code          |

Additionally, the following group keywords are supported:

| keyword            | meaning     |
| ------------------ | ----------- |
| git                | Disables all git related checks |
| all                | Disables all checks |
| none               | Re-enables all checks (if disabled by the config) |

The environment variable `TM_DISABLE_SAFEGUARDS` can also be used and it supports the same keyword list as the flag and config.
The list of keywords must be comma-separated list of strings.

Example:

```
terramate run --disable-safeguards=git-untracked,outdated-code -- <cmd>
```

The `-X` flag is a shorthand for `--disable-safeguards=all`.

The older flags `--disable-check-git-untracked`, `--disable-check-git-uncommitted` and `--disable-check-gen-code` are deprecated now.

```
yes, it adds new flags and environment variable support.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
   - `terramate list --experimental-status=` -> `terramate list --cloud-status=`
 - Add `list --run-order` flag to list stacks in the order they would be executed.
 - Add support for deployment syncing to script commands.
+- Add `disable_safeguards` configuration option and CLI flag.
 
 ## 0.4.4
 
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -21,6 +21,7 @@ import (
 	"github.com/terramate-io/terramate/cloud"
 	cloudstack "github.com/terramate-io/terramate/cloud/stack"
 	"github.com/terramate-io/terramate/cmd/terramate/cli/cliconfig"
+	"github.com/terramate-io/terramate/cmd/terramate/cli/clitest"
 	"github.com/terramate-io/terramate/cmd/terramate/cli/out"
 	"github.com/terramate-io/terramate/config/filter"
 	"github.com/terramate-io/terramate/config/tag"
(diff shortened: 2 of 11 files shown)


wmalik (approved):
Looks great!

soerenmartius:
Hope I am not saying something wrong here, but from what I understand we have the safeguards somewhat documented in https://terramate.io/docs/cli/configuration/#terramate-config-git-block-schema as well - is still this accurate or do we need to update this too?

wmalik (approved):

snakster (commented):
One thing I noticed is that the
> -X, --disable-safeguards-all

option is different from what was specified in the AC:

> new option -X -> alias for --disable-safeguards=all

Is because kong doesn't allow it any other way or was it overlooked?

wmalik:
> Hope I am not saying something wrong here, but from what I understand we have the safeguards somewhat documented in https://terramate.io/docs/cli/configuration/#terramate-config-git-block-schema as well - is still this accurate or do we need to update this too @i4ki ?

@soerenmartius good catch, I just added a deprecation label to those attributes in docs.

CHANGELOG.md
```diff
@@ -47,12 +48,6 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 - Add `terramate script info <scriptname>` to show details about a script.
 - Add `terramate script run <scriptname>` to run a script in all relevant stacks.
 - Add `stack_filter` block to `generate_hcl` for path-based conditional generation.
-- Promote experimental commands
```
snakster, line 50: This looks like an unintended change.

snakster (approved):
Looks good! 👍

PR #1354, opened 2024-01-11
fix: script commands abort if not enabled

i4ki:

If `scripts` is not enabled we must give a good error message explaining how to enable them.

```
no other than better error.
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -160,12 +160,12 @@ type cliSpec struct {
 		List struct{} `cmd:"" help:"Show a list of all scripts in the current directory"`
 		Tree struct{} `cmd:"" help:"Show a tree of all scripts in the current directory"`
 		Info struct {
-			Labels []string `arg:"" name:"labels" passthrough:"" help:"Name of the script"`
+			Cmds []string `arg:"" optional:"true" passthrough:"" help:"Script to show info"`
 		} `cmd:"" help:"Show detailed information about a script"`
 		Run struct {
 			NoRecursive bool     `default:"false" help:"Do not recurse into child stacks"`
 			DryRun      bool     `default:"false" help:"Plan the execution but do not execute it"`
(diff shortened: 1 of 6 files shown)


cmd/terramate/cli/script_info.go
```diff
@@ -15,7 +18,9 @@ import (
 )
 
 func (c *cli) printScriptInfo() {
-	labels := c.parsedArgs.Script.Info.Labels
+	c.checkScriptEnabled()
```
wmalik, line 21: nitpick:
What do you think about moving this tn the top level switch case (in cmd/terramate/cli/cli.go) before `c.printScriptInfo()` is invoked? (similar to how we use `c.setupGit()`)

wmalik (approved):
One minor nitpick, rest looks great.

cmd/terramate/cli/script_info.go
i4ki: It makes sense. I was on the fence if this or placing per command made sense. I updated.

cmd/terramate/cli/cli.go
```diff
@@ -606,14 +606,19 @@ func (c *cli) run() {
 	case "experimental cloud drift show":
 		c.cloudDriftShow()
 	case "script list":
+		c.checkScriptEnabled()
 		c.printScriptList()
 	case "script tree":
+		c.checkScriptEnabled()
 		c.printScriptTree()
-	case "script info <labels>":
+	case "script info <cmds>":
```
snakster, line 614: Since you are modifying this part and improving the errors, could add an additional `case "script info"` without args here, analogous this code below:
```
case "script run":
     c.checkScriptEnabled()
     log.Fatal().Msg("no script specified")
```

i4ki: sure

i4ki: fixed it, but I had to make the args `optional:"true"`. Just noticed that `terramate run` without args is not falling in the "run" case, for similar reason. Will be fixed separately.

snakster (approved):

PR #1355, opened 2024-01-11
feat(scripts): support for deployment cloud sync

snakster:
This PR implements cloud deployment syncing similar to `terramate run --cloud-sync-deployment`, but for individual script commands.

* The `ExecContext` was extended and has also been renamed in the process. It was used in many places and the variables had a different name each time, i.e. `run, runContext, stackContext, stacks, runStacks`. Refactored this so its always `run`/`runs`.
* The output of `script run` was changed so that code between `run` and `script run` could be re-used more easily.

```
- Support for optional script command options have been added, which allow syncing deployment info to Terramate Cloud.
- The output formatting of terramate script run has been changed slightly.
```

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -21,7 +21,7 @@ import (
 	"github.com/terramate-io/terramate/errors"
 	"github.com/terramate-io/terramate/printer"
 	prj "github.com/terramate-io/terramate/project"
-	"github.com/terramate-io/terramate/run"
+	runutil "github.com/terramate-io/terramate/run"
 	"github.com/terramate-io/terramate/run/dag"
 )
 
@@ -37,14 +37,22 @@ const (
 	ErrRunCommandNotFound errors.Kind = "command not found"
 )
 
-// ExecContext declares an stack execution context.
-type ExecContext struct {
+// runContext declares a stack run context.
+type runContext struct {
 	Stack *config.Stack
 	Cmd   []string
+
+	ScriptIdx    int
+	ScriptJobIdx int
+	ScriptCmdIdx int
+
+	CloudSyncDeployment        bool
+	CloudSyncDriftStatus       bool
+	CloudSyncTerraformPlanFile string
 }
 
-// RunResult contains exit code and duration of a completed run.
-type RunResult struct {
+// runResult contains exit code and duration of a completed run.
+type runResult struct {
 	ExitCode   int
 	StartedAt  *time.Time
 	FinishedAt *time.Time
@@ -86,7 +94,7 @@ func (c *cli) runOnStacks() {
 		}
 	}
 
-	orderedStacks, reason, err := run.Sort(c.cfg(), stacks)
+	orderedStacks, reason, err := runutil.Sort(c.cfg(), stacks)
 	if err != nil {
 		if errors.IsKind(err, dag.ErrCycleDetected) {
 			fatal(err, "cycle detected: %s", reason)
@@ -97,21 +105,6 @@ func (c *cli) runOnStacks() {
 
 	if c.parsedArgs.Run.Reverse {
 		config.ReverseStacks(orderedStacks)
-	}
-
-	var runStacks []ExecContext
-	for _, st := range orderedStacks {
-		run := ExecContext{
-			Stack: st.Stack,
-			Cmd:   c.parsedArgs.Run.Command,
-		}
-		if c.parsedArgs.Run.Eval {
-			run.Cmd, err = c.evalRunArgs(run.Stack, run.Cmd)
-			if err != nil {
-				c.fatal("unable to evaluate command", err)
-			}
-		}
-		runStacks = append(runStacks, run)
 	}
 
 	if c.parsedArgs.Run.CloudSyncDeployment && c.parsedArgs.Run.CloudSyncDriftStatus {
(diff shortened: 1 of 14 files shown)


soerenmartius:
Let's not forget to document this in https://terramate.io/docs/cli/cmdline/script/run also :)

sixstone-qq (approved):
Really awesome job in cloud client side :clap:

cmd/terramate/cli/run.go
```diff
@@ -148,12 +159,25 @@ func (c *cli) runOnStacks() {
...
+		ScriptRun:       false,
+		ContinueOnError: c.parsedArgs.Run.ContinueOnError,
+	})
 	if err != nil {
 		c.fatal("one or more commands failed", err)
 	}
 }
 
+// RunAllOptions define named flags for RunAll
+type RunAllOptions struct {
```
wmalik, line 174: We could make it package private since its only used internally?

```suggestion
// runAllOptions define named flags for RunAll
type runAllOptions struct {
```

cmd/terramate/cli/run.go
```diff
@@ -165,12 +189,16 @@ func (c *cli) runOnStacks() {
 // stacks.
 // If SIGINT is sent 3x then Terramate will send a SIGKILL to the currently
 // running process and abort the execution of all subsequent stacks.
-func (c *cli) RunAll(runStacks []ExecContext, isSuccessCode func(exitCode int) bool) error {
+func (c *cli) RunAll(
+	runs []RunContext,
+	isSuccessCode func(exitCode int) bool,
```
wmalik, line 194: the isSuccessCode seems to be the same for all callers, we could remove it as an argument and do it inline? 🤔

cmd/terramate/cli/script_run.go
wmalik, line 1: 👍🏾

cmd/terramate/cli/script_run.go
```diff
@@ -88,63 +88,95 @@ func (c *cli) runScript() {
...
-	if err := newCmd.Run(); err != nil {
-		return errors.E(err, "failed to execute command")
+	if len(deployRuns) == 0 {
+		return
 	}
 
-	return nil
-}
+	if !c.prj.isRepo {
+		fatal(errors.E("cloud features require a git repository"))
```
wmalik, line 146: Would it make sense to print a warning here instead? e.g.:

```golang
printer.Stderr.Warn("Cloud features require a git repository, will not sync deployment")
```

cmd/terramate/cli/script_run.go
```diff
@@ -88,63 +88,95 @@ func (c *cli) runScript() {
...
 
-func makeCommand(command []string, dir string, env []string, stdout, stderr io.Writer) (*exec.Cmd, error) {
-	cmdPath, err := run.LookPath(command[0], env)
-	if err != nil {
-		return nil, errors.E(err, "%s: command not found", command[0])
+	err := c.setupCloudConfig()
+	c.handleCriticalError(err)
+
+	if c.cloud.disabled {
+		return
```
wmalik, line 153: I think it would be more user friendly to print a warning here, and continue executing the command. As a user, I would prefer if my scripts continue to run irrespective of whether cloud is enabled or not.
What do you think?

wmalik: also, naked returns are a bit hard to read, e.g. the value of `err` returned here is not clear to me.

wmalik (approved):
Nice work, I left some comments.
I have not reviewd the tests yet but wanted to submit the review already.

wmalik (changes requested):
(approved by mistake earlier)

cmd/terramate/cli/script_run.go
snakster: I copied the `run` behaviour here (see `runOnStacks()`), which also fails in this case.

cmd/terramate/cli/script_run.go
snakster: Also more or less copied from how it's handled for `run`, in this case
`checkCloudSync()`.

I agree that it could be nicer though, but I'd say for the scope of this PR it's better to just stick to the current behavior of run, because it's already changing quite a few things.

cmd/terramate/cli/run.go
snakster: For `run script` its always the same, but for `run` it can have different values, see `runOnStacks()`:
```
if c.parsedArgs.Run.CloudSyncDriftStatus {
	isSuccessExit = func(exitCode int) bool {
		return exitCode == 0 || exitCode == 2
	}
}
```

cmd/terramate/cli/run.go
snakster: It's tied to `RunAll`, so we would have to make that private, too. You are not wrong, but I think most things in the cli package could be private. The function is now shared by `script run` and `run`, so you could say it's internal public as opposed to internal internal :)

IMO we should refactor this eventually, but for now I'd prefer to keep it as it is.

cmd/terramate/cli/script_run.go
wmalik: Makses sense to keep the behaviour similar to `run`, however I'd suggest we should double check if that is intentional behaviour.

cmd/terramate/cli/script_run.go
wmalik: same comment as https://github.com/terramate-io/terramate/pull/1355#discussion_r1462990310

cmd/terramate/cli/run.go
wmalik: We could make `RunAll` private too? I agree that there are other occurrences of functions that should be provided but since you are touching RunAll and adding the options it seems worth it to fix some low hanging fruit :)

cmd/terramate/cli/script_run.go
snakster: We agreed to keep the behaviour like this for now and potentially change it in a later PR.

wmalik (approved):

PR #1359, opened 2024-01-11
fix: change detector cannot read user's git config.

i4ki:

The Terramate change detector does not load user's global config and then is unable to ignore files configured in the `core.excludeFiles` option.

This fixes a reported bug where the user had a global `.gitignore` configuration but Terramate kept complaining of untracked files in the repository.

Fixes an issue reported by a customer but not tracked in GitHub.

- Depends on #1358
- I removed the @snakster implementation for retrieving global config (the inner vs outer wrappers) because they are not needed anymore.
- Now Terramate uses a single git wrapper with the exception that module change detector needs a different Workdir but that's the only different option.

```
no
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -770,9 +770,8 @@ func (c *cli) triggerStackByFilter() {
 		fatal(errors.E("trigger command expects either a stack path or the --experimental-status flag"))
 	}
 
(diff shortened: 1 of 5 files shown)


wmalik (dismissed):
Looks good

snakster (approved):
👍

snakster (approved):

PR #1375, opened 2024-01-17
fix: multiple stack_filter patterns did not match in some cases

snakster:

After some investigation it seems there is a bug in the glob library we use with certain pattern combinations:
```
g := glob.MustCompile("{**/project-factory/*/custom-roles/*,**/projects/*/landing-zone/custom-roles/*", '/')
const p = "stacks/project-factory/factory-1/custom-roles/role-1"
g.Match(p) // false, should be true
```
This PR adds a workaround by matching each pattern individually. It also escapes reserved special characters in patterns.

```
no
```

### DIFF
--- a/generate/generate_hcl_test.go
+++ b/generate/generate_hcl_test.go
@@ -1478,6 +1478,46 @@ func TestGenerateHCLStackFilters(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


wmalik (approved):
Looks good

PR #1376, opened 2024-01-18
fix: redact authorization header in API debug output

wmalik:

This PR redacts the Authorization header from API debugging output (when `TMC_API_DEBUG=1` is set) to prevent leaking credentials.

```
No. This is a change to the API debugging output only.

```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -329,7 +329,7 @@ func Request[T Resource](ctx context.Context, c *Client, method string, url url.
 	}
 
 	if debugAPIRequests {
-		data, _ := httputil.DumpRequestOut(req, true)
+		data, _ := dumpRequest(req)
 		fmt.Printf(">>> %s\n\n", data)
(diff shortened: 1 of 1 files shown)


cloud/dump_http.go
```diff
@@ -0,0 +1,14 @@
+package cloud
+
+import (
+	"net/http"
+	"net/http/httputil"
+)
+
+func dumpRequest(req *http.Request) ([]byte, error) {
```
sixstone-qq, line 8: *nitpick* I'd add some description because it's not clear what it does with regards the func name but otherwise it's good

wmalik: @sixstone-qq done.

I also removed the `dump_http.go` file and moved th function to `cloud/cloud.go`)

snakster (approved):

sixstone-qq (approved):
Nice :heavy_check_mark:

PR #1389, opened 2024-01-24
chore(cloud): add support attaching planfiles to synced deployments

snakster:
This PR adds support for using the `terramate run` options  `--cloud-sync-deployment` and `--cloud-sync-terraform-plan-file` in combination. It will attach information from a given Terraform planfile to the cloud deployment status.

Note that the plan must exist before the run is executed.

* The core of this functionality had already been implemented in #1355, because scripts is using the same code path.
* This is not tested via test server yet, because it wasn't clear to me which endpoint to use to query the result. This will be done in a later PR, together with adding tests for the scripts' variant of this feature.

```
It allows a new combination of flags that was previously not supported.
```

### DIFF
--- a/cmd/terramate/e2etests/cloud/run_cloud_deployment_test.go
+++ b/cmd/terramate/e2etests/cloud/run_cloud_deployment_test.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/terramate/e2etests/cloud/run_cloud_deployment_test.go
```diff
@@ -345,13 +373,17 @@ func TestCLIRunWithCloudSyncDeployment(t *testing.T) {
 				genIdsLayout = tc.layout
 			}
```
i4ki, line 375:
```suggestion
			// needed for invoking `terraform ...` commands in the sandbox tree (see tc.layout)
```

It wasn't obvious to me at first, maybe makes sense to add a note why this is needed.
But maybe it's just me, feel free to ignore.

i4ki (approved):
LGTM

wmalik (approved):

PR #1391, opened 2024-01-26
chore(cloud): add more metadata for GH PRs

snakster:

This PR adds additional deployment metadata for Github Pull Requests.

We want to move away from putting review-specific information in the metadata field and use the new review request model instead. That's why all new fields have been added there.

```
no
```


cloud/types.go
```diff
@@ -220,15 +223,20 @@ type (
...
+		CommitSHA             string    `json:"commit_sha"`
+		Number                int       `json:"number"`
+		Title                 string    `json:"title"`
+		Description           string    `json:"description"`
+		URL                   string    `json:"url"`
+		Labels                []Label   `json:"labels,omitempty"`
+		Reviewers             Reviewers `json:"reviewers,omitempty"`
+		ChangesRequestedCount int       `json:"changes_requested_count"`
+		ApprovedCount         int       `json:"approved_count"`
+		ChecksTotalCount      int       `json:"checks_total_count"`
```
sixstone-qq, line 237: Is it the sum of `checks_failure_count` + `checks_success_count`?

cloud/types.go
```diff
@@ -170,6 +170,9 @@ type (
 		GithubPullRequestTitle          string `json:"github_pull_request_title,omitempty"`
 		GithubPullRequestDescription    string `json:"github_pull_request_description,omitempty"`
 		GithubPullRequestState          string `json:"github_pull_request_state,omitempty"`
+		GithubPullRequestDraft          bool   `json:"github_pull_request_draft,omitempty"`
+		GithubPullRequestReviewRequired bool   `json:"github_pull_request_review_required,omitempty"`
+		GithubPullRequestMerged         bool   `json:"github_pull_request_merged,omitempty"`
```
sixstone-qq, line 175: Could we send this information inside the `DeploymentReviewRequest` instead?

sixstone-qq (commented):
Questions around API if we can change a few things

cloud/types.go
snakster: No, it's not the same. There's at least still "skipped" checks, and maybe others.

cmd/terramate/cli/cloud.go
```diff
@@ -477,30 +461,91 @@ func (c *cli) detectCloudMetadata() {
...
+	}
+
+	return allPulls, nil
+}
+
+func isGithubPRMerged(ghClient *github.Client, owner, repo string, pullNumber int) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultGithubTimeout)
+	defer cancel()
+
+	isMerged, _, _ := ghClient.PullRequests.IsMerged(ctx, owner, repo, pullNumber)
```
i4ki, line 503: when looking at the `IsMerged()` implementation, we can see that it can fail for several reasons and then we will set the state PR state to `merged` which is misleading/wrong.

I think that if the `pull.GetState()` gives the status `closed` and this request fails then in this case we must carry this information (*unable to check merged state*) and set the PR status to `closed (failed to detect merged state)` or something like that.
@sixstone-qq what do you think?

cmd/terramate/cli/cloud.go
```diff
@@ -477,30 +461,91 @@ func (c *cli) detectCloudMetadata() {
...
+		if err != nil {
+			return nil, err
+		}
+
+		allPulls = append(allPulls, pulls...)
+		if resp.NextPage == 0 {
+			break
+		}
+		opt.Page = resp.NextPage
+	}
```
i4ki, line 494: I wonder if this needs pagination if we only use the first result... wdyt?

i4ki (changes requested):
I think it's very good overall but I found some issues and things to be discussed. The most important is the the comment related to `merged` state being set even if the request fails.

cmd/terramate/cli/cloud.go
sixstone-qq: we could set as the information we have `closed` without setting the `merged` state if unknown too. Both are fine for me

cmd/terramate/cli/cloud.go
```diff
@@ -404,69 +408,49 @@ func (c *cli) detectCloudMetadata() {
...
-			if _, found := uniqueReviewers[login]; found {
-				continue
-			}
-			uniqueReviewers[login] = struct{}{}
-
-			reviewRequest.Reviewers = append(reviewRequest.Reviewers, cloud.Reviewer{
-				Login:     login,
-				AvatarURL: review.User.AvatarURL,
-			})
+		merged, err := isGithubPRMerged(githubClient, r.Owner, r.Name, pull.GetNumber())
```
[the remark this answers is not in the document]
snakster: > if this instance of terramate run is running in the GHA, then querying their own "checks state" will always return 1 with the in_progress state, right?
Yes, that's actually something I was wondering myself. The checks are specific to a commit, so for my tests I was always doing `terramate run` on commits that already had their checks completed, but in case of GHA, the checks will probably still be in progress. In this case, the only solution would be that there's another mechanism to sync them later.

cmd/terramate/cli/cloud.go
i4ki: I meant that this implementation uses `merged=false` even if the request fails, which is misleading.

cmd/terramate/cli/cloud.go
snakster: Good point. Basically they are obtained just to print them for debug purposes, then everything is discarded but the first one. Maybe this function should just get the latest pull request in the first place.

cmd/terramate/cli/cloud.go
```diff
@@ -477,30 +457,87 @@ func (c *cli) detectCloudMetadata() {
...
 		return nil, err
 	}
 
-	return commit, nil
+	return rcommit, nil
 }
 
-func getGithubPR(ghClient *github.Client, repo string, commitName string) (github.Pulls, error) {
+// returns nil, nil if there was no PR associated with commit
+func getGithubPR(ghClient *github.Client, owner, repo, commit string) (*github.PullRequest, error) {
```
i4ki, line 473: I think it's better to change the return to be `(pull *github.PullRequest, found bool, err error)`.

In idiomatic Go, usually whenever `err == nil` then the value is ready to be used, so I would expect:
```go
pr, err := getGithubPR(client, owner, repo, commit)
if err != nil {
   // handle error: continue or return
}

st := pull.GetState() // pr can be safely used.
```

but for the case where the value is empty and no error occurs, then usually I return an additional boolean flag that hints if the value can be used. So it would become:

```go
pr, found, err := getGithubPR(...)
if err != nil {
   // handle err
}
if found {
   // safely to use pr
   pr.GetState()
}
```

i4ki (approved):
LGTM but I left a small improvement idea.

cmd/terramate/cli/cloud.go
wmalik, line 473: Would it make sense move this function (and the other github specific functions: `listGithubChecks`, `listGithubPullReviews` `isGithubPRMerged`) to the `cmd/terramate/cli/github` package?

wmalik (commented):
Looks good to me (just a minor question about code organization). Since we don't have tests at the moment, I am curious to know how you tested this as I also figuring out a testing strategy for the upcoming previews feature.

cmd/terramate/cli/cloud.go
snakster: Initially I agreed, but when I tried moving them I noticed that we end up with two `github` packages (our helpers and go-github), and in cloud.go we use both, which is also not great. So for now, I opted to not refactor this further.

wmalik (approved):

cmd/terramate/cli/cloud.go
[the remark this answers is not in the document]
snakster: Resolving this for now, we have to see how it works with some deployment runs in GHA.

Issue #1213, opened 2023-11-01
[FEATURE] preserve comment in the generated code


worawatwi:
**Is your feature request related to a problem? Please describe.**
We added comment to ignore specific tfsec rule but these comments were removed in the generated codes.

**Describe the solution you'd like**
Provide comment prefix or pattern indicated that the comment should not be removed in the generated code.

**Describe alternatives you've considered**
N/A.

vtimd:
Hey there @worawatwi, just wanted to acknowledge that we've seen this. We're looking into it internally and will get back to you. Thank you for bringing this to our attention.

wanderanimrod:
+1  I wanted to generate some code with `#tflint-ignore: <rule>` lines but I can't. Would be very helpful if this feature was added.

soerenmartius:
Thanks for reporting this issue and for upvoting it, too! We're looking into possible ways on how to solve this and will get back with an ETA once we aligned on a solution

soerenmartius:
Hi all,

Again, thanks for reporting this issue. We are currently looking into solutions. Our challenge is that the HCL parser needs to preserve comments for generated code by default. This issue has been reported and discussed a couple of times, and the community has provided various fixes. However, so far, HashiCorp hasn't addressed any of the [open PRs](https://github.com/hashicorp/hcl/pull/601) aimed at fixing that.

We are still evaluating possible workarounds and will get back to you with an update asap.

Thanks for being patient with us!

mariux:
as @soerenmartius mentioned, HCL does not support this out of the box and we will need to reimplement a bunch of things to make it work. 

Instead of just preserving any comments, we plan to support the given use cases via an `tflint` and `tfsec` integration in our code generation. The team is currently investigating the details and requirement of the tools and will continue the work on finding solutions.

solovyevt:
@mariux Just a reminder that `tfsec` is now being deprecated in favor of `trivy`, so it's probably worth focusing on the latter

frimik:
> Instead of just preserving any comments, we plan to support the given use cases via an `tflint` and `tfsec` integration in our code generation. 

Interesting ... `tflint` hints is exactly my codegen needs as well now that I recently migrated some (messy) `generate_file` codegens to `generate_hcl` (which is cleaner than "text templating"). May as well bump this issue in 2026 for the off chance that someone contributes? Thanks for everything Terramate so far ❤️ .

PR #1403, opened 2024-01-30
feat: add terramate.config.generate.hcl_magic_header_comment_style

### LINKED ISSUE #1310: [FEATURE] Use "#" for comments instead of "//"
[HashiCorp recommends](https://developer.hashicorp.com/terraform/language/syntax/configuration#comments) to use "#" for comments, but Terramate is using "//" [for its marker-comment](https://github.com/terramate-io/terramate/blob/50537c2abd4aae353b29a8313aa5acd986c585be/generate/genhcl/genhcl.go#L44). By doing so, we can no longer use tools like TFLint that contain [rules to enforce consistent comments](https://github.com/terraform-linters/tflint-ruleset-terraform/blob/main/docs/rules/terraform_comment_syntax.md).

Either fully switch to using "#" for Terramate-internal comments (which could be considered a breaking change since it marks all generated files as changed), or allow us to set the comment style in the terramate config-block.

The alternative is to completely disable the linting rules, which does not allow us to use consistent comment styles in our projects.

i4ki:

Introduces `terramate.config.generate.hcl_magic_header_comment_style` configuration option.
It allows for customizing the comment style used in the magic header of generate_hcl blocks.
It fixes the problem where linters complain about the default comment style used by Terramate (`//`).

Fixes #1310 

```
yes, a new block and attribute is added.
```


wmalik (commented):
Very nice! I left some comments.

generate/genhcl/genhcl.go
```diff
@@ -34,16 +34,29 @@ import (
...
+	condition         bool
+	asserts           []config.Assert
 }
 
+// CommentStyle is the configured comment style that must be generated.
+type CommentStyle int
+
+// Comment styles supported.
+const (
+	SlashComment CommentStyle = iota
```
sixstone-qq, line 50: In my opinion, if it's an enum, the documentation and examples should say only // and # are allowed.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 - Add `list --run-order` flag to list stacks in the order they would be executed.
 - Add support for deployment syncing to script commands.
 - Add `disable_safeguards` configuration option and CLI flag.
+- Add `terramate.config.generate.hcl_magic_header_comment_style` option for setting the generated comment style.
```
snakster, line 38: FYI the name of this field was not set in stone, in the refinement meeting we discussed that maybe we could also come up with a shorter name, i.e. `hcl_header_prefix`.

generate/genhcl/genhcl.go
i4ki: make sense, thanks!

generate/genhcl/genhcl.go
```diff
@@ -123,6 +136,41 @@ func (h HCL) String() string {
...
+	case "#":
+		return HashComment
+	default:
+		panic(errors.E(errors.ErrInternal, "invalid comment style"))
+	}
+}
+
+// CommentStyleFromConfig returns the CommentStyle from the configuration or the
+// default if not defined.
+func CommentStyleFromConfig(root *config.Root) CommentStyle {
```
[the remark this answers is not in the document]
wmalik: Ah got it, I had forgotten about that requirement, it makes sense to fetch the config in this function then.
However I suggest changing the type of the function arg because we don't need the complete `*config.Root` here:

```suggestion
func CommentStyleFromConfig(tmConfig *hcl.Terramate) CommentStyle {
```

wmalik (approved):

cwe1ss:
Since you mentioned in https://github.com/terramate-io/terramate/issues/1213#issuecomment-1911123042 that there will be some kind of integration with tflint/tfsec and since they also rely on comments, would it make more sense to call this field just `hcl_comment_style`?

i4ki:
Hi @cwe1ss 

This flag only controls the comment style used in the magic header generated in HCL files.
The #1213 is a different issue and AFAICS does not require a different option.

PR #1405, opened 2024-01-30
fix(generate): tm_dynamic.attributes over wrapped.

i4ki:

The attributes node were being wrapped with a CloneExpression node type over and over at each tm_dynamic generation for no reason, leading to OS stack exhaustion in the `ast.CloneExpr()` function. The node needs to be wrapped because PartialEval does in-place substitution then it needs to be cloned if used multiple times.

Fixes an issue reported internally.

```
no
```


wmalik (approved):
⚡

PR #1410, opened 2024-02-01
chore: use script run for running tests and linter.

i4ki:

EXPERIMENTAL IDEA: use `scripts` to run Terramate tests.
Rationale: using it in Terramate itself can help us finding issues quickly and also gather experience using the feature.

While doing this I found bug #1409 so this PR depends on that fix.

Fixes #

```

```


i4ki:
closing in favor of another approach

PR #1411, opened 2024-02-01
chore: ignore some paths in ci checks

wmalik:

This PR skips testing when there are no code changes. This should should speed up the following PR types:
* the PRs where only the markdown files in root are changed (e.g. CHANGELOG.md, README.md etc)
* The PRs where only `VERSION` is changed (e.g. https://github.com/terramate-io/terramate/pull/1407)

Changes in this PR:

* Ignore markdown files and `VERSION` in the ci job
* Move release-dryrun job to release-dryrun.yml
* Run release-dryrun.yml only when `VERSION` is changing.

```
No

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@
 name: ci
 on:
   pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'VERSION'
 
 jobs:
   checks:
--- a/.github/workflows/release-dryrun.yml
+++ b/.github/workflows/release-dryrun.yml
@@ -1,0 +1,37 @@
+# Copyright 2023 Terramate GmbH
+# SPDX-License-Identifier: MPL-2.0
+
+name: release-dryrun
+on:
+  pull_request:
+    paths:
+      - 'VERSION'
+      - 'Makefile'
+      - 'makefiles/**'
+      - '.goreleaser.yaml'
+      - 'hack/release/Dockerfile'
+
+jobs:
+  release_dry_run:
+    name: Release Dry Run
+
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+
+      - name: Configure asdf and plugins needed
+        uses: asdf-vm/actions/install@83133f03f5693901c2296a8e622955087dc20267
+
+      - name: check semver in VERSION
+        run: make version/check
+
+      - name: release dry run
+        run: terramate run --tags golang --changed -- make release/dry-run
--- a/hack/version-check
+++ b/hack/version-check
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+VERSION=$(cat VERSION)
+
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -4,6 +4,9 @@
 name: ci
 on:
   pull_request:
+    paths-ignore:
+      - '*.md'
+      - 'VERSION'
```
i4ki, line 9: I guess this will be good for speed up the release process, but if the VERSION file contains a non-semver string inside then tests fail.

.github/workflows/release-dryrun.yml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 Terramate GmbH
+# SPDX-License-Identifier: MPL-2.0
+
+name: release-dryrun
+on:
+  pull_request:
+    paths:
+      - 'VERSION'
```
i4ki, line 8: the `release-dryrun` must run on all events because it checks if Terramate is releasable. We can't just discover that Terramate is not releasable right at the time of releasing it, this surprise could cause us a lot of headache.

Like, if someone introduces a change in the `make build` that now makes some compiler flags required, this will break the release if the `goreleaser.yml` is not updated. There are many other ways of making the release broken, then I think we should continue running it in the PR on each push. 
It's fine moving to a separate workflow though, I think that makes sense.

i4ki (changes requested):

.github/workflows/release-dryrun.yml
wmalik: Good point, `VERSION` is indeed not enough and I agree about the correctness of the release building process.
However, most PRs are not changing the release process so it seems suboptimal to check it on every change.

We could add the `Makefile`, `makefiles/**`, `.goreleaser.yml` ro reduce the likelihood of breaking the build. What other files could possibly break the release building process?

i4ki: Okay, yes maybe we can optimize to check for those files. Additionally you also need to check for `hack/release/Dockerfile`.

The only other of breaking it, I guess, is if we move the `main.go` location of `cmd/terramate` or `cmd/terramate-ls`, but this is very unlikely.

Go ahead 👍

.github/workflows/ci.yml
wmalik: That issue can be solved by adding a `make version/check` that validates the semver string in `VERSION`.
However, I found another issue: when we bump the semver, it always causes the testdata to change, and then full ci checks in GHA would kick in, resulting in no speed up in the release process.

Do you know why the test data is tied to the `VERSION` string? 🤔 Would it be possible to not depend on `VERSION` to generate test data?

i4ki:
```
generate_file "cloud.data.json" {
  lets {
    version = tm_trimspace(tm_file("${terramate.root.path.fs.absolute}/VERSION"))
    well_known = {
      required_version = "= ${tm_replace(let.version, "/-.*/", "")}"
    }
  }
  content = tm_jsonencode({
    orgs       = global.testserver.orgs,
    users      = global.testserver.users,
    well_known = let.well_known,
  })
}
```

We use this to test the `/.well-known/cli.json` `required_version` field. So it uses the current version (without `-.*`) as basis.

i4ki: @waxb please change this to be:

```
   lets {
    version = "> 0.4.3"
    ...
   }
```

it should work for all newer releases, then there's no test dependency on `VERSION` anymore and we unblock this.

wmalik: Makes sense, fixed it here: https://github.com/terramate-io/terramate/pull/1423

snakster (approved):

PR #1412, opened 2024-02-01
chore: Use printer for all fatal messages

wmalik:

There were four kinds of fatal messages in the codebase:
* `fatal(err, msg)`
* `c.fatal(msg, err)`
* `errlog.fatal`
* `logger.Fatal`

All fatal messages have been standardized by using `fatal(err, msg)` which is an alias for `printer.Stderr.Fatal`.

```
Yes, the fatal messages are more readable.
```


cmd/terramate/cli/cloud.go
```diff
@@ -88,7 +88,7 @@ func (c *cli) disableCloudFeatures(err error) {
 func (c *cli) handleCriticalError(err error) {
 	if err != nil {
 		if c.uimode == HumanMode {
-			fatal(err)
+			fatal("Aborting", err)
```
i4ki, line 91: nitpick: for consistency with the other messages I think this should not be capitalized.

cmd/terramate/cli/fmt.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package cli
+
+import "fmt"
+
+func sprintf(format string, a ...interface{}) string {
+	return fmt.Sprintf(format, a...)
+}
```
i4ki, line 10: I don't think it makes sense a separate file just for this function. Also the name `fmt.go` is misleading, as we have a `terramate fmt` feature. You know, we do fuzzy file search when looking for the file that implements feature XYZ, then this will pop up very often but it's actually that kind of file that would be rarely open.

i4ki (approved):
no blocker but I added some minor comments.

cmd/terramate/cli/fmt.go
wmalik: that makes sense, I could move it to the bottom of `cmd/terramate/cli/cli.go`?

i4ki (approved):
LGTM

PR #1413, opened 2024-02-01
chore: refactor run.Sort stack ordering

snakster:

The list of stacks that needs to be ordered for `script run` execution uses different data types than for `run`, which makes it difficult to re-use. This PR makes the sorting function more flexible so it can be used with both data types.

* The interface of the sort function changes from returning the sorted list as a new list, to sorting in-place. However, even before the passed list was modified with `sort.Sort()`, so this is at least obvious now.

```
no
```

### DIFF
--- a/run/order.go
+++ b/run/order.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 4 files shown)


run/order.go
```diff
@@ -70,46 +77,40 @@ func Sort(root *config.Root, stacks config.List[*config.SortableStack]) (config.
...
-	for _, id := range order {
+	orderLookup := make(map[string]int, len(order))
+	for idx, id := range order {
 		val, err := d.Node(id)
 		if err != nil {
-			return nil, "", fmt.Errorf("calculating run-order: %w", err)
+			return "", fmt.Errorf("calculating run-order: %w", err)
 		}
 		s := val.(*config.Stack)
-		if !isSelectedStack(s) {
```
i4ki, line 106: can you explain why this can be removed now? because `BuildDAG()` can pull more stacks into the pool, then if we don't de-select the ones out of scope, the returned order will be wrong.

i4ki (changes requested):

run/order.go
snakster: As you said, before, we used a given list of stacks to construct a DAG (which may pull in more stacks), which gives us the ordering, then we de-select those that were not in the original list.

What we do now is using the ordering the DAG gives to map each entry of the original list to an ordering index. Then we use this index for sorting the original list.

Before, we were iterating `order`, which may contain more stacks than the original list, but now we just sort the original list, so nothing has to be removed there. The `orderLookup` map may contain entries for stacks that have been pulled in, but they don't affect the sorted list.

i4ki: ahhh yes now we do sort in-place in the provided items. Gotcha. Thanks for the explanation!

i4ki (approved):
LGTM

Goop-arsh (commented):
R

PR #1422, opened 2024-02-05
fix: use req.GetBody when dumping req

wmalik:

The `Request.Clone()` function is misleading, it does not create a copy of the original request's body, resulting in dumprequest+printf draining the original request's body.

See https://github.com/golang/go/issues/36095

```
No.
```


i4ki (approved):
LGTM

PR #1423, opened 2024-02-05
chore: remove testdata dependency on VERSION file

wmalik:

Removing testdata dependency on the `VERSION` file will allow us to speedup the release process a bit, see https://github.com/terramate-io/terramate/pull/1411

```
No.
```


i4ki (approved):
LGTM

thanks

PR #1436, opened 2024-02-08
feat: add support for submitting previews to tmc

wmalik:

* Add flag `--cloud-sync-preview` to `terramate run` command
* Add an interop test
* Create a new package `cloud/previews` to handle the preview creation and add unit tests (using a fake HTTP transport)
* The end-to-end test implementation is skipped for now but there is some test coverage via the unit tests + interop test

* The functionality to send the stack changeset and stack logs to TMC will be implemented in a follow-up PR

```
Yes, a new flag was added to the `terramate run` command.
```


sixstone-qq (commented):
Few comments , let me know what you think of them

cloud/previews/previews.go
```diff
@@ -0,0 +1,179 @@
...
+	// loop over all affected stacks, if an item is present in the
+	// previewStacksMap, use the preview status and cmd from there
+	for _, affectedStack := range opts.AffectedStacks {
+		tags := []string{}
+		if len(affectedStack.Tags) > 0 {
+			tags = affectedStack.Tags
+		}
+
+		stack := cloud.PreviewStack{
+			PreviewStatus:   PreviewStatusAffected,
```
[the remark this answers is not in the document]
wmalik: Pending stacks are a subset of the `affectedStacks`, if there is a stack that's part of the preview run it is updated to the pending status on L132.

cmd/terramate/cli/run.go
```diff
@@ -421,3 +434,65 @@ func (c *cli) loadAllStackEnvs(runs []runContext) (map[prj.Path]runutil.EnvVars,
...
+		fatal("env var GITHUB_EVENT_PATH not found, not generating previews", nil)
+	}
+
+	prUpdatedAt := time.Now().UTC().Unix()
+	if eventPRUpdatedAt := github.GetEventPRUpdatedAt(githubEventPath); eventPRUpdatedAt != nil {
+		prUpdatedAt = eventPRUpdatedAt.Unix()
+	}
+
+	technology := "other"
+	technologyLayer := "other"
```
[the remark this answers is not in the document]
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1436/commits/1a66efdf34c521f5cdec82394f629fb9a1fcb657 (will rebase shortly)

cloud/types.go
```diff
@@ -220,15 +220,16 @@ type (
...
+		Platform          string    `json:"platform"`
+		Repository        string    `json:"repository"`
+		CommitSHA         string    `json:"commit_sha"`
+		Number            int       `json:"number"`
+		Title             string    `json:"title"`
+		Description       string    `json:"description"`
+		URL               string    `json:"url"`
+		Labels            []Label   `json:"labels,omitempty"`
+		Reviewers         Reviewers `json:"reviewers,omitempty"`
+		PlatformUpdatedAt time.Time `json:"platform_updated_at"`
```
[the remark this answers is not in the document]
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1436/commits/1a66efdf34c521f5cdec82394f629fb9a1fcb657 (will rebase shortly)

cloud/previews/previews.go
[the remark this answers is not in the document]
sixstone-qq: but they are being set to any value, should it be better to limit to pending there only. If not , it would be potentially be a Bad request.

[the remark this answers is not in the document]
wmalik: makes sense, fixed.

sixstone-qq (approved):
From my understanding in communication with TMC. All good now

snakster (commented):
Looks great, really well done 👍 

One thing I noticed is that your approach for testing with fake transport is new, wouldn't it be more consistent to add it to the e2etest like the rest of the cloud tests, or is there a specific reason you chose this approach?

wmalik:
> Looks great, really well done 👍
> 
> One thing I noticed is that your approach for testing with fake transport is new, wouldn't it be more consistent to add it to the e2etest like the rest of the cloud tests, or is there a specific reason you chose this approach?

@snakster I chose that approach mainly because it was faster to implement. Another reason is that those tests are only testing the `previews` package and hence can be used for fine grained unit tests.

However, I do plan to add end to end tests for this once the MVP has been validated/tested.

snakster (approved):

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+	PreviewStatusAffected = "affected"
+	// PreviewStatusPending is the status for a stack that is selected in a preview run
+	PreviewStatusPending = "pending"
+)
+
+// RunContext is the context for a run
+type RunContext struct {
+	Stack *config.Stack
+	Cmd   []string
+}
```
i4ki, line 27: I don't get why this is leaking here. This type is a helper type of the cli package, also used by the other sync features.

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+	AffectedStacks  map[string]*config.Stack
+	OrgUUID         cloud.UUID
+	UpdatedAt       int64
+	Technology      string
+	TechnologyLayer string
+	Repository      string
+	DefaultBranch   string
+	ReviewRequest   *cloud.DeploymentReviewRequest
+	Metadata        *cloud.DeploymentMetadata
+}
```
i4ki, line 41: this is a mix of required and optional fields so I would not call it `...Opts`.
I think it's better to pass the required information explicitly as function parameters and then this `Opt` type can be used for options.

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+	// previewStacksMap, use the preview status and cmd from there
+	for _, affectedStack := range opts.AffectedStacks {
+		tags := []string{}
+		if len(affectedStack.Tags) > 0 {
+			tags = affectedStack.Tags
+		}
+
+		stack := cloud.PreviewStack{
+			PreviewStatus:   PreviewStatusAffected,
+			Repository:      opts.Repository,
```
i4ki, line 120: this is not validated. Is repo okay to be empty?

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+		Technology:      opts.Technology,
+		TechnologyLayer: opts.TechnologyLayer,
+		ReviewRequest:   opts.ReviewRequest,
+		Metadata:        opts.Metadata,
+		Stacks:          []cloud.PreviewStack{},
+	}
+
+	previewStacksMap := map[string]RunContext{}
+	for _, run := range opts.Runs {
+		previewStacksMap[run.Stack.ID] = run
```
i4ki, line 107: TODO: duplication

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+		StackPreviewsByMetaID: stacks,
+	}, nil
+}
+
+func (o CreatePreviewOpts) validate() error {
+	errs := errors.L()
+
+	if len(o.AffectedStacks) == 0 || len(o.Runs) == 0 {
+		errs.Append(errors.E("no affected stacks or runs provided"))
+	}
```
i4ki, line 171: A deployment is created even when no stacks are selected. What was decided in the case of previews? is this never going to happen? is this going to be handled by the GHA script?

cloud/previews/previews_test.go
```diff
@@ -0,0 +1,164 @@
...
+func makeAffectedStacks(num int, isChanged bool) map[string]*config.Stack {
+	affectedStacks := map[string]*config.Stack{}
+	for i := 0; i < num; i++ {
+		affectedStacks[strconv.Itoa(i)] = &config.Stack{
+			Dir:         project.NewPath("/somepath" + strconv.Itoa(i)),
+			ID:          uuid.NewString(),
+			Name:        "stack" + strconv.Itoa(i),
+			Description: "desc of stack" + strconv.Itoa(i),
+			Tags:        []string{"tag1", "tag2"},
+			IsChanged:   isChanged,
```
i4ki, line 132: why `IsChanged` is needed here? in the case of deployments, it never matters the reason why the stack was selected.

cloud/previews/previews_test.go
```diff
@@ -0,0 +1,164 @@
...
+			numRunContexts: 3,
+			want: want{
+				numStacksReturned: 3,
+				httpEndpointsCalled: []string{
+					"POST /v1/previews/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
+				},
+				err: nil,
+			},
+		},
+	}
```
i4ki, line 68: only the *three stacks* is testing a real case. 
The first is a valid case (should not error), the second never happens because cli ensures the organization is selected. 

I think this testing approach is brittle and is not testing anything..
better do it later with the `testserver` or `interop` package.

cloud/types.go
```diff
@@ -220,23 +220,24 @@ type (
 
 	// DeploymentReviewRequest is the review_request object.
 	DeploymentReviewRequest struct {
```
i4ki, line 222: Now that this is reused better just call it `ReviewRequest`. I was confused why this change here if the feature is not related to deployment.

cloud/types_previews.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+		if stack.Path == "" {
+			errs.Append(errors.E("path is required for stack[%d]", i))
+		}
+		if stack.DefaultBranch == "" {
+			errs.Append(errors.E("default_branch is required for stack[%d]", i))
+		}
+		if stack.MetaID == "" {
+			errs.Append(errors.E("meta_id is required required for stack[%d]", i))
+		}
```
i4ki, line 86: these validations are repeated because the `Stack` type is not being reused.
The error messages are different and some validations are not in place (`meta_id` not lowercase).

cloud/types_previews.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return errs.AsError()
+}
+
+// Validate the CreatePreviewPayloadRequest object.
+func (r CreatePreviewPayloadRequest) Validate() error {
+	errs := errors.L()
+	if r.Technology == "" {
+		errs.Append(errors.E("technology is required"))
```
i4ki, line 99: please follow the error message convention: https://github.com/terramate-io/terramate/blob/main/cloud/types.go#L411-L423

```
errs.Append(errors.E(`missing "technology" field`))
```

no problem changing the way, but it would be good if we return similar messages for similar errors.

cmd/terramate/cli/cli.go
```diff
@@ -2112,15 +2113,18 @@ func (c *cli) friendlyFmtDir(dir string) (string, bool) {
...
 		return nil, err
 	}
 
 	c.gitFileSafeguards(ensureCleanRepo)
 
-	entries := c.filterStacks(report.Stacks)
+	entries := report.Stacks
+	if filterStacks {
+		entries = c.filterStacks(report.Stacks)
+	}
```
i4ki, line 2127: I see that you want to compute the selected stacks twice, first filtering (by changed, tags, workingdir, cloud status, etc) and then later without some filters but allowing others (changed, cloud status). This seems broken.

Additionally, this logic requires a lot of computation, invoking of lots of git commands, parsing of Terraform modules, etc, then doing this twice is very bad.

cmd/terramate/cli/github/event.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, err
+	}
+
+	var parsedEvent event
+	if err := json.Unmarshal(bytes, &parsedEvent); err != nil {
+		return nil, err
+	}
+
+	return &parsedEvent, nil
+}
```
i4ki, line 46: this function returns an error that's always ignored.

cmd/terramate/cli/github/event.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// GetEventPRUpdatedAt returns the updated_at field from the file at `eventPath`.
+// The file is expected to be a JSON file containing the GitHub event for a pull
+// request.
+func GetEventPRUpdatedAt(eventPath string) *time.Time {
+	event, err := getEventFromPath(eventPath)
+	if err != nil {
+		return nil
+	}
```
i4ki, line 19: I think we should at least add a warning here.

cmd/terramate/cli/run.go
```diff
@@ -421,3 +434,64 @@ func (c *cli) loadAllStackEnvs(runs []runContext) (map[prj.Path]runutil.EnvVars,
...
+	}
+
+	affectedStacksMap := map[string]*config.Stack{}
+	for _, st := range affectedStacks {
+		affectedStacksMap[st.Stack.ID] = st.Stack
+	}
+
+	githubEventPath, ok := os.LookupEnv("GITHUB_EVENT_PATH")
+	if !ok {
+		fatal("env var GITHUB_EVENT_PATH not found, not generating previews", nil)
```
i4ki, line 459: The `sync` features must never abort without executing the provided `run` or `script` command. It must warn and just disable the feature. Check the `c.disableCloudFeatures(err)`.

cmd/terramate/cli/run.go
```diff
@@ -421,3 +434,64 @@ func (c *cli) loadAllStackEnvs(runs []runContext) (map[prj.Path]runutil.EnvVars,
...
+	}
+
+	prUpdatedAt := time.Now().UTC().Unix()
+	if eventPRUpdatedAt := github.GetEventPRUpdatedAt(githubEventPath); eventPRUpdatedAt != nil {
```
i4ki, line 463: it would be nice if we can have a testing env var for injecting the updated at.
```
$ make build/test
$ TM_TEST_PREVIEW_UPDATED_AT=... ./bin/terramate ...
```

cmd/terramate/cli/run.go
```diff
@@ -421,3 +434,64 @@ func (c *cli) loadAllStackEnvs(runs []runContext) (map[prj.Path]runutil.EnvVars,
...
+			Technology:      technology,
+			TechnologyLayer: technologyLayer,
+			Repository:      c.prj.prettyRepo(),
+			DefaultBranch:   c.prj.gitcfg().DefaultBranch,
+			ReviewRequest:   c.cloud.run.reviewRequest,
+			Metadata:        c.cloud.run.metadata,
+		},
+	)
+	if err != nil {
+		fatal("unable to create cloud preview", err)
```
i4ki, line 491: it should never abort.

makefiles/unix.mk
```diff
@@ -38,6 +38,7 @@ test/helper:
 tempdir=$(shell ./bin/helper tempdir)
 test: test/helper build
 # 	Using `terramate` because it detects and fails if the generated files are outdated.
+	./bin/terramate run --no-recursive -- go test -c -tags interop ./cmd/terramate/e2etests/cloud/interop/...
```
i4ki, line 41: this must be reverted. The interop tests must be explicit invoked, as they are very slow and require proper environment set up.

i4ki (commented):
Some important details need to be addressed.

(I removed most of the nitpick comments because it's already merged but I think we are diverging a lot from the other cloud implementations and it would be good if those changes comes with a good reasoning)

makefiles/unix.mk
i4ki: just noticed the `-c` in the `go test`. Why this is needed?

wmalik: since the interop tests have a build tag, they are not built  by `go test` and syntax errors won't be caught by the CI. 
`go test -c` only builds the test (it does not run them).

cmd/terramate/cli/run.go
wmalik: In this case disabling does not make sense because `--cloud-sync-preview` feature requires a timestamp to create a preview.
However, we could handle that case gracefully by using `time.Now()` and removing the `fatal`, what do you think?

cloud/types_previews.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	// CreatePreviewPayloadRequest is the request payload for the creation of
+	// stack deployments.
+	CreatePreviewPayloadRequest struct {
+		UpdatedAt       int64                    `json:"updated_at"`
+		Technology      string                   `json:"technology"`
+		TechnologyLayer string                   `json:"technology_layer"`
+		ReviewRequest   *DeploymentReviewRequest `json:"review_request,omitempty"`
+		Metadata        *DeploymentMetadata      `json:"metadata,omitempty"`
+		Stacks          PreviewStacks            `json:"stacks"`
```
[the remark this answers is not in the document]
wmalik: There is no usecase for `nil` stacks currently. There is a validation for it in the `Validate()` method in case its ever set as `nil`. The API marks this as a required attribute. Could you please link to the case where it is marked as omitempty? I don't see a problem with marking this as `omitempty` though if it's safer for the API.

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+	"time"
+
+	"github.com/terramate-io/terramate/cloud"
+	"github.com/terramate-io/terramate/config"
+	"github.com/terramate-io/terramate/errors"
+)
+
+const (
+	// PreviewStatusAffected is the status for a stack that is affected in a PR
+	PreviewStatusAffected = "affected"
```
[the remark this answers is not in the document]
wmalik: That makes sense, I already addressed it in the previews changeset PR, see https://github.com/terramate-io/terramate/pull/1443/files#diff-7f1a481103c73ed166b25cfecbd5d9fc8b35ede4e2c89dc73d5a373192cba8e5

cloud/previews/previews.go
wmalik: The preview package is unware of the `cli.runContext` type. The `preview.CreatePreview` function needs information about the all the scripts+commands that are part of a `--cloud-sync-preview` invocation, and this type + `[]RunContext` captures that information. It could have a better name though to reduce the confusion.

cloud/previews/previews.go
wmalik: its validated here: https://github.com/terramate-io/terramate/blob/22329f6c6657eb9afe9462f48236bb1053eea6b8/cloud/types_previews.go#L75

but I will change that to use the existing `Stacks` validation to remove the duplication.

cloud/previews/previews.go
wmalik: Good question, I will think about this one and get back to you.

cloud/previews/previews.go
wmalik: Could you please elaborate?

cloud/previews/previews.go
wmalik: main motivation to add that struct was to have a smaller arg list. I would lean towards removing the struct all together and passing all of that information as function args instead of differentiating between optional/mandatory args because it's not meaningful IMO to mark optional/non-optional args. What do you think?

cloud/previews/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright 2024 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+// Package previews contains functionality for previews.
+package previews
```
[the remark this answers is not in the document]
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1451

cloud/previews/previews.go
```diff
@@ -0,0 +1,178 @@
...
+		previewStacksMap[run.Stack.ID] = run
+	}
+
+	// loop over all affected stacks, if an item is present in the
+	// previewStacksMap, use the preview status and cmd from there
+	for _, affectedStack := range opts.AffectedStacks {
+		tags := []string{}
+		if len(affectedStack.Tags) > 0 {
+			tags = affectedStack.Tags
+		}
```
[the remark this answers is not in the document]
wmalik: Removed in https://github.com/terramate-io/terramate/pull/1451

cloud/types_previews.go
```diff
@@ -0,0 +1,142 @@
...
+type (
+	// PreviewStacks is a list of stack objects for the request payload
+	PreviewStacks []PreviewStack
+
+	// PreviewStack represents the stack object of the request payload
+	// type for the creation of stack deployments.
+	PreviewStack struct {
+		PreviewStatus   string   `json:"preview_status"`
+		Repository      string   `json:"repository"`
+		Path            string   `json:"path"`
```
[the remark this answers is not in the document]
wmalik: Makes sense, fixed in https://github.com/terramate-io/terramate/pull/1451

cloud/previews/previews_test.go
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1451

cloud/types.go
wmalik: Fixed in https://github.com/terramate-io/terramate/pull/1451

cloud/types_previews.go
wmalik: I embedded the `Stack` tupe into the PreviewStack struct and used the existing Stack.Validate() method to remove the duplication in https://github.com/terramate-io/terramate/pull/1451

cloud/types_previews.go
wmalik: Makes sense to stay consistent, fixed in https://github.com/terramate-io/terramate/pull/1451

cmd/terramate/cli/run.go
wmalik: `TM_IGNORE_EVENT_PATH`

cloud/previews/previews.go
wmalik: I think the preview should be created even if there are no selected stacks, so I removed the validation.

PR #1450, opened 2024-02-15
feat: add support for parallel run

snakster:
This PR adds a new option flag `--parallel` (short `-j`) to `terramate run / script run` to run independent stacks in parallel. Ordering constraints (parent/child folders, `before`/`after`) are still respected.
The number of concurrent runs can be set explicitly with `--parallel=N`. By default, N=5.
There is no grouping of command output for parallel runs, so output lines may be printed interleaved.

The goal of this is to speed up running multiple stacks.

* To adapt some tests to work with `--parallel`, explicit ordering has been added between stacks. In some cases, the layout string was modified later, so adding `after=[...]` did not always work. For this reason, if there's a required order between s1 and s2, I moved s2 to s1/s2.
* The `resource.Throttled` type is not used by anything yet, I added it just as a proof-of-concept.

```
Yes, a new flag is added to run and script run.
```

### DIFF
--- a/cmd/terramate/e2etests/cmd/helper/main.go
+++ b/cmd/terramate/e2etests/cmd/helper/main.go
@@ -14,6 +14,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	tfjson "github.com/hashicorp/terraform-json"
@@ -63,6 +64,8 @@ func main() {
 		stackRelPath(os.Args[2])
 	case "tf-plan-sanitize":
 		tfPlanSanitize(os.Args[2])
+	case "fibonacci":
+		fibonacci()
 	default:
 		log.Fatalf("unknown command %s", os.Args[1])
 	}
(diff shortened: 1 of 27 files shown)


scheduler/resource/throttled.go
```diff
@@ -0,0 +1,37 @@
...
+	"time"
+)
+
+// Throttled in a rate-limited resource.
+type Throttled struct {
+	ticker *time.Ticker
+}
+
+// NewThrottled creates a resource with the given rate limit.
+func NewThrottled(requestsPerSecond int64) *Throttled {
```
i4ki, line 17: this is not used, right?

i4ki (commented):
Looks very good!
This is just an initial review of the implementation part.
I will look into the testcases tomorrow.

scheduler/parallel.go
```diff
@@ -0,0 +1,112 @@
...
+
+		for _, succ := range st.successors {
+			// Join paths
+			c := succ.tickCount.Add(1)
+			if c >= succ.tickThreshold {
+				s.visitNode(succ, f)
+			}
+		}
+	})
+}
```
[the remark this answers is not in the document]
snakster: Makes sense, I used this naming in the past, but it's misleading.

scheduler/resource/throttled.go
snakster: Yep, this is not used. I also mentioned it in the PR notes. I added it during development as proof-of-concept, also did some testing with it, so I figured might as well leave it in since it could be useful in the future and its not much code. But could also remove it if you think it doesn't belong with this PR.

run/order.go
```diff
@@ -105,9 +103,111 @@ func Sort[S ~[]E, E any](root *config.Root, items S, getStack func(E) *config.St
 	return "", nil
 }
 
+// BuildDAGFromStacks computes the final, reduced dag for the given list of stacks.
```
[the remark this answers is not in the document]
snakster: I refactored it to move the shared part to a separate function instead.

scheduler/resource/throttled.go
i4ki: no, you can leave it as it could definitely be useful in the future.

cmd/terramate/e2etests/cmd/helper/main.go
```diff
@@ -125,6 +128,41 @@ func tempDir() {
 	fmt.Print(tmpdir)
 }
 
+// fibonacci, when called from dir fib.N/, writes the Nth fibonacci number to ./fib.txt.
+// It may try to read values from ../fib.N-1/fib.txt and ../fib.N-2/fib.txt, which were previously
+// created by running this command in other dirs.
+func fibonacci() {
```
i4ki, line 134: nice haha

i4ki (dismissed):
LGTM

Well done @snakster

PR #1451, opened 2024-02-16
refactor: improve previews abstraction

wmalik:

This PR addresses the concerns raised in the following PR:

* https://github.com/terramate-io/terramate/pull/1436

```
No.

```


i4ki (approved):
LGTM

i4ki (approved):
LGTM

cmd/terramate/cli/run.go
```diff
@@ -454,14 +453,16 @@ func (c *cli) createCloudPreview(runs []runContext) map[string]string {
...
-		prUpdatedAt = eventPRUpdatedAt.Unix()
+	githubEventPath, ok := os.LookupEnv("GITHUB_EVENT_PATH")
+	if ok {
+		eventPRUpdatedAt := github.GetEventPRUpdatedAt(githubEventPath)
+		if eventPRUpdatedAt != nil {
+			prUpdatedAt = eventPRUpdatedAt.Unix()
+		}
+	} else {
+		printer.Stderr.Warn(
+			sprintf("env var GITHUB_EVENT_PATH not found, using %d as updated_at", prUpdatedAt))
```
mariux, line 465: We can not fall back here, we will need the event.

mariux (dismissed):
we can not fall back when there is no event payload.

cloud/cloud_preview_test.go
```diff
@@ -1,9 +1,13 @@
 // Copyright 2024 Terramate GmbH
 // SPDX-License-Identifier: MPL-2.0
 
-package previews_test
+package cloud_test
```
mariux, line 4:
```suggestion
package cloud_preview_test
```

cmd/terramate/cli/run.go
i4ki: we discussed in a long call today about introducing a `TM_TEST_IGNORE_GITHUB_EVENT_FILE` (that's only available in a test build) for making it easy to locally test this and then fallback will happen only in this case.
We agree that a silent fallback here is not correct.

I think @wmalik will introduce this test env in the new PR.

PR #1454, opened 2024-02-19
chore: update go-git to version 5.11

snakster:

There's a critical security vulnerability in go-git versions prior to 5.11 (see https://nvd.nist.gov/vuln/detail/CVE-2023-49569). The vulnerability doesn't affect us, since we only use go-git for gitignore parsing, but we should update the version to avoid security alerts or in case we may make extended use of go-git in the future.

```
no
```


wmalik (approved):

i4ki (approved):
LGTM

PR #1458, opened 2024-02-20
feat(scripts): add cloud_sync_preview + cloud_sync_layer

wmalik:

Add `cloud_sync_preview` option to script block.

Depends on https://github.com/terramate-io/terramate/pull/1493

```
Yes, users will be able to use the `cloud_sync_preview` option in the script block to store previews in TMC.
```


i4ki (changes requested):
Just the layer validation, IMO just use the layer type validation. Also I think makes sense to put the correct type in the `config. ScriptCmdOptions`.

i4ki (approved):
LGTM

PR #1461, opened 2024-02-21
fix: use testdata/pull_request_event.json in interop tests

wmalik:

Use a dummy github event file in interop tests, so they can be run when a real
github pull request event is not available. (e.g. when running tests locally, or
on a branch that is not a pull request)

```
No.
```


i4ki (approved):
LGTM

PR #1463, opened 2024-02-21
chore: improve change log

mariux:


i4ki (approved):
LGTM

snakster:
It's a nitpick but I think the trailing `.` should be consistent, otherwise it looks unpolished. Since we can have entries that have multiple sentences, IMO we should just use it everywhere.

PR #1466, opened 2024-02-22
fix: use PR info from event when available, otherwise API

wmalik:

Read PR information from the GHA pull_request event when available, otherwise
fall back to the Github API.

The fallback only happens when building the ReviewRequest object in
detectCloudMetadata().

There is no fallback for `preview.updated_at`.

```
No.
```

### DIFF
--- a/cmd/terramate/cli/cloud.go
+++ b/cmd/terramate/cli/cloud.go
@@ -21,6 +21,7 @@ import (
 	"github.com/terramate-io/terramate/cloud/preview"
 	"github.com/terramate-io/terramate/cloud/stack"
 	"github.com/terramate-io/terramate/cmd/terramate/cli/clitest"
+	tmgithub "github.com/terramate-io/terramate/cmd/terramate/cli/github"
 
 	"github.com/terramate-io/terramate/cmd/terramate/cli/out"
 	"github.com/terramate-io/terramate/config"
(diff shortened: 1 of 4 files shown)


cmd/terramate/cli/cloud.go
```diff
@@ -484,7 +485,15 @@ func (c *cli) detectCloudMetadata() {
...
-	if pull, found, err := getGithubPR(githubClient, r.Owner, r.Name, headCommit); err == nil {
+	var pullRequest *github.PullRequest
+	pullRequest, err = tmgithub.GetEventPR()
+	if err != nil {
+		printer.Stderr.WarnWithDetails(
+			sprintf("unable to read pull_request from GITHUB_EVENT_PATH, falling back to github API"),
+			err)
+	}
+
+	if pull, found, err := getGithubPR(githubClient, r.Owner, r.Name, headCommit); err == nil && pullRequest == nil {
```
i4ki, line 496: Am I missing something or the retrieved `pullRequest` from the event file is not being used and GH API call is done anyway? I would expect this line to be in the `else` if the `tmgithub.GetEventPR()` fails.

i4ki (changes requested):
Can you clarify the implementation?
Also, now it seems we are parsing the event file twice..

cmd/terramate/cli/cloud.go
wmalik: You are right, that's broken. I'll make this PR a draft for now and make changes to parse the file only once.

wmalik: fixed

i4ki (approved):
lgtm

PR #1471, opened 2024-02-23
fix: rename PushedAt to UpdatedAt in cloud.ReviewRequest

wmalik:
Also set the value for UpdatedAt in the review request sent to the
cloud (it was previously unset).


i4ki (dismissed):
LGTM

cmd/terramate/cli/cloud.go
```diff
@@ -768,6 +768,7 @@ func setGithubPRMetadata(md *cloud.DeploymentMetadata, pull *github.PullRequest)
 }
 
 func (c *cli) newReviewRequest(pull *github.PullRequest, reviews []*github.PullRequestReview, checks []*github.CheckRun, merged bool) *cloud.ReviewRequest {
+	pullUpdatedAt := pull.GetUpdatedAt()
```
sixstone-qq, line 771: Could you remove the metadata sent the same information?

sixstone-qq: If it's not delayed, I'm ok with it but sending same information twice is misleading and a waste of network resources :)

wmalik: I have added it to our backlog and will discuss with the team today.

sixstone-qq: In this case, we should make it consistent and send always this information in the same way to cloud.

PR #1475, opened 2024-02-26
chore: log debug messages when creating preview

wmalik:

* Use debug logs for printing stack preview status update messages.
* Use debug logs when a PR cannot be fetched
* Print the `Entering stack` messages before executing cloud sync logic.

```
No.
```

### DIFF
--- a/cmd/terramate/cli/cloud.go
+++ b/cmd/terramate/cli/cloud.go
@@ -303,12 +303,10 @@ func (c *cli) doPreviewBefore(run stackCloudRun) {
 		printer.Stderr.ErrorWithDetails("failed to update stack preview", err)
 		return
 	}
-	if !c.parsedArgs.Quiet {
-		printer.Stderr.Println(sprintf("terramate: stack:'%s' preview_status:%s",
-			run.Stack.Name,
-			preview.StackStatusRunning,
-		))
-	}
+	log.Debug().
+		Str("stack_name", run.Stack.Dir.String()).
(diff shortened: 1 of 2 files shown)


cmd/terramate/cli/cloud.go
```diff
@@ -347,14 +345,15 @@ func (c *cli) doPreviewAfter(run stackCloudRun, res runResult) {
 	}
 
 	if !c.parsedArgs.Quiet {
```
i4ki, line 347: the remove this `if` as well because the `--quiet` only controls the printing messages, not the logging.

cmd/terramate/cli/run.go
```diff
@@ -327,6 +327,11 @@ func (c *cli) runAll(
 				printScriptCommand(c.stderr, run.Stack, task)
 			}
 
+			if !opts.Quiet && !opts.ScriptRun {
+				printer.Stderr.Println(printPrefix + " Entering stack in " + run.Stack.String())
+				printer.Stderr.Println(printPrefix + " Executing command " + strconv.Quote(cmdStr))
+			}
+
```
i4ki, line 334: Why this was moved up? now it prints `Executing command ...` but it's not about to run but still need to execute cloud related stuff. So maybe the user can see things like:
```
Executing command ...
<cloud related print or cloud error>
<command is executed>
```

i4ki (changes requested):

i4ki (approved):
LGTM

PR #1498, opened 2024-02-28
fix(preview): draft field not sent when false.

i4ki:

The `draft` field must not have an `omitempty` JSON tag because we need to explicitly tell the backend when the draft status changed.

no

```
no
```


sixstone-qq (approved):
Good catch!

wmalik (approved):
God catch

PR #1503, opened 2024-02-29
feat: add --cloud-sync-layer to terramate run

wmalik:

Add `-cloud-sync-layer` to `terramate run`. This would allow users to create
previews for a specific terraform workspace e.g. `staging`, `prod` etc.

```
Yes, a new flag --cloud-sync-layer has been added to terramate run.
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/google/uuid"
 	hhcl "github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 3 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -738,6 +740,22 @@ func (s *kongParallelFlag) Decode(ctx *kong.DecodeContext) error {
...
 
+type cloudSyncLayer string
+
+func (csl cloudSyncLayer) String() string {
+	return string(csl)
+}
+
+func (csl cloudSyncLayer) Validate() error {
+	for _, c := range csl {
+		if !unicode.IsLetter(c) && !unicode.IsDigit(c) && c != '-' {
```
i4ki, line 751: The format must be `[a-zA-Z]{1}[a-zA-Z0-9-]`. It cannot start with numbers or dash, otherwise it could create layers like `custom:1` or `custom:-` ...

i4ki (changes requested):
I think the layer validation needs improvement, other than that it looks good.

cmd/terramate/cli/cli.go
wmalik: The layers are supposed to be used to represent terraform workspaces. I just tested the `terraform workspace new NAME` command and was able to create with the following names:

image


So even the alphanumerc validation in this PR is more restrictive than terraform, and with the regex that you suggested if a user has a workspace called `1` or `1-dev`, or `2-dev`, they wouldn't be able to use the previews feature.

i4ki (approved):
LGTM

i4ki (approved):

PR #1524, opened 2024-03-05
chore: support -j N syntax for parallel flag and remove default value

snakster:

This PR adds support for supplying a parameter to the short form of the `--parallel`flag, i.e. `-j N`.
It also removes the default value, so the long version always needs a value.

```
no
```


i4ki (approved):
LGTM

PR #1525, opened 2024-03-05
feat(cloud): add review_decision to metadata

snakster:

This PR adds a new field `review_decision` to the cloud metadata. 

```
no
```


cmd/terramate/cli/cloud.go
```diff
@@ -645,6 +667,34 @@ func getGithubPRByCommit(ghClient *github.Client, owner, repo, commit string) (*
...
 }
 
+func getGithubPRReviewDecision(qlClient *githubql.Client, owner, repo string, pullNumber int) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultGithubTimeout)
+	defer cancel()
+
+	var q struct {
+		Repository struct {
+			PullRequest struct {
+				ReviewDecision string
```
i4ki, line 677: this can be `null`.

cloud/types.go
```diff
@@ -236,6 +236,7 @@ type (
 		Reviewers             Reviewers  `json:"reviewers,omitempty"`
 		Status                string     `json:"status"`
 		Draft                 bool       `json:"draft"`
+		ReviewDecision        string     `json:"review_decision"`
```
sixstone-qq, line 239:
```suggestion
		ReviewDecision        string     `json:"review_decision,omitempty"`
```

cmd/terramate/cli/cloud.go
snakster: What do you mean? The string can be ""? Yes, it can. How would you handle it differently?

cloud/types.go
snakster: I changed it as it can be empty, but when I originally added the field I looked at other fields for reference in the same struct that can also be "" I saw description:
```
Description           string     `json:"description"`
```
Is there a difference between the two? I didn't see any, so I when looking at this I'd wonder why one is omitted if empty and the other not.

sixstone-qq: >  Is there a difference between the two? I didn't see any, so I when looking at this I'd wonder why one is omitted if empty and the other not.

Zero-value for description is valid value. Is zero value a valid value for ReviewDecision?

snakster: It can be "" in case of an error, i.e. rate limit exceeded, but I just tested on a new repository without any review setup and there the API returns null, which will also result in "", so it is a valid value. IMO it's sufficient to treat both cases the same to keep it simple, but if you think it's better to be able to differentiate between the two let me know.

cmd/terramate/cli/cloud.go
i4ki: haha I actually meant the response, sorry.
Here: https://github.com/terramate-io/terramate/pull/1525/files#diff-338c2ed255d3f2d07e303ff13d3302050f7986df605f31cd33fa648c24050078R239

cloud/types.go
snakster: I changed it so it will be "" in case of an error and "none" for a the case where there is no review necessary (i.e. the API null case). This way it works similar to other status fields we have. We could use omitempty, but same goes for many other fields in this struct where we don't, so for consistency reasons I did not add it, but we could revise this in a future PR.

i4ki (approved):
LGTM

Issue #1260, opened 2023-11-24
[BUG] `generate_hcl` doesn't detect dotted files


waxb:
**Describe the bug**
If `generate_hcl`'s label starts with a dot the files are always generated and if the condition is unmet the files are remaining also. Seems like Terramate isn't aware of these files at all hence the generation always.

On the other side `generate_file` behaves exactly as expected.

**To Reproduce**

```
waxb@mbp:terramate-bug$ tree -a
.
├── stack.tm.hcl
├── terramate.tm.hcl
└── test.tm.hcl

1 directory, 3 files
waxb@mbp:terramate-bug$ cat test.tm.hcl 
globals {
  to_generate = true
}

generate_hcl ".file.hcl" {
  condition = global.to_generate
  content {}
}

generate_hcl ".directory/file.hcl" {
  condition = global.to_generate
  content {}
}

generate_file ".file" {
  condition = global.to_generate
  content = ""
}

generate_file ".directory/file" {
  condition = global.to_generate
  content = ""
}
waxb@mbp:terramate-bug$ terramate generate 
Code generation report

Successes:

- /
	[+] .directory/file
	[+] .directory/file.hcl
	[+] .file
	[+] .file.hcl

Hint: '+', '~' and '-' mean the file was created, changed and deleted, respectively.
waxb@mbp:terramate-bug$ tree -a
.
├── .directory
│   ├── file
│   └── file.hcl
├── .file
├── .file.hcl
├── stack.tm.hcl
├── terramate.tm.hcl
└── test.tm.hcl

2 directories, 7 files
waxb@mbp:terramate-bug$ terramate generate 
Code generation report

Successes:

- /
	[+] .directory/file.hcl
	[+] .file.hcl

Hint: '+', '~' and '-' mean the file was created, changed and deleted, respectively.
waxb@mbp:terramate-bug$ terramate generate 
Code generation report

Successes:

- /
	[+] .directory/file.hcl
	[+] .file.hcl

Hint: '+', '~' and '-' mean the file was created, changed and deleted, respectively.
waxb@mbp:terramate-bug$ tree -a
.
├── .directory
│   ├── file
│   └── file.hcl
├── .file
├── .file.hcl
├── stack.tm.hcl
├── terramate.tm.hcl
└── test.tm.hcl

2 directories, 7 files
waxb@mbp:terramate-bug$ sed -i 's/true/false/' test.tm.hcl 
[...]

PR #1538, opened 2024-03-11
refactor!: disallow code generation inside dotdirs.

i4ki:

Disallow code generation inside dot directories.
Note: this is a breaking change.

Examples below will not work anymore:

```hcl
generate_hcl ".terraform/myfile.tf" {
  content {}
}

generate_hcl "somedir/other/.dotdir/myfile.tf" {
  content {}
}

generate_hcl "somedir/other/.dotdir/test/myfile.tf" {
  content {}
}
```

In the cases above, note that part of the file path contains a "dot directory".

Relates to #1260 

I split the PR for fixing the dotfile bug because the disallowing dotdirs is related but not actual part of the bugfix, then by having this separate is easier if a rollback is needed.
Note: this is a breaking change.

- [x] A major version bump is required before this PR can be merged.

```
yes, this is a breaking change.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Changed
+
+- The code generation of HCL and plain files was disallowed inside dot directories.
+
 ## v0.5.5
 
 - Add `script.job.name` and `script.job.description` attributes.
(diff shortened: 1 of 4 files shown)


wmalik (dismissed):

i4ki:
Hi @rgl 

> Why restrict the file names at all?

The reasoning is that Terraform can download dependent modules inside `.terraform/modules` directory and then if that module contains generated files then they would end up detected as orphaned files, making Terramate delete them. So by disallowing generation inside dotdirs we don't descend into them ie we dont detect Terramate files from other projects/repositories.

PR #1533, opened 2024-03-07
chore: add e2e tests for previews

wmalik:
## What this PR does / why we need it:

* Create testserver endpoints for cloud previews (and add tests)
* Create testserver endpoints for github
* Add `TM_GITHUB_API_URL` env var to configure test cli to use testserver as the github api server
* Add basic end-to-end tests for `--cloud-sync-preview`

## Special notes for your reviewer:

* This is an MVP for testing github functionality in the end-to-end tests. The github API response testdata is currently static and suffices the for basc cloud-sync-preview tests, however in future it would be nice to allow tests to customize it based on the test scenario.

## Does this PR introduce a user-facing change?
```
No.

```

PR #1542, opened 2024-03-12
tests: fix race condition by acquiring wrong lock.

i4ki:

Makes use of `mu.Lock()` in the `AppendPreviewLogs()` because `sp.Logs` is being mutated.

As the stack previews need to be updated in the handlers, then it was changed to store a list of pointers so the brittle code below is avoided:
```
for spIndex, sp := range p.StackPreviews {
    for _, sp := range p.StackPreviews { // THIS CREATES A COPY OF SP
        if sp.ID == stackPreviewID {
            sp.Logs = append(sp.Logs, logs...)
            p.StackPreviews[spIndex] = sp // THIS UPDATES WITH THE NEW COPY
	    return nil
        }
    }
}
```
Another example is changes like:
```
org.Previews[i].StackPreviews[spIndex].Status = preview.StackStatus(status)
```
If the value must be mutated by several entities, then better be a pointer protected by mutex than values copied over and also protected by mutex.

Targets #1533 

none

@wmalik if you agree, just merge onto your branch (you can rebase later, dont need to maintain this commit)

```
no
```

### DIFF
--- a/cloud/testserver/cloudstore/store.go
+++ b/cloud/testserver/cloudstore/store.go
@@ -59,7 +59,7 @@ type (
 		TechnologyLayer string                    `json:"technology_layer"`
 		ReviewRequest   *cloud.ReviewRequest      `json:"review_request,omitempty"`
 		Metadata        *cloud.DeploymentMetadata `json:"metadata,omitempty"`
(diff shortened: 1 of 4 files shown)


wmalik (approved):
Looks good, thanks.

PR #1546, opened 2024-03-12
fix: use lower case stack.id for lookup

wmalik:

This would allow users to sync stack logs for stacks having uppercase (or mixed case) stack ids.

```
Yes, this would allow users to sync stacks with uppercase (or mixed case) stack
ids to Terramate Cloud.
```

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -373,7 +373,7 @@ func (c *cli) syncLogs(logger *zerolog.Logger, run runContext, logs cloud.Deploy
 	logger.Debug().RawJSON("logs", data).Msg("synchronizing logs")
 	ctx, cancel := context.WithTimeout(context.Background(), defaultCloudTimeout)
 	defer cancel()
-	stackID := c.cloud.run.meta2id[run.Stack.ID]
+	stackID := c.cloud.run.meta2id[strings.ToLower(run.Stack.ID)]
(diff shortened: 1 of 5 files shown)


i4ki (approved):
LGTM

cmd/terramate/cli/run.go
```diff
@@ -373,7 +373,7 @@ func (c *cli) syncLogs(logger *zerolog.Logger, run runContext, logs cloud.Deploy
 	logger.Debug().RawJSON("logs", data).Msg("synchronizing logs")
 	ctx, cancel := context.WithTimeout(context.Background(), defaultCloudTimeout)
 	defer cancel()
-	stackID := c.cloud.run.meta2id[run.Stack.ID]
+	stackID := c.cloud.run.meta2id[strings.ToLower(run.Stack.ID)]
```
sixstone-qq, line 376: Could we a test for testing this bugfix? Syncing a deployment for a stack whose ID has upper case and their logs?

wmalik: we tested this on staging and it works as expected.

sixstone-qq: In my humble opinion, we should add this in long term to interop tests

wmalik: I agree, in the interest of time I will move forward with this PR now and add interop tests later.

PR #1549, opened 2024-03-12
fix: only get changeset when preview status != canceled

wmalik:

If a stack  preview status is canceled, the status should remain canceled and terraform  changeset should not be fetched.

```
This fixes a user facing issue with stack preview status in Terramate Cloud.
```


i4ki (approved):
I improve the error message in a separate PR

i4ki (approved):
LGTM

PR #1569, opened 2024-03-18
fix(cloud): stack.id with uppercase breaks --cloud-status.

i4ki:

When using `--cloud-status` for stacks synchronized with uppercase `stack.id` they are not listed. It affects `list`, `run` and `trigger`.

none

This was discovered while adding `stack.id` with uppercase.

```
yes, fixes a bug.
```


cmd/terramate/e2etests/cloud/interop/interoperability_test.go
```diff
@@ -17,167 +18,186 @@ import (
...
-		tmcli.Run("run", "--quiet", "--", TerraformTestPath, "init"),
-		RunExpected{
-			Status:       0,
-			IgnoreStdout: true,
-			IgnoreStderr: true,
-		},
-	)
+	for _, stackpath := range []string{
+		"testdata/interop-stacks/basic-drift",
+		// "testdata/interop-stacks/basic-drift-uppercase-id", buggy, not yet
```
i4ki, line 106: another bug, will create another PR for it.

cmd/terramate/e2etests/cloud/run_cloud_deployment_test.go
```diff
@@ -364,32 +371,12 @@ func TestCLIRunWithCloudSyncDeployment(t *testing.T) {
...
-								id = id[:64]
-							}
-							ids = append(ids, id)
-							layout += ":id=" + id
-						}
-						genIdsLayout = append(genIdsLayout, layout)
-					}
-				} else {
-					genIdsLayout = tc.layout
-				}
```
i4ki, line 386: this was created when no special handling for `stack.id` was required, then generating them made it easier to test but too risky to keep now that case insensitivity is handled.

wmalik (approved):

PR #1580, opened 2024-03-19
docs: remove duplicated link

mariux:


i4ki (approved):
LGTM but the other link is at the very end of the sidebar, then maybe less people will find it.

PR #1590, opened 2024-03-21
tests: ensure tests do not use GITHUB_TOKEN exported.

i4ki:

If the developer machine has `GITHUB_TOKEN` exported then the tests are picking it and then testcases try to fetch metadata for dummy repos.

none

```
no
```


wmalik (approved):
Good catch

PR #1596, opened 2024-03-25
chore: improve error messages in --cloud-sync-preview

wmalik:

Print a warning message to stderr when `run` or `script run` is called outside Github Actions.

```
A new warning message is added to the stderr of `run` and `script run`.
```


cmd/terramate/cli/run.go
```diff
@@ -143,6 +145,11 @@ func (c *cli) runOnStacks() {
 		c.detectCloudMetadata()
 	}
 
+	if c.parsedArgs.Run.CloudSyncPreview && os.Getenv("GITHUB_ACTIONS") == "" {
+		printer.Stderr.Warn(cloudSyncPreviewGHAWarning)
+		c.disableCloudFeatures(errors.E(cloudSyncPreviewGHAWarning))
+	}
```
i4ki, line 151: the `createCloudPreview()` is missing a guard for the case cloud is disabled.

cmd/terramate/e2etests/cloud/run_cloud_sync_preview_test.go
```diff
@@ -176,6 +214,8 @@ func TestCLIRunWithCloudSyncPreview(t *testing.T) {
 
 			s := sandbox.New(t)
 
+			err := os.Unsetenv("GITHUB_ACTIONS")
```
i4ki, line 217: This environment variable is already removed from the CLI environ, check the `NewCLI()` implementation. We never remove env vars from the test process, we just filter out and dont pass them down to the Terramate binary.
Also, calling `os.Unsetenv()` unsets it globally, so does not make sense to run on each test case.

i4ki (changes requested):

cmd/terramate/cli/run.go
wmalik: I am not sure if I understand, could you please link to the `createCloudPreview` where the guard is missing? 🤔

i4ki: just saw that it's handled in a `if` in another place:
```
if c.parsedArgs.Run.CloudSyncPreview && c.cloudEnabled() {
 ...
```

this is annoying because we are handling those cases in different ways..

for deployments, its handled inside the `createDeployment()`..
```
func (c *cli) createCloudDeployment(deployRuns []stackCloudRun) {
	// Assume each task in deployRuns has the CloudSyncDeployment set.
	logger := log.With().
		Logger()

	if !c.cloudEnabled() {
		return
	}
```

and for `drift` it's handled in the `cloudSyncAfter()`..

but okay, all good!

cmd/terramate/e2etests/cloud/run_script_cloud_sync_preview_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/terramate-io/terramate/test/sandbox"
 )
 
+// RESUME add e2e test for warning message in SCRIPT/RUN
```
i4ki, line 26: what does that mean?

i4ki (approved):

i4ki (approved):

PR #1603, opened 2024-03-26
feat: add --terragrunt option for run and scripts

snakster:

This PR adds a new option flag `--terragrunt` to `terramate run` and the equivalent `terragrunt = true` script command option. This instructs the CLI to use the `terragrunt show` when generating a plan file instead of `terraform show`. In the future this flag may enable other terragrunt-specific behaviour that should be used when using terramate to invoke terragrunt.

* There are no elaborate tests for this yet, just tests to ensure that the option flag is parsed.

```
Yes, it adds a new flag and script option.
```

### DIFF
--- a/cmd/terramate/cli/cloud_sync_terraform_plan.go
+++ b/cmd/terramate/cli/cloud_sync_terraform_plan.go
@@ -33,10 +33,13 @@ func (c *cli) getTerraformChangeset(run stackCloudRun, planfile string) (*cloud.
 		return nil, errors.E(clitest.ErrCloudInvalidTerraformPlanFilePath, "path must be relative to the running stack")
 	}
 
-	absPlanFilePath := filepath.Join(run.Stack.HostDir(c.cfg()), planfile)
-	_, err := os.Lstat(absPlanFilePath)
-	if err != nil {
-		return nil, errors.E(err, "checking plan file")
+	// Terragrunt writes the plan to a temporary directory, so we cannot check for its existence.
+	if !run.Task.UseTerragrunt {
+		absPlanFilePath := filepath.Join(run.Stack.HostDir(c.cfg()), planfile)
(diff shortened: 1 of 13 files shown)


wmalik (approved):

cmd/terramate/cli/cloud_sync_terraform_plan.go
```diff
@@ -33,10 +33,13 @@ func (c *cli) getTerraformChangeset(run stackCloudRun, planfile string) (*cloud.
...
-	_, err := os.Lstat(absPlanFilePath)
-	if err != nil {
-		return nil, errors.E(err, "checking plan file")
+	// Terragrunt writes the plan to a temporary directory, so we cannot check for its existence.
+	if !run.Task.UseTerragrunt {
+		absPlanFilePath := filepath.Join(run.Stack.HostDir(c.cfg()), planfile)
+		_, err := os.Lstat(absPlanFilePath)
+		if err != nil {
+			return nil, errors.E(err, "checking plan file")
+		}
```
i4k-tm, line 42: what about removing this? this special handling for terraform maybe can create more issues in the future when we introduce more integrations.

i4ki, line 42: What about removing this altogether? I feel like it could create more issues in the future. Wdyt?
I don't have strong opinions, so feel free to address this or not.

i4ki (dismissed):
LGTM

cmd/terramate/cli/cloud_sync_terraform_plan.go
snakster: I'll leave it in for now, because this path returns an error, and the other checks don't, but print a warning instead. This is probably fine, but I'd rather not change existing behaviour without testing it some more.

i4ki (approved):

PR #1604, opened 2024-03-26
feat: add continue-on-error to script run

wmalik:

This PR adds a new flag `--continue-on-error` to the `script run` command.
When this flag is set and a command in a script returns an error:
* the script execution will be aborted and no further
  commands or jobs from that script will be run on the current stack node
* the script execution will continue to run on the next stack node
* the `terramate script run` will return exit code 1 (same behaviour as
  `terramate run --continue-on-error`.

```
Yes, a new flag `--continue-on-error` is added to the `script run` command.
```


cmd/terramate/cli/run.go
```diff
@@ -455,9 +455,11 @@ func (c *cli) runAll(
 			}
 
 			err = errs.AsError()
-			if err != nil && !continueOnError {
+			if err != nil {
 				logger.Info().Msg("interrupting execution of further stacks")
```
snakster, line 459: This will say it cancels further stacks, but it doesn't in case of continueOnError.

snakster (changes requested):
There are still some other occurrences:

https://github.com/terramate-io/terramate/blob/cde8f4ef5b530d6b07fa626a9cec02b4a6973b3d/cmd/terramate/cli/run.go#L345

https://github.com/terramate-io/terramate/blob/cde8f4ef5b530d6b07fa626a9cec02b4a6973b3d/cmd/terramate/cli/run.go#L400

cmd/terramate/cli/run.go
wmalik: Fixed

wmalik:
> There are still some other occurrences:
> 
> https://github.com/terramate-io/terramate/blob/cde8f4ef5b530d6b07fa626a9cec02b4a6973b3d/cmd/terramate/cli/run.go#L345
> 
> https://github.com/terramate-io/terramate/blob/cde8f4ef5b530d6b07fa626a9cec02b4a6973b3d/cmd/terramate/cli/run.go#L400
> 
> I think it should break there instead of continue.

@snakster Addressed this as well and improved the tests.

snakster (dismissed):
Looks good! 👍

snakster (approved):

PR #1606, opened 2024-03-26
chore: undo re-enable signal tests for MacOS

snakster:

Undo #1490 because it results in spurious test failures. Indicates that we have problem with Signal handling on MacOS, must be investigated further and fixed before we can re-enable. 

```
no
```


i4ki (approved):
LGTM
Thanks!

PR #1612, opened 2024-04-02
fix(tmc): triggered_by metadata is not the real actor.

i4ki:

The `github_actions_deployment_triggered_by` was using the `GiTHUB_ACTOR` environment variable, which returns the actor that submitted the change but not the one that actually triggered the deployment.

A separate metadata was created for synchronizing the original actor and the triggered_by now uses the `GITHUB_TRIGGERING_ACTOR` environment variable.

none

In the beginning it was planned to fetch this information from the GH API but the specific endpoint where this information is available is not fetched yet, so using this environment variable is less error-prone. 

```
no
```


cmd/terramate/cli/cloud.go
```diff
@@ -798,7 +798,8 @@ func setDefaultGitMetadata(md *cloud.DeploymentMetadata, commit *git.CommitMetad
 }
 
 func setGithubActionsMetadata(md *cloud.DeploymentMetadata) {
-	md.GithubActionsDeploymentTriggeredBy = os.Getenv("GITHUB_ACTOR")
+	md.GithubActionsDeploymentActor = os.Getenv("GITHUB_ACTOR")
+	md.GithubActionsDeploymentTriggeredBy = os.Getenv("GITHUB_TRIGGERING_ACTOR")
```
sixstone-qq, line 802: I don't see this documented [here](https://docs.github.com/en/actions/learn-github-actions/variables). Do you have any source for this or is it pure experimental?

sixstone-qq: Oh [found it](https://docs.github.com/en/actions/learn-github-actions/contexts) `github.triggering_actor`

sixstone-qq (approved):
:rock:

wmalik (approved):

PR #1626, opened 2024-04-10
refactor: remove cloud flag prefix

snakster:

This PR renames all `--cloud-*` flags to `--*` to make the options more accessible. The old flags are still supported as aliases, but not shown in help or error messages.

Same goes for script command options, where `cloud_*` command option prefix is dropped.

```
Yes. Flags are renamed and promoted, but the old ones are still supported.
```


sixstone-qq (approved):
LGTM

i4ki (approved):
LGTM

PR #1635, opened 2024-04-15
fix: naming of cloud script options

snakster:

In the previous release, the flags `--cloud-sync-terraform-plan-file` and `--cloud-sync-layer` were shorted to `--terraform-plan-file` and `--layer`, but the script option for both still had the `sync_` prefix. This PR shortens them further to `terraform_plan_file` and `layer`, so everything is aligned. As before, the old names are still supported.

```
Two script options are renamed, old names (including the `sync_*` ones) are still supported.
```

### DIFF
--- a/config/script.go
+++ b/config/script.go
@@ -370,6 +370,8 @@ func unmarshalScriptCommandOptions(obj cty.Value, expr hhcl.Expression) (*Script
 			}
 			r.CloudSyncPreview = v.True()
 
+		case "layer":
+			fallthrough
 		case "sync_layer":
 			fallthrough
 		case "cloud_sync_layer":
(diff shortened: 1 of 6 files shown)


config/script.go
```diff
@@ -370,6 +370,8 @@ func unmarshalScriptCommandOptions(obj cty.Value, expr hhcl.Expression) (*Script
 			}
 			r.CloudSyncPreview = v.True()
 
+		case "layer":
+			fallthrough
 		case "sync_layer":
 			fallthrough
```
i4ki, line 376: we just broke the world record for highest number of fallthrough in the same Go file 😂

i4ki (approved):
LGTM

PR #1640, opened 2024-04-17
fix: generate inherit=false not working when imported.

i4ki:

Fixes a problem that generate blocks with `inherit=false` are not generated if the blocks are imported and inherited in child stacks.

Thanks @ckd

Reported on discord: https://discord.com/channels/1088753599951151154/1229901710319685692/1230006750325768192

```
yes, fixes a bug.
```


soerenmartius (approved):
lgtm

PR #1675, opened 2024-04-26
fix: incorrect warnings emitted by the parser.

i4ki:

Fix incorrect warnings emitted by the parser when `terramate.config.run.env` is declared in child directories.

none

```

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Fixed
+
+- Fix incorrect warning emitted by the parser in the case `terramate.config.run.env` is defined outside project root.
+
 ## v0.8.0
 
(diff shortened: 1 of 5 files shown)


snakster:
I did some testing and the warnings are gone in the scenarios I tried, but when adding a disallowed block at the non-root level, the warning says
```
2024-04-29T13:01:37+02:00 WRN error="XXXXXXXX/stack.tm.hcl:7,1-10: `terramate` block is only allowed at the project root directory" action=TerramateParser.checkConfigSanity()
Error: unable to parse configuration
> XXXXXXXXXXXX/s1/s2/stack.tm.hcl:8,3-6: terramate schema error: loading from /Users/snakster/Workspace/warning-fix/s1: loading from XXXXXXXXXX/s1/s2: unrecognized block "git"
```
Maybe the message should be worded differently now?

snakster (approved):
👍

PR #1685, opened 2024-05-02
fix: only collects review_decision if token is present.

i4ki:

Improve the metadata collection.

- only collects `review_decision` if token is present.
- fix preview tests dependent on `TM_GITHUB_API_URL` being set.

none

```
no
```


sixstone-qq (approved):
LGTM

snakster (approved):

PR #1701, opened 2024-05-07
fix: deployments with empty stack list are no longer sent to cloud

snakster:

This PR fixes empty deployments being created with `terramate run` or `script run` are invoked without any selected stacks.

```
no
```


i4ki:
can we close this? AFAIK this works as intended.

PR #1708, opened 2024-05-08
fix: allow map type in tm_dynamic.attributes.

i4ki:

Allow map type in the `tm_dynamic.attributes`.

none

```
yes, fixes a usage limitation.
```


sixstone-qq (approved):
Lgtm

PR #1711, opened 2024-05-10
fix: run.env not including complete env var.

### LINKED ISSUE #1710: [BUG] runtime removes all characters from env value after `=`
**Describe the bug**
If the env value contains `=`, then all characters after `=` will be removed in `terramate.config.run.env`

**To Reproduce**

1. Source data
```
env | grep KEY
KEY1=adasd=
KEY2=adasd=adasd
KEY3=adasd==
KEY4=a=adasd

```
```
terramate {
  config {
    run {
      env {
        KEY1 = env.KEY1
        KEY2 = env.KEY2
        KEY3 = env.KEY3
        KEY4 = env.KEY4
      }
    }
  }
}
```

2. Run command `tm debug show runtime-env`
3. See modified values
```
stack "/terraform/test":
	KEY1=adasd
	KEY2=adasd
	KEY3=adasd
	KEY4=a
	TF_PLUGIN_CACHE_DIR=~/.terraform.d/plugin-cache
```

**Expected behavior**
Do not change env values and handle the `=` character in env values correctly.

```
# tm debug show runtime-env
stack "/terraform/test":
	KEY1=adasd=
	KEY2=adasd=
	KEY3=adasd=adasd
	KEY4=a=adasd
	TF_PLUGIN_CACHE_DIR=~/.terraform.d/plugin-cache
```

**Log Output**
-

 - OS: macOS
 - OS Version [14.4.1]
 - Git Version [2.39.3] (git --version)
 - Terramate Version [0.8.3] (terramate --version)

i4ki:

Fixes a bug in the processing of Go's environ, which has the form `NAME=VALUE` but the value can also contain equal signs.

Fixes #1710 

```
yes, fixes a bug.
```

### DIFF
--- a/hcl/eval/eval.go
+++ b/hcl/eval/eval.go
@@ -57,8 +57,8 @@ func (c *Context) SetFunction(name string, fn function.Function) {
 func (c *Context) SetEnv(environ []string) {
 	env := map[string]cty.Value{}
 	for _, v := range environ {
-		parsed := strings.Split(v, "=")
-		env[parsed[0]] = cty.StringVal(parsed[1])
+		equalAt := strings.Index(v, "=") // must always find
+		env[v[:equalAt]] = cty.StringVal(v[equalAt+1:])
 	}
 	c.SetNamespace("env", env)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Fixed
+
+- Fix the loading of `terramate.config.run.env` environment variables not considering equal signs in the value.
+
 ## v0.8.3
 
 ### Fixed
--- a/run/env_test.go
+++ b/run/env_test.go
@@ -85,6 +85,33 @@ func TestLoadRunEnv(t *testing.T) {
 			},
 		},
 		{
+			// GH issue: https://github.com/terramate-io/terramate/issues/1710
+			name: "regression test: incomplete env due to equal sign split",
+			hostenv: map[string]string{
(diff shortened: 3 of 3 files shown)


hcl/eval/eval.go
```diff
@@ -57,8 +57,8 @@ func (c *Context) SetFunction(name string, fn function.Function) {
 func (c *Context) SetEnv(environ []string) {
 	env := map[string]cty.Value{}
 	for _, v := range environ {
-		parsed := strings.Split(v, "=")
```
sixstone-qq, line 60: Why don't we use `strings.SplitN(v, "=", 1)` instead? [Link](https://pkg.go.dev/strings#SplitN) I think it would be more intuitive

soerenmartius (approved):
lgtm

hcl/eval/eval.go
i4ki: yes `SplitN()` does the job but I think I will keep just the `Index()` because we are going make Terramate more GC friendly.
The `SplitN()` does:
```
if n == 0 {
	return nil
}
if sep == "" {
	return explode(s, n)
}
if n < 0 {
	n = Count(s, sep) + 1
}

if n > len(s)+1 {
	n = len(s) + 1
}
a := make([]string, n)
n--
i := 0
for i < n {
	m := Index(s, sep)
	if m < 0 {
		break
	}
	a[i] = s[:m+sepSave]
		s = s[m+len(sep):]
		i++
	}
a[i] = s
return a[:i+1]
```

all those `ifs` are just branch predictors nightmare because `strings.Split()` is a very common function used in many different ways, then the processor is unable to know which branch is more plausible. We also don't need to allocate a new array+slice header.
In our case, just the last bits of this function makes sense, the thing I implemented.

sixstone-qq (approved):

PR #1734, opened 2024-05-30
feat: add triggers of type ignore-change.

i4ki:

Introduce the `ignore-change` trigger. This kind of trigger can be used when a set of changes would trigger stacks when should not, then the user can now force them to not trigger as changed in the same PR.

Usage example:
```
$ terramate list --changed
my-stack
$ terramate experimental trigger --ignore-change my-stack
Created trigger for stack "/my-stack"

$ terramate list --changed
$ 
```

As you can see, after the trigger file is created with `--ignore-change` the stack is not listed as changed anymore. The same is true for the other features supporting `--changed`.

Fixes none

```
yes, adds a new feature.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
     - `terramate list --status`
     - `terramate cloud drift show`
 - Add `script.lets` block for declaring variables that are local to the script.
+- Add `--ignore-change` flag to `terramate experimental trigger`, which makes the change detection ignore the given stacks.
+  - It inverts the default trigger behavior.
 
 ### Fixed
 
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -245,6 +245,8 @@ type cliSpec struct {
 
 		Trigger struct {
 			Stack              string `arg:"" optional:"true" name:"stack" predictor:"file" help:"The stacks path."`
+			Change             bool   `default:"false" help:"Trigger stacks as changed"`
+			IgnoreChange       bool   `default:"false" help:"Trigger stacks to be ignored by change detection"`
 			Reason             string `default:"" name:"reason" help:"Set a reason for triggering the stack."`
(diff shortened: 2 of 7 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -965,15 +967,27 @@ func (c *cli) triggerStackByFilter() {
 }
 
 func (c *cli) triggerStack(stack string) {
+	changeFlag := c.parsedArgs.Experimental.Trigger.Change
+	ignoreFlag := c.parsedArgs.Experimental.Trigger.IgnoreChange
+
+	if changeFlag && ignoreFlag {
+		fatal("flag --change and --ignore-change conflicts")
```
snakster, line 974:
```suggestion
		fatal("flags --change and --ignore-change are conflicting")
```

CHANGELOG.md
```diff
@@ -34,6 +34,9 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
     - `terramate list --status`
     - `terramate cloud drift show`
 - Add `script.lets` block for declaring variables that are local to the script.
+- Add `--ignore-change` flag to `terramate experimental trigger`.
```
snakster, line 37:
```suggestion
- Add `--ignore-change` flag to `terramate experimental trigger`, which makes the change detection ignore the given stacks.
  - This inverts the default trigger behaviour.
```

stack/trigger/trigger.go
```diff
@@ -71,6 +78,23 @@ func StackPath(triggerFile project.Path) (project.Path, bool) {
 	return project.NewPath(stackPath), true
 }
 
+// Is checks if the filename is a trigger file and if that's the case then it returns
+// the parsed trigger info and the triggered path.
+// If the file is not inside the triggerDir then it returns an empty Info, empty path, false and no error.
+// If there's an error parsing the trigger file, it returns empty info, false and the error.
```
snakster, line 84: This doesn't match the code, it returns true.

snakster (commented):
I did only a partial review today, but I found some small things.

snakster (approved):
LGTM 👍

PR #1745, opened 2024-06-05
fix: trigger type not used in the filename.

i4ki:

When using `trigger --ignore-change` the filename has the pattern `changed-.*` instead of `ignore-change-.*`.

None

```
yes but no changelog entry is needed because this is unreleased.
```


soerenmartius (approved):
lgtm

PR #1752, opened 2024-06-07
chore: improve the trigger message.

i4ki:

Improves the trigger message feedback showing which trigger kind was created.

none
```
yes, but cosmetic and only makes sense with the unreleased ignore-change trigger. Before that, all triggers were of kind "change".
```


cmd/terramate/cli/cli.go
```diff
@@ -978,14 +978,19 @@ func (c *cli) triggerStack(basePath string) {
 		fatal("flags --change and --ignore-change are conflicting")
 	}
 
-	var kind trigger.Kind
+	var (
+		kind     trigger.Kind
+		kindName string
```
sixstone-qq, line 983: Why not create a `String()` or similar method to `trigger.Kind` instead?

i4ki: Thanks for the suggestion but the `kind` name used in the trigger file are the strings below:
```
// Supported trigger kinds.
const (
	Changed Kind = "changed"
	Ignored Kind = "ignore-change"
)
```

IMO, a better name for "changed" trigger would be just "change" and then yes, a simple `.String()` would suffice. The trigger `changed` type was introduced in `v0.2.8` and changing this is a breaking change.

Having said that, a `trigger.String()` cannot be used but maybe another method name but I'm unable to come up with something good :/
Also I don't know if this would be useful in other places, as it only deal with a user message in the screen.

snakster (approved):
👍

PR #1753, opened 2024-06-11
refactor: rename config.targets to config.cloud.targets

snakster:

This PR moves the `targets` config block as a sub-block of `cloud`, because targets are a cloud option.

- Targets is only released as RC, so I just changed the existing changelog entry.

```
no, the feature is not released yet
```


i4ki (approved):
LGTM

PR #1760, opened 2024-06-14
fix: when targets experiment is disabled, the default must be used.

i4ki:

Fix targets in case user synced targets but later decide to disable the feature.

none

```
no, the feature is unreleased
```


snakster (approved):
👍

PR #1769, opened 2024-06-21
feat: add gpg signing for all packages

### LINKED ISSUE #1759: [FEATURE] GPG Sign RPM and DEB packages
Our security folks are concerned about package integrity when installing terramate in our CI system.  Their concerns is that without strong validation of packages, we leave ourselves open to supply chain attacks. 

We'd love it if the packages created were GPG signed so we can install without having an approval in a number of places

* The terramate github action suffers a similar problem, it uses the github release but doesn't check the signatures file or signing
* We're probably going to use jaxxstorm/action-install-gh-release and submit a patch to check a checksum, but this requires a security review for the action

**Additional context**
GoReleaser will do [this](https://goreleaser.com/customization/nfpm/#signing-key-passphrases) if you add a `sign:` section to the `nfpm:` build and provide it a private key and passphrase, so this should allow you to  work with DEBs and switch to `repo_gpgcheck=false` in the yum/dnf config

RocketRene:
## Reason for Changes

closes #1759   

The primary reason for these changes is to enhance the security of our release process. By signing our artifacts with a GPG key, we provide a layer of verification that ensures the artifacts have not been tampered with and are indeed from our organization.

## Impact of Changes

These changes will make our releases more secure and trustworthy. Users can verify the integrity and authenticity of our artifacts using the provided signatures. This is particularly important for open-source projects where trust is paramount.

This PR represents a significant step forward in securing our software delivery process, and I look forward to any feedback or suggestions for further improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Fixed
+
+- Fix crash when supplying a tag list with a trailing comma separator.
+
 ## v0.9.0
 
 ### Added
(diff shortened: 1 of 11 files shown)


.goreleaser.yaml
```diff
@@ -63,6 +63,19 @@ dockers:
...
+    args:
+        [
+          "-u",
+          "github@terramate.io",
+          "--output",
+          "${signature}",
+          "--detach-sign",
+          "${artifact}",
+        ]
+
```
soerenmartius, line 77:
```suggestion
```

i4ki (approved):
LGTM

PR #1772, opened 2024-06-25
chore: add gpg action to dry run

RocketRene:
This PR adds the GPG Action step to the dry run workflow. 

This is required ,otherwise CI tests fail


i4ki (approved):
LGTM!

Thanks!

PR #1816, opened 2024-08-02
fix(generate): deletion of orphaned sharing files.

i4ki:

Fixes the deletion of orphaned files configured by input/output blocks.
Depends on #1812 

none

Depends #1812 

```
no, the feature is not released.
```


soerenmartius (dismissed):
lgtm

PR #1822, opened 2024-08-08
feat: extend sharing support to scripts.

i4ki:

Add `job.command[n].enable_sharing` and `job.command[n].mock_sharing_on_fail`.

none
```
yes
```


config/script.go
```diff
@@ -43,6 +43,8 @@ type ScriptCmdOptions struct {
 	CloudTerraformPlanFile string
 	CloudTofuPlanFile      string
 	UseTerragrunt          bool
+	EnableSharing          bool
+	MockOnFail             bool
```
sixstone-qq, line 47: *nitpick* Should we name it like the flag name because it may be misleading mixed with other concepts? This applies to script configuration too
```suggestion
	MockSharingOnFail             bool
```

e2etests/core/run_sharing_test.go
```diff
@@ -292,29 +267,69 @@ func TestRunSharing(t *testing.T) {
...
-		t.Run(tc.name, func(t *testing.T) {
-			s := sandbox.New(t)
-			s.BuildTree(tc.layout)
-			tmcli := NewCLI(t, s.RootDir())
-			tmcli.PrependToPath(filepath.Dir(TerraformTestPath))
-			res := tmcli.Run("run", HelperPath, "echo", "hello")
-			if res.Status == 0 {
-				// generate safeguard must trigger
-				t.Fatal("run must fail if sharing is not generated")
+		runCases := []bool{false, true}
```
sixstone-qq, line 270: *nitpick* This one comes from my ah moment when I saw the var name, no need to apply but I feel it more readable.
```suggestion
		scriptCaseRuns := []bool{false, true}
```

sixstone-qq (approved):
Just two nitpicky comments, you can discard them. Good job

config/script.go
i4ki: It makes sense! 
I'm waiting for @mariux feedback if the `mock_sharing_on_fail` name is okay, so once this is decided I will change.

e2etests/core/run_sharing_test.go
i4ki: I will merge the PR stack and address this and the other request in a separate PR.

PR #1824, opened 2024-08-09
fix: handle --continue-on-error when sharing outputs fail.

i4ki:

Fix the handling of the `--continue-on-error` flag when used with sharing outputs feature.

none

```
yes but the feature is unreleased.
```


sixstone-qq (dismissed):
LGTM

PR #1830, opened 2024-08-10
fix: release locking resource when continuing through run loop

### LINKED ISSUE #1828: [BUG] Terramate run becomes unresponsive when encountering errors
**Describe the bug**
`terramate run` will often become unresponsive when running parallel commands that encounter errors. This was observed both locally and in CICD.

The repro example contains a timeout however in practice terramate will hang indefinitely and not respond to interrupt signals.

**To Reproduce**
```bash
brandongr brandongr$ git init tm-bug
Initialized empty Git repository in /Users/brandongr/gitlab/brandongr/tm-bug/.git/
brandongr brandongr$ cd tm-bug/
tm-bug brandongr$ for i in $(seq 1 5); do terramate create stacks/$i; done
Created stack /stacks/1
Created stack /stacks/2
Created stack /stacks/3
Created stack /stacks/4
Created stack /stacks/5
tm-bug brandongr$ timeout 30 terramate run --parallel 2 -- cat foo.tf
terramate: Entering stack in /stacks/2
terramate: Entering stack in /stacks/5
terramate: Executing command "cat foo.tf"
terramate: Executing command "cat foo.tf"
cat: foo.tf: No such file or directory
terramate: Entering stack in /stacks/1
terramate: Executing command "cat foo.tf"
cat: foo.tf: No such file or directory
terramate: Entering stack in /stacks/4
cat: foo.tf: No such file or directory
terramate: Entering stack in /stacks/3
Error: one or more commands failed
> execution failed: running /bin/cat foo.tf (in /stacks/5): exit status 1
> execution failed: running /bin/cat foo.tf (in /stacks/2): exit status 1
> execution failed: running /bin/cat foo.tf (in /stacks/1): exit status 1
tm-bug brandongr$ for i in $(seq 6 10); do terramate create stacks/$i; done
Created stack /stacks/6
Created stack /stacks/7
Created stack /stacks/8
Created stack /stacks/9
Created stack /stacks/10
tm-bug brandongr$ timeout 30 terramate run --parallel 2 -- cat foo.tf
terramate: Entering stack in /stacks/2
terramate: Entering stack in /stacks/6
terramate: Executing command "cat foo.tf"
terramate: Executing command "cat foo.tf"
cat: foo.tf: No such file or directory
terramate: Entering stack in /stacks/7
cat: foo.tf: No such file or directory
terramate: Entering stack in /stacks/9
tm-bug brandongr$ echo $?
124
```

1. Init a git respository
2. Create stacks
3. Execute `terramate run --parallel n <some failing command>`, where num. stacks > n > 1

**Expected behavior**
Terramate execution predictably terminates with errors

**Log Output**
[tm-bug.log](https://github.com/user-attachments/files/16564090/tm-bug.log)

 - OS: Darwin
 - OS Version: 14.6
 - Git Version: 2.39.3 (Apple Git-146)
 - Terramate Version: 0.10.0

CICD environment is Docker 
Linux/Debian 12

**Additional context**
N/A

bgrams:
Parallel tasks executed by `terramate run` do not release their lock when continuing through the run loop, preventing other tasks from running or exiting when `parallelism < num. tasks`. This occurs during both context cancellation and dry runs. PR ensures that the lock is released under these conditions so that the program is able to terminate.

Tested locally.

Fixes #1828 

TIA 🙂 

No

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -530,6 +530,7 @@ func (c *cli) runAll(
 			select {
 			case <-cancelCtx.Done():
 				c.cloudSyncAfter(cloudRun, runResult{ExitCode: -1}, errors.E(ErrRunCanceled))
+				releaseResource()
 				continue
 			default:
 			}
@@ -579,6 +580,7 @@ func (c *cli) runAll(
 			}
 
 			if opts.DryRun {
+				releaseResource()
 				continue
 			}
 
--- a/e2etests/core/run_parallel_test.go
+++ b/e2etests/core/run_parallel_test.go
@@ -6,6 +6,7 @@ package core_test
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"testing"
 
@@ -50,6 +51,34 @@ func TestParallelFibonacci(t *testing.T) {
 			assert.EqualInts(t, tc.WantValue, int(got))
 		})
 	}
+}
+
+func TestParallelBug1828Regression(t *testing.T) {
+	// see: https://github.com/terramate-io/terramate/issues/1828
+
+	s := sandbox.NoGit(t, true)
+	layout := []string{}
+	for i := 0; i < 10; i++ {
+		layout = append(layout, fmt.Sprintf(`s:stack-%d`, i))
+	}
+	s.BuildTree(layout)
+	tmcli := NewCLI(t, s.RootDir())
+	AssertRunResult(t,
+		tmcli.Run("run", "--parallel=2", "--quiet", "--", "cat", "foo.txt"),
+		RunExpected{
+			Status:      1,
+			StderrRegex: "No such file or directory",
+		},
+	)
+
+	// dry-run check
+	AssertRunResult(t,
+		tmcli.Run("run", "--parallel=2", "--dry-run", "--", "cat", "foo.txt"),
+		RunExpected{
+			Status:      0,
+			StderrRegex: regexp.QuoteMeta("terramate: (dry-run)"),
+		},
+	)
 }
 
 func makeFibLayout(n int) []string {


bgrams:
Hi @i4ki, thanks for the pointers! I was looking for tests initially but overlooked this folder. The requested changes have been made.

i4ki (approved):
LGTM

Thank you @bgrams

i4ki:
Merged at https://github.com/terramate-io/terramate/commit/6471a82d87b94db56dd70bf8689a1e8e25480e56

Thank you!

There are more cases to be handled and they were introduced by [this PR](https://github.com/terramate-io/terramate/pull/1727).
We will address the other cases in a separate PR.

bgrams:
Thanks @i4ki!

Just FYI I chose not to handle those cases given that the semaphore is locked/unlocked from only within the [task run loop](https://github.com/terramate-io/terramate/blob/4bc688f88451e10f8d79173b6d1dbe1dc69b7d69/cmd/terramate/cli/run.go#L388) in the function passed to `sched.Run`. `return`s from this function and `break`s from this loop do not require additional locks to be acquired, so I understand the existing code to be functionally correct with these changes despite being somewhat unintuitive.

IMO this may benefit from a refactor to make locking an implementation detail of the `Parallel` scheduler. This could guarantee correctness while also simplifying the code a considerable amount. I avoided bringing this in scope for the sake of keeping the diff small and getting the change out but I have a basic working example that I can share if you'd like.

i4ki:
> Just FYI I chose not to handle those cases given that the semaphore is locked/unlocked from only within the [task run loop](https://github.com/terramate-io/terramate/blob/4bc688f88451e10f8d79173b6d1dbe1dc69b7d69/cmd/terramate/cli/run.go#L388) in the function passed to sched.Run. returns from this function and breaks from this loop do not require additional locks to be acquired, so I understand the existing code to be functionally correct with these changes despite being somewhat unintuitive.

Initially I thought more cases were needed but actually it's just one. Have a look [here](https://github.com/terramate-io/terramate/blob/4bc688f88451e10f8d79173b6d1dbe1dc69b7d69/cmd/terramate/cli/run.go#L543-L551).

I added an ugly fix here: https://github.com/terramate-io/terramate/pull/1832
[This subtle change](https://github.com/terramate-io/terramate/pull/1727) in this code path, of just moving lines around, drastically changed the behavior and introduced the bugs (it also break the `--continue-on-error` in some cases). They will be fixed now in a quick way but refactored later.

> IMO this may benefit from a refactor to make locking an implementation detail of the Parallel scheduler. This could guarantee correctness while also simplifying the code a considerable amount. I avoided bringing this in scope for the sake of keeping the diff small and getting the change out but I have a basic working example that I can share if you'd like.

Please share if you have ideas and we discuss. You can also reach out to me in our discord server or private discord DM. My username there is `_i4k`.

bgrams:
> Initially I thought more cases were needed but actually it's just one. Have a look [here](https://github.com/terramate-io/terramate/blob/4bc688f88451e10f8d79173b6d1dbe1dc69b7d69/cmd/terramate/cli/run.go#L543-L551).

Ah, that makes sense. Nice catch!

> Please share if you have ideas and we discuss. You can also reach out to me in our discord server or private discord DM. My username there is _i4k.

Sounds good, I'll ping you 👍

snakster:
> IMO this may benefit from a refactor to make locking an implementation detail of the Parallel scheduler.

To add some context, I initially implemented it like this, but the reason I changed it is the way scripts are handled. The scheduler operates on a DAG, and the nodes of the DAG are stacks, not individual commands. Changing this would require a bit of work.

For scripts, however, there are multiple commands per stack, which are executed sequentially in an inner loop of the `sched.Run` callback (this loop: `for _, task := range run.Tasks {...`). For parallel scheduling this is fine, since all these commands are executed sequentially, it might as well be a single command, but it also means currently the scheduler would only be able do a lock/release for all script commands of a stack. I thought it would be nicer to do the acquire/release per command, so that's why it's not done in the scheduler anymore.

bgrams:
Hi @snakster, that makes sense! I imagine we both arrived at a similar implementation. The code is here if you are curious - https://github.com/terramate-io/terramate/commit/6b6b63f208447e3c724f9f09e7762d395c0908a3.

I also came to this realization around the change in execution behavior after the fact. I don't know that we would have a use for command-level locking given that a "script" may more often be thought of as a single operation, but I recognize that the community may find this to be a useful feature.

Coincidentally, we started playing with the scripts feature today and it's been very handy! Appreciate all your team's great work on this project.

PR #1831, opened 2024-08-12
fix: release locking resource when continuing through run loop

i4ki:

At the moment, we can't run pipelines from fork contributions, then we need to cherry-pick the commits in a separate PR, to run CI and merge the contribution. The original authorship is kept.

none

This apply the contribution from #1830 and add a CHANGELOG entry.

```
yes, fixes a deadlock in the orchestration.
```


snakster:
What about https://github.com/terramate-io/terramate/blob/8a141f6da962aa6035a168afd290aecaa574ad35/cmd/terramate/cli/run.go#L417 etc.?

Looks to me like there are still a lot of places where the resource is not released.

snakster (approved):
LGTM. I think there are still some more releases missing here, but we can do that later.

i4ki:
yes, let's address them separately.

PR #1832, opened 2024-08-13
fix(parallel): deadlock in case command does not exist.

i4ki:

Note: this is just a quick fix and a refactor in this code path is being worked on.

This fixes a missing `releaseResource()` that led to a deadlock in the case that user-supplied command does not exist. The fix also release the resource in all other code paths that abort the execution, for correctness purposes, but AFAICS they are optional.

Relates to #1828

A refactor in the outputs-sharing handling is being designed.

```
yes, fixes a --parallel bug.
```

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -411,8 +411,10 @@ func (c *cli) runAll(
 					if err != nil {
 						errs.Append(errors.E(err, "failed to evaluate input block"))
 						if continueOnError {
+							// TODO(i4k): must continue in next stack.
 							break
 						}
+						releaseResource()
 						cancel()
 						return errs.AsError()
 					}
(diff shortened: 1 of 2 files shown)


cmd/terramate/cli/run.go
```diff
@@ -544,6 +561,7 @@ func (c *cli) runAll(
 			if err != nil {
 				c.cloudSyncAfter(cloudRun, runResult{ExitCode: -1}, errors.E(ErrRunCommandNotFound, err))
 				errs.Append(errors.E(err, "running `%s` in stack %s", cmdStr, run.Stack.Dir))
+				releaseResource()
```
i4ki, line 564: This line fixes the deadlock reproduced in the `TestParalleCmdNotFoundContinueOnError` testcase.

PR #1843, opened 2024-08-20
fix(tmc): outputs sharing failure cases not synced to TMC.

i4ki:

It fixes the `outputs-sharing` experiment not synchronizing failure cases to Terramate Cloud.

none
```
yes, fixes a bug in v0.10.1
```

### DIFF
--- a/cmd/terramate/cli/run.go
+++ b/cmd/terramate/cli/run.go
@@ -49,8 +49,9 @@ const (
 
 // stackRun contains a list of tasks to be run per stack.
 type stackRun struct {
-	Stack *config.Stack
-	Tasks []stackRunTask
+	Stack         *config.Stack
+	Tasks         []stackRunTask
+	SyncTaskIndex int // index of the task with sync options
 }
 
 // stackCloudRun is a stackRun, but with a single task, because the cloud API only supports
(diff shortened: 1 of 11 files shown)


snakster (commented):
I think there are still problems in this function. Some of them were already there before, so this was my mistake, but it has been copied.

I think it's never ok to return from the inner function:
```
c.cloudSyncAfter(...)
...
return errs.AsError()
```
The problem is that `c.cloudSyncAfter` will only sync, if the task is the cloud sync command. For scripts, this means if an earlier command fails which is not the cloud sync command (i.e. terraform init), even though it goes through the `cloudSyncAfter(...)`, it will not sync.

Cancel does it correctly:
```
select {
case <-cancelCtx.Done():
	c.cloudSyncAfter(cloudRun, runResult{ExitCode: -1}, errors.E(ErrRunCanceled))
	continue
default:
}
```

soerenmartius (approved):
LGTM

cmd/terramate/cli/run.go
```diff
@@ -385,10 +386,23 @@ func (c *cli) runAll(
...
+			// For cloud sync, we always assume that there's a single task per stack.
+			cloudRun := stackCloudRun{Stack: run.Stack, Task: task}
+
+			select {
+			case <-cancelCtx.Done():
+				c.cloudSyncAfter(cloudRun, runResult{ExitCode: -1}, errors.E(ErrRunCanceled))
+				releaseResource()
+				continue tasksLoop
+			default:
+			}
```
i4ki, line 404: This is moved up so we don't try any outputs-sharing handling if the stack execution is already canceled.

snakster (approved):
Looks good now 👍

PR #1867, opened 2024-09-12
feat: make inputs of type 'any' by default.

### LINKED ISSUE #1861: [FEATURE] Support specifying data types for "Input" block for output sharing
We followed the instructions from the documentation and created sharing_backend, output and input. After generting the code we see that input blocks creates a variable which has type "string". Our output parameter type however is a map. We did not find any way to override or remove this "string" type and hence we could not use the output sharing feature.  

Ability to specify data type in the Input block or just remove the type? 

No alternatives that we could find

i4ki:

Change default sharing backend type to be `any` instead of `string`.

Closes #1861 

```
yes, add support for any input type.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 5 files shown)


snakster (approved):
👍

PR #1871, opened 2024-09-13
chore: use script to run tests and sync results to cloud

i4ki:

Experiment: Uses Terramate to run its own tests and sync the results to Terramate Cloud.

none

```
no
```


e2etests/cloud/run_script_cloud_drift_test.go
```diff
@@ -420,6 +420,7 @@ func TestScriptRunDriftStatus(t *testing.T) {
 			},
 			env: []string{
 				`TF_VAR_content=my secret`,
+				"TMC_API_DEBUG=1",
```
i4ki, line 423: please remove this

hack/packages.sh
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Terramate GmbH
+# SPDX-License-Identifier: MPL-2.0
+
+packages() {
+    go list -f '{{.Dir}}' ./... | sort
+}
+
+packages_with_tests() {
+    go list -f '{{.Dir}}' ./... | sort
```
snakster, line 9: Is it intended that both are the same?

snakster (approved):

hack/packages.sh
i4ki: no, will fix in a separate PR.

PR #1880, opened 2024-09-24
tests: fix regression tests.

i4ki:

Fix the regression tests that were broken when we switched to use `targets` in the `terramate-tests` organization.

none

```
no
```


e2etests/cloud/interop/main_test.go
```diff
@@ -27,10 +26,6 @@ func TestMain(m *testing.M) {
 		fmt.Printf("TMC_API_HOST conflicts with TMC_API_URL")
 		os.Exit(1)
 	}
-	if apiHost == cloud.Host || apiURL == cloud.BaseURL {
```
i4ki, line 30: I removed this because in a separate PR I will create another workflow to run this against our `terramate-tests` prod org.

snakster: The workflow uses `TMC_API_HOST: api.stg.terramate.io` though and not prod, is this intended?

i4ki: yes, @sixstone-qq asked for running on staging.

snakster (approved):

PR #1883, opened 2024-09-24
fix: outdated safeguard descending into stack child dirs.

i4ki:

The outdated safeguard handling logic is descending into stack subdirs even when its parent stack was already handled.

Issue reported in Discord.

```
yes, fixes a bug.
```


soerenmartius (approved):
lgtm

PR #1886, opened 2024-09-25
fix: cleanup of orphaned files must respect -C flag.

i4ki:

When using `terramate -C <dir> generate`, Terramate is still cleaning up orphaned files in the entire repository.

none

```
yes, fixes a bug.
```

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -194,7 +194,7 @@ func Do(
 	stackReport := doStackGeneration(root, tree, vendorDir, vendorRequests)
 	rootReport := doRootGeneration(root, tree)
 	report := mergeReports(stackReport, rootReport)
-	return cleanupOrphaned(root, report)
+	return cleanupOrphaned(root, tree, report)
 }
 
 func doStackGeneration(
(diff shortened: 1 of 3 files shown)


soerenmartius (dismissed):
LGTM

generate/generate.go
```diff
@@ -1339,21 +1339,23 @@ func loadStackCodeCfgs(
...
 		Str("action", "generate.cleanupOrphaned()").
+		Stringer("dir", target.Dir()).
 		Logger()
-	// If the root of the tree is a stack then there is nothing to do
-	// since there can't be any orphans (the root parent stack owns
-	// the entire project).
-	if root.Tree().IsStack() {
-		logger.Debug().Msg("project root is a stack, nothing to do")
+
+	defer report.sort()
```
i4ki, line 1348: by luck I got a failing test in the macos CI runner. The problem was in the early return:
```
if target.IsStack() {
	return report
}
```
which was not sorting the report in the case root was an stack.
This only triggers in the new test I added, and rarely for some reason..

snakster (approved):

PR #1892, opened 2024-09-27
test: fix interop test when running on CI push event.

i4ki:

When interop tests run on `push` event, the `GITHUB_EVENT_FILE` needs to be mocked or else previews cannot be created.

Internal CIs.

```
no
```


soerenmartius (approved):
LGTM

PR #1897, opened 2024-10-04
chore: reuse git diff-tree output in change detection.

i4ki:

Improve change detection performance.

none

```
no
```

### DIFF
--- a/stack/manager.go
+++ b/stack/manager.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"github.com/rs/zerolog/log"
 	"github.com/terramate-io/terramate/config"
@@ -31,7 +32,8 @@ type (
 		git  *git.Git
 
 		cache struct {
-			stacks []Entry
+			stacks       []Entry
+			changedFiles []string
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


soerenmartius (approved):
lgtm

stack/manager.go
```diff
@@ -573,7 +575,7 @@ func (m *Manager) tfModuleChanged(
 }
 
 func (m *Manager) tgModuleChanged(
-	stack *config.Stack, tgMod *tg.Module, gitBaseRef string, changedFiles []string, stackSet map[project.Path]Entry, tgModuleMap map[project.Path]*tg.Module,
+	stack *config.Stack, tgMod *tg.Module, gitBaseRef string, stackSet map[project.Path]Entry, tgModuleMap map[project.Path]*tg.Module,
```
snakster, line 578: An idea:
`tgModuleChanged` now relies on the cache being populated, which could lead to bugs in the future.

I think it would be nicer to have a function

```
func (m) cachedChangedFiles(hostDir, gitBaseRef) {
if m.cache.changedFiles == nil {
		var err error

		m.cache.changedFiles, err = m.listChangedFiles(m.root.HostDir(), gitBaseRef)
		if err != nil {
			return nil, errors.E(errListChanged, err)
		}
	}
}
```
that is invoked by the caller of `tgModuleChanged`. In general, make it so it doesn't matter who invoked the computation first.

But only if it's possible, i.e. I don't know if all the callers have the arguments available.

i4ki: good idea, thanks!

cwe1ss:
Cross posting from https://discord.com/channels/1088753599951151154/1288776213107572759/1291731953846059031 to have it here as a reference as well:

results on my client's windows VDI + WSL (ubuntu 24.04):

- ` terramate list -changed` on v0.10.6 takes 12.3sec (8sec user + 3sec sys)
- ` terramate list -changed` on this branch takes 4.6 sec (3.7sec user  +0.4sec sys)

so this is already a great improvement!!! ❤️

stack/manager.go
i4ki: will apply in later PR.

PR #1900, opened 2024-10-04
fix: fallback to file listing if change detection inside a skipped di…

i4ki:

Fix the optimization done in #1898 for the case a directory in the scope of change detection was skipped with `.tmskip`.

It's just a fallback to the older directory listing in this case.

none

```
no because the offending PR is unreleased
```


soerenmartius (dismissed):
lgtm

PR #1904, opened 2024-10-07
feat: consider untracked/uncommitted in the change detector.

### LINKED ISSUE #1398: [FEATURE] Change detection for uncommitted files
The current change detection feature using the `--changed`-flag seems to only work once a change has been committed. 

But there are use-cases, where I would like to execute Terramate scripts on changed stacks only, even before I have committed any files:
* Local build scripts: To run `terraform init`, `terraform validate`, `terraform-docs`, ... via a VS Code build task on demand
* Pre-commit hooks: To validate any changed stacks before the commit is accepted

It takes too long to run these scripts against all stacks within the entire repository every time, so having this additional change detection would be a massive time-saver.

[Pre-commit](https://pre-commit.com/) (and [pre-commit-terraform](https://github.com/antonbabenko/pre-commit-terraform)) have some support for this by sending the changed files to the hook, but this does not work with Terramate's advanced features like [Terraform module change detection](https://terramate.io/docs/cli/change-detection/integrations/terraform#module-change-detection) to detect dependant stacks or [File Watchers](https://terramate.io/docs/cli/change-detection/file-watchers). 

E.g., if I change a file within my reusable module `modules/my-module`, I would also like to run my build script & pre-commit hook against any stacks that use that module.

An option within the CLI that uses the locally changed files (unstaged & staged) as the input for the change detection logic - either integrated into the existing `--changed` flag, or via a new flag.

Using pre-commit with custom scripts, or manually parsing `git diff --name-only` to find changed stacks. But as described, both options will not detect dependant stacks.

**Additional context**
See discussion on discord: https://discord.com/channels/1088753599951151154/1088769572305383484/threads/1200441992233099384

i4ki:

**This is a breaking change**

This change considers untracked and uncommitted files in the change detection.
It means `terramate list --changed` will show stacks changed by uncommitted or untracked files in the index. The `terramate run` safeguards will continue to kick in but the user can disable the `git` safeguards and in this case commands can run in dirty stacks.

Closes #1398 

```
yes, it's a breaking change that needs documentation!
```

### DIFF
--- a/stack/manager.go
+++ b/stack/manager.go
@@ -37,6 +37,12 @@ type (
 		}
 	}
 
+	ChangeConfig struct {
+		BaseRef            string
+		UncommittedChanges *bool
+		UntrackedChanges   *bool
+	}
+
 	// Report is the report of project's stacks and the result of its default checks.
 	Report struct {
 		Stacks []Entry
@@ -112,7 +118,7 @@ func (m *Manager) List(checkRepo bool) (*Report, error) {
 // It's an error to call this method in a directory that's not
 // inside a repository or a repository with no commits in it.
 // It never returns cached values.
-func (m *Manager) ListChanged(gitBaseRef string) (*Report, error) {
+func (m *Manager) ListChanged(cfg ChangeConfig) (*Report, error) {
 	logger := log.With().
 		Str("action", "ListChanged()").
 		Logger()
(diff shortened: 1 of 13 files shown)


stack/manager.go
```diff
@@ -130,11 +136,40 @@ func (m *Manager) ListChanged(gitBaseRef string) (*Report, error) {
 		return nil, errors.E(errListChanged, err)
 	}
 
-	changedFiles, err := m.changedFiles(gitBaseRef)
+	_, err = m.changedFiles(cfg.BaseRef)
```
snakster, line 139: I think this is hard to reason about, the memoization is just an internal thing, but here the caller relies on that behaviour.

Would it be possible to add `checks.UncommittedFiles` and `checks.UntrackedFiles` as params to m.changedFiles (with the option of them being empty), then do the appending in there?

snakster (approved):
👍

PR #1909, opened 2024-10-10
refactor!: remove deprecated experimental run-order.

i4ki:

The command `terramate experimental run-order` was deprecated a while ago in preference of `terramate list --run-order`.

none

```
yes, remove an old feature.
```


soerenmartius:
just doubled check the docs also, they are already up to date

PR #1918, opened 2024-10-13
test: fix preview tests on Windows.

i4ki:

The script preview tests were embedding a host path in the command string but Windows paths need backlash escaping for Windows. Instead of escaping, the fix is to just use "terraform" and rely on the %Path% environment having the Terraform installation directory prepended.

none

```
no
```


soerenmartius (approved):
LGTM

PR #1933, opened 2024-10-17
ci: fix change detection in backport branches.

i4ki:
## What this PR does / why we need it:

Previews targeting `v0.10.x` branch were comparing changes against `main`.

## Which issue(s) this PR fixes:
none

## Does this PR introduce a user-facing change?
```
no
```

PR #1938, opened 2024-10-22
ci: fix change detection in push event.

i4ki:

Terramate CI rollout releases in `main` and `vX.Y.x` branches. The PR #1933 works for previews but broke the TMC deployment because push events has no `github.base_ref` variable.

The solution is to compare against first parent commit (the default in Terramate but not in branches other than main);

none

```
no
```


soerenmartius (approved):
lgtm

PR #1937, opened 2024-10-21
refactor: parallel generate

i4ki:
## What this PR does / why we need it:

Introduces parallel code generation.

## Which issue(s) this PR fixes:
none

## Does this PR introduce a user-facing change?
```
yes, performance
```

PR #1943, opened 2024-10-23
chore: avoid in-place modifications of AST nodes in the partial evaluation

i4ki:

The in-place AST modifications were a nice optimization when generation was serial but with the introduction of #1937 it's racy.

none
```
no
```


zied-elouaer (dismissed):
Very nice. This branch works perfectly fine on the customer repo and take 6 seconds to run generate there!

snakster (approved):
👍 
It could be nice have a test in the future that invokes `partialEval` (or something further up the call tree) in parallel to potentially catch any remaining race conditions.

snakster (approved):

PR #1947, opened 2024-10-24
chore: pin github actions.

i4ki:

For security reasons we should pin the github action versions.

A `make pin-gha` command is added to automate this.

none

```
no
```


soerenmartius (approved):
lgtm

PR #1948, opened 2024-10-24
chore: use hasUnknowns from partialEval.

i4ki:

The partial evaluator returns a `hasUnknowns` boolean that can be used by the `tm_ternary()` to figure if unknowns are still present in the expr, then we can avoid walking the resulted expression.

none

```
no
```


soerenmartius (approved):
lgtm

PR #1955, opened 2024-11-18
feat: add terramate.stack.parent.id.

i4ki:

Implements `terramate.stack.parent.id` metadata for stacks that has parent stacks.

none

```
yes, add a new feature
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Added
+
+- Add `terramate.stack.parent.id` metadata to stacks that are part of a parent-child hierarchy.
+
 ## v0.11.1
 
 ### Added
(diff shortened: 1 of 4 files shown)


soerenmartius (approved):
LGTM, I just tested this throughly as well.

- Works with directly nested stacks
- Works with deeply nested stacks whereas multiple levels of directories can be in between stacks

Great job!

soerenmartius:
Only suggestion is that we could throw a better exception in case no parent stack is found but it's non-critical

> - /02_example_data_sources/network/aks
        error: data.tf.tmgen:3,12-8,4: evaluating content: tmgen file "/02_example_data_sources/network/aks/data.tf.tmgen": partial evaluation failed: data.tf.tmgen:7,69-76: eval expression: This object does not have an attribute named "parent".

i4ki:
@soerenmartius the field `terramate.stack.parent` is only defined if there's a parent stack, otherwise it's not populated. As this is not a function call, but just accessing an object field/member, we can't easily inject a different error message. If this was a function call, like `tm_parent_stack()` then for sure it would make sense to return an error saying that no parent stack exists.

wdyt?

PR #1957, opened 2024-11-19
fix: nested maps not rendered if no value attribute is present.

i4ki:

The nested `map` blocks are not rendered if the `value` block do not contain attributes.

none

```
yes, fixes a bug.
```


soerenmartius (approved):
LGTM

PR #1965, opened 2024-11-20
chore: upgrade Kong library to v1.4.0

i4ki:

Upgrade Kong in preparation for upcoming cli help changes.

none

```
no
```


soerenmartius (approved):
lgtm

soerenmartius (approved):
lgtm

sixstone-qq (approved):

cmd/terramate/cli/run.go
```diff
@@ -119,6 +119,12 @@ func selectPlanFile(terraformPlan, tofuPlan string) (planfile, provisioner strin
 func (c *cli) runOnStacks() {
 	c.gitSafeguardDefaultBranchIsReachable()
 
+	if len(c.parsedArgs.Run.Command) > 0 {
```
snakster, line 122: Can you briefly explain why this becomes necessarily now?

i4ki: they change the behavior of specifically `arg:"" passthrough:""`:
https://github.com/alecthomas/kong/pull/436

Maybe I'm missing something but I cannot make it work in in the way it was before...
The `passthrough:""` prevents kong from interpretating the flags, so it's needed otherwise commands with same flags as Terramate gets interpreted. The `arg:""` is also needed to ... make it an argument, unless there's other option to keep old behavior :(

PR #1973, opened 2024-11-26
chore: refactor project paths in change detection.

i4ki:

While fixing #1971 another problem was identified that prevented the fix.
The `project.PrjAbsPath()` function had many misuses that worked by luck in many cases but the other fix depends on this one.

none

```
no
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -1221,10 +1221,10 @@ func (c *cli) checkGitUncommited() bool {
 	return !cfg.Terramate.Config.HasSafeguardDisabled(safeguard.GitUncommitted)
 }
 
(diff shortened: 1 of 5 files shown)


cmd/terramate/cli/cli.go
```diff
@@ -2335,7 +2335,7 @@ func (c *cli) setupEvalContext(st *config.Stack, overrideGlobals map[string]stri
 			globalPath,
 			expr,
 			info.NewRange(c.rootdir(), hhcl.Range{
-				Filename: "<eval argument>",
+				Filename: filepath.Join(c.rootdir(), "<cmdline>"),
```
i4ki, line 2338: Improving this requires another refactor :(
I think for now it's fine

PR #2024, opened 2024-12-30
fix: race condition with parallel + enable_sharing

snakster:

Fixes #

```

```


e2etests/core/run_sharing_test.go
```diff
@@ -582,3 +583,85 @@ func TestRunSharing(t *testing.T) {
...
+		IgnoreStdout: true,
+	})
+	AssertRunResult(t, tmcli.Run("run", "--quiet", "-X", "terraform", "init"),
+		RunExpected{
+			IgnoreStdout: true,
+		},
+	)
+	s.Git().CommitAll("all")
+
+	AssertRunResult(t, tmcli.Run("run", "--quiet", "--enable-sharing", "--mock-on-fail", "--parallel=10", "terraform", "apply", "-auto-approve"),
```
sixstone-qq, line 662: Don't we want to check something in the stdout?

run/oncemap.go
```diff
@@ -0,0 +1,43 @@
...
+
+// GetOrInit obtains a value for the given key if the value exists in the map, or
+// initializes it with the given init function.
+// This function is safe to be called concurrently.
+func (m *OnceMap[K, V]) GetOrInit(k K, init func() (V, error)) (V, error) {
+	m.mtx.RLock()
+	v, found := m.data[k]
+	m.mtx.RUnlock()
+
+	if !found {
```
sixstone-qq, line 27: I'd say to use early return here to ease the readibility:
```suggestion
	if found {
	    return v, nil
        }
```

run/oncemap.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2023 Terramate GmbH
+// SPDX-License-Identifier: MPL-2.0
+
+package run
+
+import "sync"
+
+// OnceMap contains map of key/value pairs that is safe for concurrent initialization.
+type OnceMap[K ~string, V any] struct {
```
sixstone-qq, line 9: I don't get why don't we use https://pkg.go.dev/sync@go1.23.4#Map here.

I think it's optimised for this use case:
>  when the entry for a given key is only ever written once but read many times, as in caches that only grow

In addition to this, `GetOrInit` is quite similar to `LoadOrStore` :)

run/oncemap.go
```diff
@@ -0,0 +1,43 @@
...
+// This function is safe to be called concurrently.
+func (m *OnceMap[K, V]) GetOrInit(k K, init func() (V, error)) (V, error) {
+	m.mtx.RLock()
+	v, found := m.data[k]
+	m.mtx.RUnlock()
+
+	if !found {
+		m.mtx.Lock()
+		v, found = m.data[k]
+		if !found {
```
sixstone-qq, line 30: Same as above about early return

sixstone-qq (commented):
Great work, just a few comments :)

run/oncemap.go
snakster: I looked at it initially, but it's not the same. `sync.Map` supports concurrent access, but it doesn't coordinate the execution of the init function - `LoadOrStore` expects a value, not the init func, and we must only execute this function once (it runs an external command).

I also considered using `sync.OnceValue`, but you still need a thread-safe map.

e2etests/core/run_sharing_test.go
snakster: Copied this from the other related test cases, so I think it's fine. stdout will contain the terraform specific output. We just want to test that there is no error.

run/oncemap.go
snakster: Done in [9e11aeb](https://github.com/terramate-io/terramate/pull/2024/commits/9e11aeb6457a550b0e0d30ffa68c2fdb5c03d61d)

run/oncemap.go
snakster: Done in [9e11aeb](https://github.com/terramate-io/terramate/pull/2024/commits/9e11aeb6457a550b0e0d30ffa68c2fdb5c03d61d)

run/oncemap.go
sixstone-qq: You can combine both, no? `map.LoadOrStore(key, sync.OnceValue(your_fn))`?

Anyway, it's just a suggestion to reduce the code , not the code per se which is working as expected. We can dismiss this comment :)

sixstone-qq (approved):

mariux (approved):

PR #2036, opened 2025-01-10
feat: terramate experimental clone --no-generate

### LINKED ISSUE #2007: [FEATURE] Support --no-generate for terramate experimental clone
We use some template stacks to reduce boilerplate that we can't/don't want to handle via globals. When running `terramate experimental clone`, Terramate immediately calls "terramate generate". However, we would like to do some more post-processing before actually running "terramate generate". 

Support the flag `--no-generate` (which already exists for `terramate create`) for `terramate experimental clone`. When it is set, Terramate should *not* invoke code generation after the stack has been cloned.

i4ki:

Add `--no-generate` to `terramate experimental clone`.

Fixes #2007 

```
yes, add `--no-generate` to clone
```


soerenmartius (approved):
LGTM, thanks!

snakster (approved):

PR #2041, opened 2025-01-28
feat: promote trigger command

snakster:

This PR promotes the trigger command from experimental to stable.
The old command is still there and working, but hidden from the help.

The behaviour is changed slightly so that when a trigger file fails to parse, terramate will return an error instead of printing a warning and continuing.

I see no reason to declare this as a breaking change, because even before the trigger would not work as expected and I can see no reason why anyone would depend on this. Let me know in the review if you disagree.

```
Yes, it moves the trigger command from the experimental command group to the top level.
```


soerenmartius (dismissed):

i4ki (dismissed):
LGTM

i4ki (approved):
LGTM

PR #2044, opened 2025-02-04
chore: refactor cloud SDK to use text/plain for bytes payload.

i4ki:

The cloud SDK at the moment only supports `application/json` payloads but other products makes use of `text/plain` content type.

none

- not used in the CLI.

```
no
```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -327,12 +327,8 @@ func Get[T Resource](ctx context.Context, client *Client, u url.URL) (entity T, 
 
 // Post requests the endpoint components list making a POST request and decode the response into the
 // entity T if validates successfully.
-func Post[T Resource](ctx context.Context, client *Client, payload interface{}, url url.URL) (entity T, err error) {
-	dataPayload, err := json.Marshal(payload)
-	if err != nil {
-		return entity, errors.E(err, "marshaling request payload")
-	}
-	resource, err := Request[T](ctx, client, "POST", url, bytes.NewBuffer(dataPayload))
+func Post[T Resource](ctx context.Context, client *Client, payload any, url url.URL) (entity T, err error) {
+	resource, err := Request[T](ctx, client, "POST", url, payload)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cloud/cloud.go
```diff
@@ -369,14 +357,14 @@ func Put[T Resource](ctx context.Context, client *Client, payload interface{}, u
 
 // Request makes a request to the Terramate Cloud using client.
 // The instantiated type gets decoded and return as the entity T,
-func Request[T Resource](ctx context.Context, c *Client, method string, url url.URL, postBody io.Reader) (entity T, err error) {
+func Request[T Resource](ctx context.Context, c *Client, method string, url url.URL, payload any) (empty T, err error) {
```
sixstone-qq, line 360: This returned variable is misleading, could we rename it? From my perspective, it sounds like always empty response.

cloud/cloud.go
```diff
@@ -496,4 +490,27 @@ func (c *Client) dumpRequest(req *http.Request) ([]byte, error) {
 	return httputil.DumpRequestOut(reqCopy, true)
 }
 
-const contentType = "application/json"
+func preparePayload(payload any) (io.Reader, string, error) {
```
sixstone-qq, line 493: _nitpick_ that would help reading
```suggestion
func preparePayload(payload any) (body io.Reader, contentType string, err error) {
```

sixstone-qq (commented):
Just two small comments.

cloud/cloud.go
sixstone-qq:
```suggestion
func Request[T Resource](ctx context.Context, c *Client, method string, url url.URL, payload any) (res T, err error) {
```

sixstone-qq: btw, it's not blocking to approve the PR

i4ki: all good, I improved docs as well.
thanks

sixstone-qq (approved):

PR #2046, opened 2025-02-05
chore: implement /v1/store sdk methods and testserver handlers.

i4ki:

Improve the cloud sdk package adding client methods for CRUD operations in the `/v1/store` endpoint and testserver handlers for e2e tests.

none

This is needed for implementation and tests of the provider.

```
no
```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -53,6 +53,8 @@ const (
 	StacksPath = "/v1/stacks"
 	// ReviewRequestsPath is the review requests endpoint base path.
 	ReviewRequestsPath = "/v1/review_requests"
+	// StorePath is the store endpoint base path.
+	StorePath = "/v1/store"
 )
 
 // ErrUnexpectedStatus indicates the server responded with an unexpected status code.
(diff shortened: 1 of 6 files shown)


cloud/cloud.go
```diff
@@ -355,6 +403,12 @@ func Put[T Resource](ctx context.Context, client *Client, payload interface{}, u
 	return resource, nil
 }
 
+// Delete requests the endpoint url with a DELETE method.
+func Delete[T Resource](ctx context.Context, client *Client, url url.URL) error {
+	_, err := Request[T](ctx, client, "DELETE", url, nil)
```
sixstone-qq, line 408: _nitpick_ Why don't you use `http.Method*` constants?
```suggestion
	_, err := Request[T](ctx, client, http.MethodDelete, url, nil)
```

sixstone-qq (approved):

cloud/cloud.go
i4ki: I actually don't know :) but I used the plain string in all other methods so I will keep it this way for now for consistency.

PR #2047, opened 2025-02-06
fix(terragrunt): make Terragrunt keep command output as is.

i4ki:

Newer versions of Terragrunt transform the user-supplied command output and redirect stdout/stderr and this does not work well with Terramate, which always respects the user command stdout/stderr.

This change exports `TERRAGRUNT_FORWARD_TF_STDOUT=true` and `TERRAGRUNT_LOG_FORMAT=bare` in order to force Terragrunt to keep old behavior.

none

- [x] Changelog

```
yes fixes an issue found at a customer.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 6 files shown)


soerenmartius (dismissed):
LGTM

snakster (dismissed):

snakster (approved):

PR #2055, opened 2025-02-10
fix(terragrunt,tmc): only export TG envs during changeset creation.

i4ki:

As discussed, it's better if we export the `TERRAGRUNT_FORWARD_TF_STDOUT` and `TERRAGRUNT_LOG_FORMAT` only during the changeset details.

no
```
not released yet
```


snakster (approved):
👍

PR #2059, opened 2025-02-13
chore: add support for region in the TMC sdk.

i4ki:

This is required for an upcoming change in the CLI configuration for setting the Terramate Cloud region.

none
```
no
```

### DIFF
--- a/cloud/cloud.go
+++ b/cloud/cloud.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 
 	hversion "github.com/apparentlymart/go-versions/versions"
 	"github.com/rs/zerolog"
@@ -29,11 +30,8 @@ import (
 	"github.com/terramate-io/terramate/versions"
 )
 
-// Host of the official Terramate Cloud API.
-const Host = "api.terramate.io"
-
-// BaseURL is the default cloud.terramate.io base API URL.
-const BaseURL = "https://" + Host
+// BaseDomain is the Terramate Cloud base domain.
+const BaseDomain = "api.terramate.io"
 
 const defaultPageSize = 50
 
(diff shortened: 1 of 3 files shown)


cloud/cloud.go
```diff
@@ -83,8 +85,13 @@ type (
 
 		Logger *zerolog.Logger
 		noauth bool
+
+		mu sync.Mutex
```
sixstone-qq, line 89:
```suggestion
                // mu guards the changes in Client attributes concurrently.
		mu sync.Mutex
```

cloud/cloud.go
```diff
@@ -96,6 +103,13 @@ type (
 	}
 )
 
+// Available cloud locations.
+const (
+	EU Region = iota
```
sixstone-qq, line 108: IMHO it's strange to use `invalidRegion` as latest value in const enum instead of the first one and keep default EU as `iota+1`

sixstone-qq (commented):
Left two comments, disregard them if you don't agree

cloud/cloud.go
i4ki: because the `cloud.Client` has no constructor function, the idea is that just doing `cloud.Client{}` will automatically configure `cloud.Client.Region` as `EU` by default (using the zero value). So it uses the zero value.

I actually forgot to remove the `invalidRegion` because before I was "validating" the region with `if c.Region >= invalidRegion { return errors.E(...) }` but later decided to just validate for 'EU` and `US` for now, as I don't think this enum will be extended further ever.

i4ki: actually I will need `invalidRegion` in the last PR

cloud/cloud.go
i4ki: added this in my last PR (just want to avoid rebasing all of them, sorry)
thank you

soerenmartius (approved):
LGTM

cloud/cloud.go
sixstone-qq: I noticed in other PRs I'd say to add a comment because it's really confusing

sixstone-qq (approved):

cloud/cloud.go
i4ki: I will add a comment but just to clarify further: we usually make the first enum the illegal one to avoid the zero value misrepresenting something. But here we actually do want to make EU the default region.

but yes, a comment makes sense because usually you want to prevent that.
In the future, if we introduce a construtor to the `cloud.Client` then I think makes sense to make the zero value illegal.

Thank you!

PR #2062, opened 2025-02-14
fix: misleading error message when TMC onboarding is incomplete.

i4ki:

When the user uses `terramate cloud login` but still haven't signup in Terramate, we were giving an error message saying he needs authenticate in the CLI when actually the correct message is that he needs to visit the cloud website and signup.

none

```
yes, fixes an issue
```

### DIFF
--- a/e2etests/cloud/run_cloud_uimode_test.go
+++ b/e2etests/cloud/run_cloud_uimode_test.go
@@ -318,7 +318,7 @@ func TestCloudSyncUIMode(t *testing.T) {
 					want: RunExpected{
 						Status: 1,
 						StderrRegexes: []string{
-							string(clitest.CloudLoginRequiredMessage),
+							string(clitest.CloudOnboardingIncompleteMessage),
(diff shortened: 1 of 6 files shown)


e2etests/cloud/run_cloud_uimode_test.go
```diff
@@ -318,7 +318,7 @@ func TestCloudSyncUIMode(t *testing.T) {
 					want: RunExpected{
 						Status: 1,
 						StderrRegexes: []string{
-							string(clitest.CloudLoginRequiredMessage),
+							string(clitest.CloudOnboardingIncompleteMessage),
```
i4ki, line 321: all of those tests were actually for incomplete onboarding because it fakes a 404 in the `/v1/users` endpoint, which means user is authenticated but didn't signup.

soerenmartius (approved):
LGTM

zied-elouaer (approved):

PR #2071, opened 2025-02-19
fix(tmc): github login should check "email_verified" from JWT.

i4ki:

When using `terramate cloud login --github` the CLI is giving an incorrect error telling the user that email is not verified. The problem is because it's checking the `email_verified` field from the response of the `signInWithIDP` Firebase endpoint and the field relates to the federated `email_verified` state but this is not important for Terramate Cloud. The JWT `email_verified` claim that should be used instead.

Fixes an issue reported by a customer.

```
yes
```


mariux (approved):
lgtm

zied-elouaer (approved):

PR #2077, opened 2025-02-24
chore(gha): quote shell arguments controlled by contributors.

i4ki:


cmd/terramate/main.go
```diff
@@ -3,7 +3,7 @@
 
 // Terramate is a tool for managing multiple Terraform stacks. Providing stack
 // execution orchestration and code generation as a way to share data across
-// different stacks.
```
sixstone-qq, line 6: This change should not be here :)

i4ki: I used this change to trigger the workflow because it only run if any Go file changes. Just for testing purposes. Removing now.

PR #2079, opened 2025-02-26
refactor!: remove default organization fallback if not set.

i4ki:

Note: This is a **BREAKING CHANGE**.

In order to avoid confusion and misconfiguration we are removing the organization fallback in case the user don't select an organization. Before, if the user had exactly 1 organization membership with `status=active` the CLI used it automatically.

Now users will need to explicitly set `terramate.config.cloud.organization` or export `TM_CLOUD_ORGANIZATION=<my org name>`.

none

```
yes, introduces a breaking change.
```


cloud/testserver/cloudstore/store.go
```diff
@@ -152,17 +152,21 @@ const (
...
-func LoadDatastore(fpath string) (*Data, error) {
+func LoadDatastore(fpath string) (store *Data, defaultOrg string, err error) {
 	data, err := os.ReadFile(fpath)
 	if err != nil {
-		return nil, errors.E(err, "failed to read testserver data")
+		return nil, "", errors.E(err, "failed to read testserver data")
 	}
-	var dstore Data
+	type serverData struct {
+		DefaultOrg string `json:"default_org"`
```
sixstone-qq, line 161: Instead of being the default org, I'd name it selected one:
```suggestion
		SelectedOrg string `json:"selected_org"`
```
What do you think? It's more accurate in my opinion with the use case.

e2etests/cloud/generate-e2e-testserver.tm
```diff
@@ -23,9 +23,10 @@ generate_file "testdata/cloud.data.json" {
 
   # default organizations of the e2e test server.
   lets {
+    default_org = "terramate"
     orgs = {
-      terramate = {
-        name         = "terramate"
+      (let.default_org) = {
```
sixstone-qq, line 28: :magic_wand:

sixstone-qq (commented):
Only one suggestion , let me know what you think

cloud/testserver/cloudstore/store.go
i4ki: the idea is that this is the "default org used in tests". I shorted it to "default_org".
Using "selected_org" would make more sense if this field were picked up by the CLI itself, I think.

sixstone-qq (approved):

cloud/testserver/cloudstore/store.go
i4ki: I changed to `default_test_org` to make the intention more clear. As I said, if using `selected_org` imply this is picked by CLI automatically or only this one is used which is not true, some tests use a different org.

sixstone-qq (commented):
Some questions otherwise LGTM

cmd/terramate/cli/tmcloud/auth/cloud_credential_github_oidc.go
```diff
@@ -206,29 +206,31 @@ func (g *githubOIDC) fetchDetails() error {
 }
 
 func (g *githubOIDC) Info(selectedOrgName string) {
-	if len(g.orgs) > 0 && g.orgs[0].Status == "trusted" {
-		printer.Stdout.Println("status: signed in")
-	} else {
+	printer.Stdout.Println(fmt.Sprintf("provider: %s", g.Name()))
+	if selectedOrgName == "" {
+		printer.Stdout.Println("status: untrusted")
```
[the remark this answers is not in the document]
i4ki: thanks for asking, I just removed this line in both OIDC cases.
My idea was to say that it's always `untrusted` if you don't set the organization but this is misleading... someone could think that TMC settings config is missing when in fact everything is right but just the org is not set.

thank you

cmd/terramate/cli/tmcloud/auth/cloud_credential_google.go
```diff
@@ -689,20 +688,44 @@ func (g *googleCredential) Info(selectedOrgName string) {
...
+			),
+		)
+		return
 	}
 
-	if len(g.orgs) == 0 {
-		printer.Stderr.Warnf("You are not part of an organization. Please visit %s to create an organization.", cloud.HTMLURL(g.client.Region))
+	org, found := g.orgs.LookupByName(selectedOrgName)
+	if found {
+		if org.Status != "active" {
```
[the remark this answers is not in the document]
i4ki: I know, but they are used everywhere, also in tests. I don't like inconsistencies, then I will try to fix this soon in the entire repo.

PR #2086, opened 2025-03-10
feat: add support for SingleSignOn login.

i4ki:

Implements `terramate cloud login --sso` for Single-Sign-On login.

none
```
yes, adds a new flag to `terramate cloud login`
```

### DIFF
--- a/cmd/terramate/cli/cli.go
+++ b/cmd/terramate/cli/cli.go
@@ -187,8 +187,9 @@ type cliSpec struct {
 
 	Cloud struct {
 		Login struct {
-			Google bool `optional:"true" help:"authenticate with google credentials"`
-			Github bool `optional:"true" help:"authenticate with github credentials"`
+			Google bool `optional:"true" xor:"login_provider" help:"authenticate with google credentials"`
+			Github bool `optional:"true" xor:"login_provider" help:"authenticate with github credentials"`
+			SSO    bool `optional:"true" xor:"login_provider" help:"authenticate with SSO credentials"`
 		} `cmd:"" help:"Sign in to Terramate Cloud."`
 		Info  struct{} `cmd:"" help:"Show your current Terramate Cloud login status."`
 		Drift struct {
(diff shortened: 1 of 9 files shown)


cmd/terramate/cli/tmcloud/auth/cloud_credential_google.go
```diff
@@ -161,13 +168,18 @@ func startServer(
...
+			ports = append(ports, minPort+rng.Intn(maxPort-minPort))
+		}
+	}
 
 	var ln net.Listener
-	const maxretry = 5
-	var retry int
-	for retry = 0; retry < maxretry; retry++ {
-		addr := "127.0.0.1:" + strconv.Itoa(minPort+rng.Intn(maxPort-minPort))
+	for _, port := range ports {
```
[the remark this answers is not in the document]
i4ki: we discussed using this at the time but it's not portable. For example, MacOS gives sequential ports without checking if they are in use. It's also not clear how other operating systems handle it (openbsd, freebsd, etc) so we decided to pick a random port and retry 5x. This is working fine for almost two years so I don't think it's worth changing now.

cmd/terramate/cli/cli.go
```diff
@@ -187,8 +187,9 @@ type cliSpec struct {
 
 	Cloud struct {
 		Login struct {
-			Google bool `optional:"true" help:"authenticate with google credentials"`
-			Github bool `optional:"true" help:"authenticate with github credentials"`
+			Google bool `optional:"true" xor:"login_provider" help:"authenticate with google credentials"`
+			Github bool `optional:"true" xor:"login_provider" help:"authenticate with github credentials"`
+			SSO    bool `optional:"true" xor:"login_provider" help:"authenticate with SSO credentials"`
```
sixstone-qq, line 192: Do you need to provide a flag? As the endpoint call latency is low and guess it based on the response. It is more convenient to end user as the organization name is now mandatory to sign with TMC

i4ki: Good question. Theoretically we can detect that current project uses SSO and do it automatically but I don't think this is a good idea. The `--google` is the default option and I think it's confusing if this is overridden if executed from a SSO based project.

@mariux wdyt?

cmd/terramate/cli/tmcloud/auth/cloud_credential_google.go
```diff
@@ -43,6 +43,12 @@ const (
 	maxPort = 52023
 
 	defaultGoogleTimeout = 60 * time.Second
+
+	googleProviderID = "google.com"
+	ssoProviderID    = "oidc.workos"
```
snakster, line 48: Could  define `ssoProviderID` and `ssoOauthScope` in loud_credential_sso.go instead.

snakster (approved):
👍

PR #2104, opened 2025-03-27
chore: miscellanious improvements in the workflow tooling.

i4ki:

Use globals to configure the tooling versions.

none
```
no
```

### DIFF
--- a/.github/workflows/ci-sync-preview.yml
+++ b/.github/workflows/ci-sync-preview.yml
@@ -63,6 +63,9 @@ jobs:
       - name: make generate
         run: make generate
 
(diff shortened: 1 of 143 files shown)


.github/workflows/ci-sync-preview.yml
```diff
@@ -90,9 +84,6 @@ jobs:
         if: ${{ steps.list_go_packages.outputs.stdout || steps.list_e2e_packages }}
         run: make license/check
 
-      - name: Check Terramate formatting
```
i4ki, line 93: fmt was running twice :/

PR #2106, opened 2025-03-28
chore: install Go from go.mod in GHA workflows

i4ki:

Use same Go from go.mod in tests.

none

```
no
```

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@v5
         with:
-          go-version: '1.21'
+          go-version-file: 'go.mod'
 
       - name: run benchcheck
         id: benchmark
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@v5
         with:
-          go-version: '1.21'
+          go-version-file: 'go.mod'
```
sixstone-qq, line 31: you could cache dependencies with:
```suggestion
          go-version-file: 'go.mod'
          cache-dependency-path: go.sum
```
Take into account, that's the minimal version to use but not the maximum as stated in their [doc](https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file)

i4ki: I think it caches by default, no? The docs seems to imply that you need this option in case you have the go.sum elsewhere in the repo (not at root).
Fri Mar 28 02:22:00 PM UTC 2025

or am I wrong?

PR #2118, opened 2025-04-09
fix: parsers reusing states in subparsers.

i4ki:

In order to pass hcl options downstream to sub-parsers they need to be re-initialized otherwise they can end up updating the wrong `hcl.Config`.

none

```
no, because the related refactor is not released.
```


snakster (approved):
LGTM

PR #2139, opened 2025-04-22
refactor: improve configuration loading.

i4ki:

This change is required for upcoming performance optimizations in the Terragrunt module detection.

none

Main goal of this change is making the *Root available during the loadTree() tree walking for allowing a context and pool of workers for the Terragrunt async loading.

```
no
```

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/terramate-io/terramate/config"
 	"github.com/terramate-io/terramate/errors"
 	"github.com/terramate-io/terramate/project"
+	. "github.com/terramate-io/terramate/test/hclwrite/hclutils"
 	"github.com/terramate-io/terramate/test/sandbox"
 )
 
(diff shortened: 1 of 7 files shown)


config/config_test.go
```diff
@@ -32,6 +33,80 @@ func TestIsStack(t *testing.T) {
...
+			layout: []string{
+				"f:config.tm:" + Block("terramate",
+					Str("required_version", fmt.Sprintf("= %s", terramate.Version())),
+				).String(),
+			},
+			fromdir: "/",
+			want:    want{cfgpath: "/", found: true},
+		},
+		{
+			name: "has root config - find from nested subdir",
```
i4ki, line 76: this case was only tested indirectly in the e2e but it can go undetected if all tests use a git repo... as this is only used if it's not a repo and `terramate.required_version` is used.

ui/tui/cli.go
```diff
@@ -264,7 +264,7 @@ func (c *CLI) Exec(args []string) {
 	}
 
 	if !foundRoot {
-		printer.Stderr.Fatal(`Error: Terramate was unable to detect a project root.
```
i4ki, line 267: it was printing: `Error: Error: Terramate was unable...`

snakster (approved):

PR #2140, opened 2025-04-23
refactor(perf): concurrent Terragrunt module parsing.

i4ki:

Refactor configuration loading for parallel parsing of Terragrunt modules.

none
```
no
```

### DIFF
--- a/commands/generate/generate.go
+++ b/commands/generate/generate.go
@@ -98,7 +98,7 @@ func (s *Spec) Exec(_ context.Context) error {
 
 	log.Trace().Msg("all handlers stopped, generating final report")
 
-	if s.PrintReport {
+	if s.PrintReport || report.HasFailures() {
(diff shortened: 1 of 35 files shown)


commands/generate/generate.go
```diff
@@ -98,7 +98,7 @@ func (s *Spec) Exec(_ context.Context) error {
 
 	log.Trace().Msg("all handlers stopped, generating final report")
 
-	if s.PrintReport {
+	if s.PrintReport || report.HasFailures() {
```
i4ki, line 101: small issue: when a `experimental clone` fails to generate, no report was being shown. I think this was introduced in the commands refactor.

i4ki: got this when solving issues with the `config.LoadSubTree()`

snakster (dismissed):
👍

PR #2149, opened 2025-05-30
refactor: Do not read body unless required

sixstone-qq:

This could improve a little the performance upon non-successful
requests.

```
no
```


sixstone-qq:
No merging it as it will affect the connection pooling

PR #2161, opened 2025-06-19
feat: allow tags in wants and wanted_by

snakster:

Fixes #

```

```


run/order.go
```diff
@@ -192,11 +192,6 @@ func BuildDAG(
 		uniqPaths := map[string]struct{}{}
 		for _, pathstr := range paths {
 			if strings.HasPrefix(pathstr, "tag:") {
-				if fieldname != "before" && fieldname != "after" {
```
sixstone-qq, line 195: I'd say it's more secure to do this instead:
```go
if !slices.Contains([]string{"after", "before",  "wanted_by", "wants"}, fieldName) {
```

Could we do that instead?

snakster: The function is only ever called with these 4 values, so I would find this check confusing, as it would do nothing. If we ever add another link attribute, we can still decide if we want to support `tag:` there or not, but it is not insecure.

PR #2162, opened 2025-06-19
chore: hide deprecated experimental trigger in help

snakster:

Trigger has been promoted to stable so the experimental version should be removed from the help listing.

```
No, just the help is affected.
```

### DIFF
--- a/ui/tui/cli_spec.go
+++ b/ui/tui/cli_spec.go
@@ -109,7 +109,7 @@ type FlagSpec struct {
 		IgnoreChange bool   `default:"false" help:"Trigger stacks to be ignored by change detection"`
 		Reason       string `default:"" name:"reason" help:"Set a reason for triggering the stack."`
 		cloudFilterFlags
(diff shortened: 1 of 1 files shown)


ui/tui/cli_spec.go
```diff
@@ -109,7 +109,7 @@ type FlagSpec struct {
 		IgnoreChange bool   `default:"false" help:"Trigger stacks to be ignored by change detection"`
 		Reason       string `default:"" name:"reason" help:"Set a reason for triggering the stack."`
 		cloudFilterFlags
-	} `cmd:"" help:"Mark a stack as changed so it will be triggered in Change Detection."`
+	} `cmd:"" hidden:""  help:"Mark a stack as changed so it will be triggered in Change Detection. (DEPRECATED)"`
```
sixstone-qq, line 112: There is an inconsistency with `hidden:"true"`, but not at the scope of this PR

PR #2165, opened 2025-06-24
fix!: improve performance of stack order evaluation

### LINKED ISSUE #2164: [BUG] Performances issues due to a high number of tags
**Performances issues**
When we want to run a simple command `terramate run` or `terramate script run` the command will hang and maybe wil provide a return (after 5 minutes, I stop the command).
We use the tags for differents things : ordering + managing providers and generating required.providers.tf properly.

**To Reproduce**

1. Given context, see attachments. 
2. Run command `terramate list` and see it returns almost immediately
3. Run command `terramate list --run-order` and see the command will hang

**Expected behavior**
The command will return in a timely manner.

**Log Output**
We don't have any logs for this error, even the verbose command does not provide any more output.

 - OS: Linux
 - OS Version Ubuntu 20.04
 - Git Version 2.49.0
 - Terramate Version 0.13.2

**Additional context**

Attachment 1 : `terramate debug show metadata`

[terramate-debug-show-metadata-result.txt](https://github.com/user-attachments/files/20879364/terramate-debug-show-metadata-result.txt)

snakster:

This PR improves the graph traversal algorithm to determine the stack run order, because it would scale poorly for graph layouts with many edges. Such graphs would typically result when using tags and before/after.

As a result of changing the traversal algorithm, the order for unrelated stacks may change.
Before, it was a DFS traversal, now its BFS. See changelog for some examples.

Fixes #2164 

```
Yes, it changes the run order between unrelated stacks.
```


e2etests/core/exp_run_graph_test.go
```diff
@@ -175,7 +175,7 @@ func TestOrderGraphAfter(t *testing.T) {
 				digraph  {n1[label="stack-a"];
 					n2[label="stack-b"];
 					n1->n2[color="red"];
-					n2->n1;
+					n2->n1[color="red"];
```
snakster, line 178: Cycles are now shown in more edges in the graph. Here we have a cycle a->b->a, and both edges are red. I think the new one is more correct than it was before, but this is just for debugging.

run/dag/dag.go
```diff
@@ -98,26 +102,26 @@ func (d *DAG[V]) addAncestor(node, ancestor ID) {
...
 // Reduce removes nodes that match the given predicate.
 // When a node is removed, all edges to it are replaced by edges to its children,
 // so the order within the graph is preserved.
+// May only be called after Validate().
 func (d *DAG[V]) Reduce(predicate func(id ID) bool) {
-	visited := Visited{}
-
 	shouldRemove := make(map[ID]bool, len(d.dag))
-	ids := d.IDs()
+	ids := d.Order()
```
snakster, line 124: Reducing means if we identify nodes b and c to be removed, we re-connect the graph so that
a <- b <- c <- d
=>
a <- d

Doing this from "left to right" by using Order(), which is the similiar behaviour as the walkFrom before would do.

run/dag/dag.go
```diff
@@ -153,6 +150,14 @@ func (d *DAG[V]) Reduce(predicate func(id ID) bool) {
 			delete(d.values, id)
 		}
 	}
+
+	newOrder := make([]ID, 0, len(d.order))
```
snakster, line 155: Order sequence is cached. We do not have to re-compute it after we removed nodes, we just remove the nodes from the cached order.

run/dag/dag_test.go
```diff
@@ -222,6 +222,7 @@ func dagTests() []testcase {
 				"B": {
 					ancestors: []dag.ID{"C", "D"},
 				},
+				"C": {},
```
snakster, line 225: This was a bug in the test case. The old algorithm coincidentally supported it even even you referenced nodes that were never added with `AddNode`, but this has no use case and just complicates everything.

run/dag/dag_test.go
```diff
@@ -404,6 +405,9 @@ func TestTransformDAG(t *testing.T) {
 		))
 	}
 
+	_, err := a.Validate()
```
snakster, line 408: We always need the validate now.
1. Add nodes
2. Validate (which computes order and detects cycles)
3. Order, Reduce, etc

Before validate was optional.

run/dag/dag.go
```diff
@@ -98,26 +102,26 @@ func (d *DAG[V]) addAncestor(node, ancestor ID) {
 
 // Validate the DAG looking for cycles.
 func (d *DAG[V]) Validate() (reason string, err error) {
-	d.cycles = make(map[ID]bool)
-	d.validated = true
-
-	for _, id := range d.IDs() {
+	d.calcTopologicalOrder()
+	for _, id := range slices.Sorted(maps.Keys(d.cycles)) {
```
snakster, line 106: `validateNode` will still be non-optimized, but now we only call it on nodes where we already know that we have a cycle, and it stops once we found one.

The `calcTopologicalOrder` marks cycles, but doesn't calculate the paths.

snakster (commented):
🦆

run/dag/dag.go
```diff
@@ -227,33 +232,82 @@ func (d *DAG[V]) HasCycle(id ID) bool {
...
+	if !d.validated {
+		panic(errors.E(errors.ErrInternal, "please report this as a bug"))
+	}
+	return d.order
+}
+
+// calcTopologicalOrder does topological ordering based on Kahns algorithm
+func (d *DAG[V]) calcTopologicalOrder() {
+	// Helper maps to lookup predecessors and successors of a node.
+	// Roots of the grph are nodes with no predecessors.
```
sixstone-qq:
```suggestion
	// Roots of the graph are nodes with no predecessors.
```

run/dag/dag.go
```diff
@@ -227,33 +232,82 @@ func (d *DAG[V]) HasCycle(id ID) bool {
...
+	}
+
+	// Will contain the topological order.
+	order := []ID{}
 
-			visited[id] = struct{}{}
-		})
+	// Temporary set that contains nodes without predecessors for the algorithm.
+	// Will be expanded as we remove edges.
+	s := []ID{}
```
sixstone-qq: As I read, it may be better to rename the variable to queue for better readability?
```suggestion
	queue := []ID{}
```

run/dag/dag.go
```diff
@@ -227,33 +232,82 @@ func (d *DAG[V]) HasCycle(id ID) bool {
...
+			delete(predMap[succ], n)
+			delete(succMap[n], succ)
+
+			// If succ is now without predecessors, it will be added to s next.
+			if len(predMap[succ]) == 0 {
+				next = append(next, succ)
+			}
+		}
+
+		// Add in a deterministic manner.
```
sixstone-qq:
```suggestion
		// Add using lexicographical order to be deterministic.
```

run/dag/dag.go
sixstone-qq: Could we do this in the previous `for` loop? Or does not help this optimisation to read it better.

sixstone-qq (commented):
Just few comments that may improve the PR.

Great work! :people_hugging:

run/dag/dag.go
snakster: Fixed in [e623fed](https://github.com/terramate-io/terramate/pull/2165/commits/e623fed5c4b669bf1da740756ede97c6f733374a)

run/dag/dag.go
snakster: Agreed, done in [e623fed](https://github.com/terramate-io/terramate/pull/2165/commits/e623fed5c4b669bf1da740756ede97c6f733374a).

run/dag/dag.go
snakster: Done in [e623fed](https://github.com/terramate-io/terramate/pull/2165/commits/e623fed5c4b669bf1da740756ede97c6f733374a)

run/dag/dag.go
snakster: You are right, it could be refactored a bit, but I also think it's ok like this. Added an extra comment in [e623fed](https://github.com/terramate-io/terramate/pull/2165/commits/e623fed5c4b669bf1da740756ede97c6f733374a)

sixstone-qq (approved):

PR #2173, opened 2025-07-24
fix: module change detection for common name prefix

snakster:

This PR fixes a bug in the change detection that would mistakenly list a stack as changed in some circumstances.

The scenario is that if a stack named `foobar` references a Terraform module, then changes in that module will also flag a stack with a common name prefix as changed, i.e. a stack named `foo`.

```
No, it is a change in behaviour but does not break compatibility.
```

### DIFF
--- a/e2etests/core/change_detection_test.go
+++ b/e2etests/core/change_detection_test.go
@@ -626,3 +626,69 @@ func TestTriggerChangeDetection(t *testing.T) {
 		})
 	}
 }
+
+func TestChangeDetectionCommonPrefix(t *testing.T) {
+	// Test regressions for previous bug in change detection.
+
+	t.Run("tf modules with common prefix", func(t *testing.T) {
+		s := sandbox.New(t)
+		layout := []string{
+			"s:stack-1",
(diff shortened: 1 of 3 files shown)


soerenmartius (dismissed):
lgtm

e2etests/core/change_detection_test.go
```diff
@@ -626,3 +626,69 @@ func TestTriggerChangeDetection(t *testing.T) {
 		})
 	}
 }
+
+func TestChangeDetection_SC15523(t *testing.T) {
+	// Test regression for bug described in SC-15523.
```
sixstone-qq: I think we should avoid internal references in public repos

snakster: I wasn't sure about it myself, but I saw it was done before so I kept it.

Removed in [69bd741](https://github.com/terramate-io/terramate/pull/2173/commits/69bd741f5d3e99de3dba28c5955fd7117897b4f6)

sixstone-qq (approved):
LGTM

PR #2189, opened 2025-08-22
feat: Sync CI/CD server to TMC for GitHub and Gitlab platforms

sixstone-qq:

This PR synchronises the server host for Gitlab CI/CD and the server URL for GitHub actions to
Terramate Cloud to give better support to self hosted platforms.

```
no
```

### DIFF
--- a/cloudsync/metadata.go
+++ b/cloudsync/metadata.go
@@ -839,6 +839,7 @@ func setGithubActionsMetadata(md *resources.DeploymentMetadata) {
 	md.GithubActionsRunAttempt = os.Getenv("GITHUB_RUN_ATTEMPT")
 	md.GithubActionsWorkflowName = os.Getenv("GITHUB_WORKFLOW")
 	md.GithubActionsWorkflowRef = os.Getenv("GITHUB_WORKFLOW_REF")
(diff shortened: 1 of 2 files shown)


cloudsync/metadata.go
```diff
@@ -499,6 +499,7 @@ func setGitlabCIMetadata(_ *engine.Engine, md *resources.DeploymentMetadata, sta
 	md.GitlabCICDUserLogin = os.Getenv("GITLAB_USER_LOGIN")
 	md.GitlabCICDCommitBranch = os.Getenv("CI_COMMIT_BRANCH")
 	md.GitlabCICDBranch = md.GitlabCICDCommitBranch
+	md.GitlabCIServerHost = os.Getenv("CI_SERVER_HOST")
```
snakster, line 502: Checking the docs, it says
```
The host of the GitLab instance URL, without protocol or port.
```
Is this sufficient for what we use it for in the backend? There is also `CI_SERVER_FQDN` etc.

sixstone-qq: Yes, that's enough but I'm adding `CI_SERVER_URL` for the sake of completeness

sixstone-qq: Done in 14099827

PR #2191, opened 2025-09-01
feat: add tm_slug utility function

snakster:

This PR adds a convenience function `tm_slug(t)` that is equivalent to `replace(lower(t), "/[^a-z0-9_]/", "-")`.
It can be used to sanitze characters that are not valid in identifiers.

```
yes, it adds a new terraform function.
```


sixstone-qq (approved):
LGTM

PR #2193, opened 2025-09-02
feat: add tm_tree function to build tree branches from parent-child pairs

mcalhoun:

This PR introduces a new built-in Terramate function `tm_tree()` that converts an unordered list of (parent, child) pairs into a list of branch paths. This function enables users to derive hierarchical paths from relational data for use in code generation.

Key features:

- Converts parent-child pairs into full hierarchical paths
- Supports multiple roots (forest structures)
- Returns deterministic, lexicographically sorted output
- Handles unordered input gracefully
- Comprehensive error handling with custom error types

Implementation details:

- Added `tm_tree()` function in `stdlib/funcs_tree.go`
- Created custom error types in `stdlib/funcs_tree_errors.go` for better error reporting
- Comprehensive test suite with 100% coverage of all error cases
- Registered function in the standard library

Fixes # (if applicable)

The function implements all requirements from the PRD including:
  - Detection and reporting of cycles in the tree structure
  - Detection of conflicting parents (same child with multiple parents)
  - Validation of unknown parent references
  - Proper handling of null values and empty strings
  - All error conditions return descriptive, actionable error messages

The implementation uses custom error types to provide clear feedback when input data is malformed or contains logical errors.

Yes - adds new `tm_tree()` function to Terramate's standard library.

#  Example usage:

```hcl
locals {
  pairs = [
    [null, "root"],
    ["root", "child1"],
    ["root", "child2"],
    ["child1", "grandchild"]
  ]
  branches = tm_tree(local.pairs)
  # Returns: [["root"], ["root", "child1"], ["root", "child1", "grandchild"], ["root", "child2"]]
}
```

The function signature is:

`tm_tree(pairs: list(tuple([string, string]))) -> list(list(string))`

Where:

- Input: List of [parent, child] tuples (parent can be null to indicate a root)
- Output: List of paths, where each path is a list of strings from root to node

## Testing:

 - All unit tests pass: go test ./stdlib -run TestTmTree
 - Full stdlib test suite passes: go test ./stdlib
 - Comprehensive error coverage with tests for all error conditions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 - Backward compatibility in versions `0.0.z` is **not guaranteed** when `z` is increased.
 - Backward compatibility in versions `0.y.z` is **not guaranteed** when `y` is increased.
 
+## Unreleased
+
+### Added
+
+- Add `tm_tree()` function for constructing hierarchical paths from parent-child relationships.
+
+```hcl
+  tm_tree([[null, "root"], ["root", "child1"], ["root", "child2"], ["child1", "leaf"]])
+  # Returns: [["root"], ["root", "child1"], ["root", "child1", "leaf"], ["root", "child2"]]
+```
+
+- Add support for slashes (`/`) in tags.
+
 ## v0.14.4
 
 ### Fixed
@@ -60,54 +73,58 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
   This affects and improves commands that use the stack order, i.e. `run`, `run script`, `list --run-order`.
   As a consequence, the evaluation order of _unrelated_ stacks may change.
-
   - Example 1:
     Nested stacks `a` must be executed after their parent stacks, but `/stack1` and `/stack2` are independent.
 
     Old run order:
-      ```
-      /stack1
-      /stack1/a
-      /stack2
-      /stack2/a
-      ```
+
+    ```
+    /stack1
+    /stack1/a
+    /stack2
+    /stack2/a
+    ```
 
     New run order:
-      ```
-      /stack1
-      /stack2
-      /stack1/a
-      /stack2/a
-      ```
+
+    ```
+    /stack1
+    /stack2
+    /stack1/a
+    /stack2/a
+    ```
 
   - Example 2:
     `stack2` must be executed after `stack1`, but `stack1` and `stack3` are independent in the follwing configuration:
-      ```
-      /stack1 
-      /stack2 (after=[stack1])
-      /stack3
-      ```
+
+    ```
+    /stack1
+    /stack2 (after=[stack1])
+    /stack3
+    ```
 
     Old run order:
-      ```
-      /stack1
-      /stack2
-      /stack3
-      ```
+
+    ```
+    /stack1
+    /stack2
+    /stack3
+    ```
 
     New run order:
-      ```
-      /stack1
-      /stack3
-      /stack2
-      ```
+
+    ```
+    /stack1
+    /stack3
+    /stack2
+    ```
 
   The new rule recursively aligns stacks that are independent into groups.
   - The first group contains all initially independent stacks.
   - The second group contains all stacks that just depended on stacks in the first group.
   - The third group contains all stacks that just depended on stacks in both previous groups.
   - The fourth group and following groups continue in the same way.
-  
+
   Stacks in each group are ordered lexicographically and returned as the order of execution when running sequentially.
 
   The run order when using the `--parallel` flag is not affected by this change.
@@ -230,7 +247,7 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 - Add support for tracking `file()` usages in Terragrunt files for enhancing the change detection.
   - Now if you have Terragrunt modules that directly read files from elsewhere in the project, Terramate will
-  mark the stack changed whenever the aforementioned file changes.
+    mark the stack changed whenever the aforementioned file changes.
 - Add telemetry to collect anonymous usage metrics.
   - This helps us to improve user experience by measuring which Terramate features are used most actively.
     For further details, see [documentation](https://terramate.io/docs/cli/telemetry).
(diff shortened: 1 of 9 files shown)


stdlib/funcs_tree_test.go
```diff
@@ -0,0 +1,499 @@
...
+			errorMsg: []string{"multiple parents", "A"}, // This creates conflicting parents
+		},
+		{
+			name:     "self-reference cycle",
+			input:    `tm_tree([["A", "A"]])`,
+			errorMsg: []string{"cycle detected", "A"},
+		},
+		{
+			name:     "cycle detection with proper tree",
+			input:    `tm_tree([[null, "root"], ["root", "A"], ["A", "B"], ["B", "A"]])`,
```
snakster: This looks redundant as it is the same input as `simple two-node cycle` above.

stdlib/funcs_tree_errors.go
```diff
@@ -0,0 +1,102 @@
...
+
+package stdlib
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TreeError represents errors that can occur in the tm_tree function
+type TreeError struct {
```
snakster: In my opinion we don't need the special error type here, unless you find it convenient for testing. But in application logic we will not inspect it.

stdlib/funcs_tree.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// Sort children for deterministic output
+	for _, node := range nodes {
+		sort.Slice(node.children, func(i, j int) bool {
+			return node.children[i].name < node.children[j].name
+		})
+	}
+
+	// Detect cycles using DFS
```
snakster, line 152: There is a `dag` package we use to model the stack dependency relations, it could be an option to use this here. The benefit would be to reuse its cycle detection, for all the rest you would still to prepare the data structures yourself.

snakster (commented):
Looks good, nice work 👍 
Left some minor comments.

mariux:
i tested it in action and it works great for my use case ;) thx a lot :rocket:

stdlib/funcs_tree_test.go
mcalhoun: Good catch @snakster! 👀 You're absolutely right about the duplicate test cases on lines 402 and 412.

I removed the duplicate "cycle detection with proper tree" test case (lines 412-416) since it was testing the exact same input and expected error as the "simple two-node cycle" test. The first one has a cleaner name and is less verbose, so I kept that one.

stdlib/funcs_tree_errors.go
mcalhoun: Good question about the custom error types! 🤔

You're right to question whether they're just for testing - that was actually a key design decision I wrestled with. Here's why I think they're worth it beyond just tests:

**User tooling integration:** These enable programmatic error handling. Someone building a Terramate wrapper or IDE integration can catch specific error types and provide contextual help. For example:

  ```go
  var unknownParentErr TreeErrorUnknownParent
  if errors.As(err, &unknownParentErr) {
      // Suggest available parents or highlight the problematic reference
  }

  Better developer experience: Instead of parsing error strings, users get structured information. The TreeErrorConflictingParents type tells you exactly which child has conflicts and what the conflicting
  parents are.

  Go best practices: This follows the error handling patterns from Go 1.13+ where the standard library moved toward typed errors with errors.As() for APIs.

  Future extensibility: If we need to add suggestions or additional context to errors later, we can extend these structs without breaking existing code.

  But I'm definitely open to discussion if the team thinks the extra complexity isn't worth it! What's your take?

stdlib/funcs_tree.go
mcalhoun: Interesting idea about using the existing dag package!

In my quick assessment, I don't think we need it because:

Different abstraction level: The dag package seems designed for more general graph operations, while tm_tree has very specific requirements (parent-child pairs, null handling, string-specific validation, etc.).

Error message specificity: We need very precise error messages like "unknown parent 'foo' referenced by child 'bar'" - the dag package might give us cycle detection but not the Terramate-specific error context users need.

Input format: We're working with [parent, child] tuples directly from HCL, not a general graph structure.

Lightweight approach: The current implementation is pretty minimal - just basic parent/child tracking and validation. Adding a dependency might be overkill?

That said, if there's specific functionality from the dag package that would simplify the implementation, I'm definitely interested! Did you have particular functions in mind? Always happy to reduce reinventing wheels where it makes sense.

stdlib/funcs_tree_errors.go
snakster:
```
var unknownParentErr *TreeErrorUnknownParent
if errors.As(err, &unknownParentErr) {
    // Suggest available parents or highlight the problematic reference
}
```
But there is no `TreeErrorUnknownParent`, it's just `TreeError`. You'd need a specific error type for each of the subtypes if you want to be able to extract this information on a different layer.

The error messages themselves are great, but for the `TreeError` type as it is here I see no point beyond testing.

stdlib/funcs_tree.go
snakster: I think it would just be the part about the cycle detection, as you said, anything else like checking the specific constraints for the input data etc you still have to do yourself. So I agree it's good the way it is.

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 - Backward compatibility in versions `0.0.z` is **not guaranteed** when `z` is increased.
 - Backward compatibility in versions `0.y.z` is **not guaranteed** when `y` is increased.
 
+## Unreleased
+
+### Added
+
+- Add `tm_joinlist()` function for joining nested lists of strings with a separator.
```
snakster: Mixed up the changelog entries :)

stdlib/funcs_tree_errors.go
snakster: For just tagging an error with a sub-type, we also have the generic errors.E(...) already. Usage example:

```
import "github.com/terramate-io/terramate/errors"
...
TreeErrorUnknownParent errors.Kind = "unknown tree parent"
...
func TreeErrorUnknownParentValue(parent, child string) error {
	return errors.E(TreeErrorUnknownParent,
		"unknown parent %q for child %q. Declare [null, %q] or provide a chain to a root",
		parent, child, parent)
}
...
assert.IsTrue(t, errors.IsKind(err1, stdlib.TreeErrorUnknownParent))
```

mcalhoun: Ok, @snakster, I've updated the error patterns to your suggestion and fixed the changelog.

snakster (dismissed):
LGTM 👍 Can be merged after rebase/merge from main.

PR #2203, opened 2025-09-15
refactor: add validate step for merged blocks parsing

snakster:

This PR refactors the interface for merged block handlers (with and without labels) to include an additional `Validate` function that will be called after parsing. This allows handlers to validate postconditions that can only be checked after parsing (i.e. to support incremental definitions).

This is not used by anything within Terramate itself, but will be required by external code.

```
no
```


sixstone-qq (approved):
LGTM

PR #2206, opened 2025-09-17
fix: prevent panic when tm_slug(null) is tokenized

mcalhoun:

- Fixes panic when evaluating `tm_slug(null)` by ensuring a concrete, printable output type instead of leaking `cty.DynamicPseudoType`.
- Updates `tm_slug` typing and implementation to handle dynamic/null/unknown inputs:
    - In `stdlib/funcs_slug.go`:
        - Type: if param type is `cty.DynamicPseudoType`, default return type to `cty.String`.
        - Impl: map dynamic null → `cty.NullVal(cty.String)`; dynamic unknown → `cty.UnknownVal(cty.String)`; keep tuple null/unknown mapping to `list(string)`.
- Adds tests covering null/dynamic/unknown behavior and tokenization safety in `stdlib/funcs_slug_test.go`:
    - `TestTmSlugDynamicNull` (via eval): `tm_slug(null)` → typed null string; tokenization does not panic.
    - `TestTmSlugDynamicUnknown`: unknown dynamic (`cty.DynamicVal`) → unknown string.
    - `TestTmSlugNullStringDirect`: `cty.NullVal(cty.String)` → typed null string; tokenizable.
    - `TestTmSlugNullListDirect`: `cty.NullVal(cty.List(cty.String))` → typed null list(string); tokenizable.
- Result: `hcl/ast.TokensForValue` no longer hits the unsupported-type panic path for `tm_slug(null)`.

Fixes #sc-15798

Chose `string` as the default output type for bare dynamic/null to keep behavior sensible and formatting stable without altering the printer.

Yes:

`tm_slug(null)` now returns a typed null string instead of potentially causing a panic during formatting. Unknown dynamic inputs to `tm_slug` now propagate as unknown string values.

---

> [!NOTE]
> Prevents panics when tokenizing tm_slug(null) by rendering unknown/ dynamic values as null and simplifying tm_slug handling, with tests adjusted accordingly.
> 
> - **HCL AST**:
>   - Treat unknown/partially-known `cty` values as `null` in `hcl/ast/value.go` tokenization to avoid panics.
> - **Stdlib: tm_slug**:
>   - Simplify `tmSlug` by removing explicit null/unknown propagation and related error kind; rely on HCL typing and existing string/list/tuple handling.
> - **Tests**:
>   - Add `generate/genhcl/TestGenerateHCLTmSlugNull` to assert `tm_slug(null)` evaluates to `null` in generated code.
>   - Add tokenization safety tests: `TestTmSlugDynamicNull`, `TestTmSlugDynamicUnknown` ensuring no panic when tokenizing dynamic results.
>   - Remove unknown-propagation and related error tests; retain list/tuple element-type validation tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e339669e296f8595c0f0f1b144bd34413c075da9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/generate/genhcl/genhcl_test.go
+++ b/generate/genhcl/genhcl_test.go
@@ -1958,6 +1958,45 @@ func assertHCLEquals(t *testing.T, got string, want string) {
 	}
 }
 
+func TestGenerateHCLTmSlugNull(t *testing.T) {
+	t.Parallel()
+
+	tcases := []testcase{
+		{
+			name:  "tm_slug_null_evaluates_to_null",
+			stack: "/stack",
+			configs: []hclconfig{
+				{
+					path: "/stack",
+					add: GenerateHCL(
+						Labels("scope_traversal"),
+						Content(
+							Block("traversals",
+								Expr("local", "tm_slug(null)"),
+							),
+						),
+					),
+				},
+			},
+			want: []result{
+				{
+					name: "scope_traversal",
+					hcl: genHCL{
+						condition: true,
+						body: Block("traversals",
+							Expr("local", "null"),
+						),
+					},
+				},
+			},
+		},
+	}
+
+	for _, tcase := range tcases {
+		tcase.run(t)
+	}
+}
+
 func init() {
 	zerolog.SetGlobalLevel(zerolog.Disabled)
 }
--- a/hcl/ast/value.go
+++ b/hcl/ast/value.go
@@ -31,6 +31,11 @@ func (builder *tokenBuilder) fromValue(val cty.Value) {
 		builder.fromExpr(customdecode.ExpressionFromVal(val))
 	case val.IsNull():
 		builder.add(ident("null", 0))
+	case !val.IsWhollyKnown():
+		// Handle unknown values (including dynamic pseudo-type unknowns)
+		// These represent values that couldn't be resolved during evaluation
+		// Render as null to avoid panics when trying to extract concrete values
+		builder.add(ident("null", 0))
 	case typ == cty.Bool:
 		if val.True() {
 			builder.add(ident("true", 0))
--- a/stdlib/funcs_slug.go
+++ b/stdlib/funcs_slug.go
@@ -49,28 +49,6 @@ func SlugFunc() function.Function {
 func tmSlug(arg cty.Value) (cty.Value, error) {
 	argType := arg.Type()
 
-	if arg.IsNull() {
-		// For tuples, return null List(String) to match Type function declaration
-		if argType.IsTupleType() {
-			return cty.NullVal(cty.List(cty.String)), nil
-		}
-		return cty.NullVal(argType), nil
-	}
-
-	if !arg.IsWhollyKnown() {
-		// Propagate unknown values with correct output type
-		switch {
-		case argType.Equals(cty.String):
-			return cty.UnknownVal(cty.String), nil
-		case argType.IsListType() && argType.ElementType().Equals(cty.String):
-			return cty.UnknownVal(argType), nil
-		case argType.IsTupleType():
-			return cty.UnknownVal(cty.List(cty.String)), nil
-		default:
-			return cty.DynamicVal, errUnknownValue("tm_slug")
-		}
-	}
-
 	switch {
 	case argType.Equals(cty.String):
 		return cty.StringVal(slugify(arg.AsString())), nil
--- a/stdlib/funcs_slug_errors.go
+++ b/stdlib/funcs_slug_errors.go
@@ -14,9 +14,6 @@ const SlugWrongType errors.Kind = "tm_slug: wrong type"
 // SlugListElementNotString is returned when a list element is not a string.
 const SlugListElementNotString errors.Kind = "tm_slug: list element not string"
 
-// SlugUnknownValue is returned when the value is not wholly known.
-const SlugUnknownValue errors.Kind = "tm_slug: unknown value"
-
 // errWrongRootType returns an error for invalid root types
 func errWrongRootType(_ string, t cty.Type) error {
 	return errors.E(SlugWrongType,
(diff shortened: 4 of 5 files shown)


snakster:
It still seems to panic, try it with this test case in `generate/genhcl/genhcl_test.go`:

```
func TestGenerateHCLPanic(t *testing.T) {
	t.Parallel()

	tcases := []testcase{
		{
			name:  "panic",
			stack: "/stack",
			configs: []hclconfig{
				{
					path: "/stack",
					add: GenerateHCL(
						Labels("scope_traversal"),
						Content(
							Block("traversals",
								Expr("local", "tm_slug(null)"),
							),
						),
					),
				},
			},
			want: []result{
				{
					name: "scope_traversal",
					hcl: genHCL{
						condition: true,
						body: Block("traversals",
							Expr("local", "tm_slug(null)"),
						),
					},
				},
			},
		},
	}

	for _, tcase := range tcases {
		tcase.run(t)
	}
}
```

mcalhoun:
* **#2206** <a href="https://app.graphite.dev/github/pr/terramate-io/terramate/2206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terramate-io/terramate/2206?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2210** <a href="https://app.graphite.dev/github/pr/terramate-io/terramate/2210?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

generate/genhcl/genhcl_test.go
```diff
@@ -1958,6 +1958,45 @@ func assertHCLEquals(t *testing.T, got string, want string) {
 	}
 }
 
+func TestGenerateHCLPanic(t *testing.T) {
+	t.Parallel()
+
+	tcases := []testcase{
+		{
+			name:  "panic",
```
snakster: Could you give this test a more descriptive name? (both function and test case name)

For later (beyond the scope of this PR) we could also consider to add some more tests that use the functions in `generate_hcl` context.

hcl/ast/value.go
```diff
@@ -31,6 +31,15 @@ func (builder *tokenBuilder) fromValue(val cty.Value) {
 		builder.fromExpr(customdecode.ExpressionFromVal(val))
 	case val.IsNull():
 		builder.add(ident("null", 0))
+	case !val.IsWhollyKnown():
```
snakster, line 34: My only concern here would be that this could silently render values as null that are not meant to be, but I was not able to produce any examples, so I guess it's fine :)

BTW I experimented with an alternative similar to `tm_ternary`, which also looked promising, that is using
```
{
				Name: "input",
				Type: customdecode.ExpressionClosureType,
},
```

then doing what looks to me like a way of deferred evaluation in the function `Impl`
```
closure := customdecode.ExpressionClosureFromVal(arg)
realArg, diags := newexpr.Value(closure.EvalContext)
if diags.HasErrors() {
	...
}
```
This will actually invoke the handler with a `null` as an argument.

Something to keep in mind if we discover issues with the current approach.

stdlib/funcs_slug.go
```diff
@@ -28,6 +28,10 @@ func SlugFunc() function.Function {
 		Type: func(args []cty.Value) (cty.Type, error) {
 			argType := args[0].Type()
 			switch {
+			case argType == cty.DynamicPseudoType:
+				// Bare nulls and unconstrained dynamics default to string output
```
snakster: Is this  comment still true? When `AllowNull` is set, it seems to me this `Type` function is not invoked at all. Maybe this case can just be removed?

snakster:
There is a typo in the commit message: `-bux+fix`

hcl/ast/value.go
mcalhoun: I think the key is that we are only adding the rendering of things as null to values whose values can't be determined (and aren't already concrete nulls). If we don't do this, the same case statement will panic a few lines down.

stdlib/funcs_slug.go
mcalhoun: The `Type` function is still called even when `AllowNull: true`. The `AllowNull: true` setting allows null values to be passed to both the `Type` and `Impl` functions, but the `Type` function still needs to determine the return type.

When you call `tm_slug(null)` in HCL, it creates a `cty.DynamicPseudoType` value that gets passed to the `Type` function. Without this case, it would fall through to the default case and return `errWrongRootType("tm_slug", argType)`.

The `TestTmSlugDynamicNull` test demonstrates this behavior - the dynamic case is necessary to provide a concrete return type (`cty.String`) for `tm_slug(null)` instead of erroring out.

snakster: I tested it by setting some breakpoints in the debugger (or just immediately panic for both functions), they are not being called for the `generate_hcl` test case when `AllowNull` is true.

Edit: Same for `TestTmSlugDynamicNull`, the test always passes for me, never invoking the `Type` function.

snakster (approved):

mcalhoun:
### Merge activity

* **Oct 2, 8:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/terramate-io/terramate/2206).
* **Oct 2, 8:43 PM UTC**: @mcalhoun merged this pull request with [Graphite](https://app.graphite.dev/github/pr/terramate-io/terramate/2206).

PR #2209, opened 2025-09-23
fix: status and tag filtering for trigger

snakster:

This PR fixes unsupported filter flags for the trigger command, namely `--drift-status` and `--deployment-status`.

To enable this, the `FilterStacks` function has been refactored to support composition of predicate functions. Before, it did not support filtering for tags and stack path at the same time.

The tag filtering can either by used with cloud status filters, or with --recursive on local stacks. 

```
yes
```


mcalhoun (approved):
This looks good to me!

sixstone-qq (approved):

PR #2220, opened 2025-10-06
chore: fix flaky tests

snakster:

Fixes #

```

```


e2etests/internal/runner/runner.go
```diff
@@ -358,14 +358,14 @@ func (tm CLI) RunInteractive(ioFunc InteractiveIO, args ...string) RunResult {
 	outFunc := func(r io.Reader) ExpectedRead {
 		return func(want string) {
 			t.Helper()
-			// We try to read more than len(want), assuming its <1024.
-			// So if there is more than the expected output, we will fail.
-			buf := make([]byte, 1024)
-			n, err := r.Read(buf)
+
+			// Try to read in len(want) bytes. Will block while waiting for more output.
+			buf := make([]byte, len(want))
```
mcalhoun, line 363: Consider verifying that there isn't a deadlock if the len < want (legitimiately)

e2etests/internal/runner/runner.go
```diff
@@ -358,14 +358,14 @@ func (tm CLI) RunInteractive(ioFunc InteractiveIO, args ...string) RunResult {
...
+			n, err := io.ReadFull(r, buf)
```
mcalhoun, line 364: Do we need some kind of timeout here?

mcalhoun (approved):
Overall I think it looks good. You can free to merge or address my suggestions.

e2etests/internal/runner/runner.go
snakster: I agree it would be nice to have. I spent some time on this and in my first attempt I implemented it with polling a buffer and a context for timeout, but it makes the code more complicated. So for now, I settled for this approach, even though it has some limitations.

e2etests/internal/runner/runner.go
snakster: In this case it would block until the test timeout is hit. That is not great, but testing for interactive input is a rare use case, so I figured it is good enough for now.

PR #2228, opened 2025-10-08
feat: add cache for terragrunt files

mcalhoun:
# feat: add shared cache for Terragrunt module parsing

Implements a shared cache for Terragrunt module parsing to improve performance when loading modules in parallel. The cache stores parsed Terragrunt configuration files to avoid re-parsing the same files multiple times, which is particularly beneficial when multiple modules include the same parent configurations.

Key changes:

- Added a thread-safe `TgParseCache` to store parsed Terragrunt configurations
- Modified `LoadModule` to accept an optional shared cache parameter
- Updated the root config to create a single shared cache for all Terragrunt workers
- Added benchmarks demonstrating significant performance improvements with large-scale Terragrunt repositories

The benchmarks show substantial performance gains, especially in repositories with many modules that include common configuration files (root terragrunt.hcl, account.hcl, region.hcl, etc.).

Fixes #

The cache is particularly effective in large-scale Terragrunt repositories where many modules include the same parent configurations. The benchmarks include a realistic large-scale test with 800 modules across multiple accounts, regions, and environments.

No user-facing changes, but users should see improved performance when working with large Terragrunt repositories.

## Performance Benchmarks

This PR adds a parse cache for Terragrunt configuration files to eliminate redundant parsing. The optimization provides **measurable improvements that scale with repository size**.

#### Quick Comparison (Small-Scale: 6 modules)

```
go test ./tg -bench=BenchmarkModuleDiscovery -benchmem
```

#### Stable Comparison (Small-Scale with Multiple Runs)

```
go test ./tg -bench=BenchmarkModuleDiscovery -benchtime=50x -benchmem -count=3
```

#### Large-Scale Validation (800 modules)

```
go test ./tg -bench=BenchmarkLargeScale -benchmem -benchtime=5x -timeout=30m
```

#### All Benchmarks

```
go test ./tg -bench=. -benchmem
```

---

#### Small-Scale (6 modules - realistic infrastructure-live example)

| Metric | With Cache | Without Cache | Improvement |
| --- | --- | --- | --- |
| Time | 37\.7ms | 37\.9ms | 0\.5% faster |
| Memory | 62\.2 MB | 64\.7 MB | **4\.0% less** |
| Allocations | 367\.8K | 385\.6K | **4\.8% fewer** |

#### Large-Scale (800 modules, 939 files)

**Setup:** 5 accounts × 5 regions × 4 environments × 8 services

| Metric | With Cache | Without Cache | **Improvement** |
| --- | --- | --- | --- |
| **Time** | 3\.98s | 4\.37s | **🚀 9.0% faster (395ms saved)** |
| **Memory** | 6\.73 GB | 7\.47 GB | **💾 9.9% less (736 MB saved)** |
| **Allocations** | 43\.0M | 47\.9M | **📉 10.3% fewer (4.9M saved)** |

---

### What This Means

In typical Terragrunt repositories, modules share common configuration files:

- Root `terragrunt.hcl` (included by all modules)
- Account, region, environment configs (shared hierarchically)
- Template files in `_envcommon/` (shared by service type)

**Before:** Each of the 800 modules independently parsed all shared files

- Root config parsed **800 times** ❌
- Shared files parsed **hundreds of times each** ❌

**After:** Shared files parsed once and cached

- Root config parsed **1 time** ✅
- Shared files cached on first access ✅
- **Scales with repo size** - more modules = bigger benefit ✅

---

### Testing

All tests pass including new cache-specific tests:

```
# All tests (includes 23 existing + 4 new cache tests)
go test ./tg/... -v

# Race detector (verifies thread safety)
go test -race ./tg -run TestCacheConcurrency -v

# Config integration tests
go test ./config -v
```

**Race detector:** ✅ No race conditions detected  
**Existing tests:** ✅ All 23 test cases pass  
**New tests:** ✅ 4 cache-specific tests added  
**Benchmarks:** ✅ 4 benchmarks (2 scales × 2 variants)

---

> [!NOTE]
> Introduce a thread-safe shared cache for Terragrunt config parsing, integrate it into loaders/workers/scan, and add a CLI flag/env to disable; includes tests and benchmarks.
> 
> - **Terragrunt Core (tg)**:
>   - **Caching**: Add `ParseCache` (thread-safe) with `NewParseCache()`, default shared cache, and env-controlled disable via `TM_DISABLE_TG_CACHE`.
>   - **API**: Introduce `LoadModuleWithCache(...)`; make `LoadModule(...)` delegate to it; `Module` now holds `parseCache`.
>   - **Functions**: `readTerragruntConfigImpl(...)` caches parsed values and replays transitive `DependsOn`; dedupes deps and canonicalizes cache keys.
>   - **Scanning**: `ScanModules(...)` uses a shared cache across files.
> - **Config Integration**:
>   - Terragrunt workers initialize a shared cache (unless disabled) and use `LoadModuleWithCache(...)`.
> - **CLI**:
>   - Add `--disable-tg-cache` flag (`TM_DISABLE_TG_CACHE`) and set env in handler to disable caching.
> - **Tests/Benchmarks**:
>   - Add cache API, concurrency, transitive-deps, and env-disable tests.
>   - Add small and large-scale benchmarks comparing with/without cache.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5aa877bc00685697932eee3deafcd5aacde0b152. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -52,6 +52,7 @@ type Root struct {
 
 	maxTgWorkers int
 	tgTaskChan   chan *Tree
+	tgParseCache *tg.ParseCache
 
 	// The mutex and files below are needed until we figure a better way
 	// of discovering Terragrunt modules.
@@ -174,6 +175,13 @@ func NewRoot(tree *Tree, hclOpts ...hcl.Option) *Root {
 }
 
 func (root *Root) initTgWorkers() {
+	// Initialize cache unless disabled via --disable-tg-cache flag or TM_DISABLE_TG_CACHE env var.
+	// A nil cache disables caching, which may be needed for configs with conditional
+	// file loading based on per-module context.
+	if os.Getenv("TM_DISABLE_TG_CACHE") == "" {
+		root.tgParseCache = tg.NewParseCache()
+	}
+
 	if !root.changeDetectionEnabled {
 		return
 	}
@@ -192,7 +200,10 @@ func (root *Root) tgWorker() {
 		root.tgTransientErrs[tgFile] = errors.L()
 		root.tgmu.Unlock()
 
-		tgMod, isRootModule, err := tg.LoadModule(root.HostDir(), tree.Dir(), tree.TgRootFile, trackTerragruntDependencies)
+		// Use the shared cache across all workers to maximize cache hit rates.
+		// When modules are loaded in parallel, they can benefit from each other's
+		// parsing work on shared files (root terragrunt.hcl, account.hcl, etc).
+		tgMod, isRootModule, err := tg.LoadModuleWithCache(root.HostDir(), tree.Dir(), tree.TgRootFile, trackTerragruntDependencies, root.tgParseCache)
 		root.tgmu.Lock()
 		if err != nil {
 			root.tgTransientErrs[tgFile] = err
--- a/tg/funcs.go
+++ b/tg/funcs.go
@@ -159,6 +159,9 @@ func tgReadTerragruntConfigFuncImpl(ctx *tgconfig.ParsingContext, tgLogger log.L
 // Check the original version here: https://github.com/gruntwork-io/terragrunt/blob/b47b57ae0cd2c8644ca5625fceed0a2258b1a763/config/config_helpers.go#L578-L612
 // The important changes are:
 //   - The read file is added to the `mod.DependsOn` slice.
+//   - Results (value + transitive dependencies) are cached to avoid re-parsing the same file multiple times.
+//   - Caching can be disabled via --disable-tg-cache flag or TM_DISABLE_TG_CACHE env var if your
+//     configs use conditional file loading based on per-module context (rare but possible).
 func readTerragruntConfigImpl(ctx *tgconfig.ParsingContext, tgLogger log.Logger, configPath string, defaultVal *cty.Value, rootdir string, mod *Module) (cty.Value, error) {
 	targetConfig := getCleanedTargetConfigPath(configPath, ctx.TerragruntOptions.TerragruntConfigPath)
 	targetConfigFileExists := util.FileExists(targetConfig)
@@ -168,7 +171,52 @@ func readTerragruntConfigImpl(ctx *tgconfig.ParsingContext, tgLogger log.Logger,
 		return *defaultVal, nil
 	}
 
-	mod.DependsOn = append(mod.DependsOn, project.PrjAbsPath(rootdir, targetConfig))
+	// Track dependency - check if already present to avoid duplicates
+	depPath := project.PrjAbsPath(rootdir, targetConfig)
+	alreadyTracked := false
+	for _, existing := range mod.DependsOn {
+		if existing == depPath {
+			alreadyTracked = true
+			break
+		}
+	}
+	if !alreadyTracked {
+		mod.DependsOn = append(mod.DependsOn, depPath)
+	}
+
+	// Capture position after adding direct dependency so cache only stores transitive deps
+	depsBefore := len(mod.DependsOn)
+
+	// Canonicalize path for cache key to ensure consistent lookups
+	// regardless of how the file is referenced (relative paths, symlinks, etc.)
+	cacheKey, err := filepath.EvalSymlinks(targetConfig)
+	if err != nil {
+		// If EvalSymlinks fails (e.g., broken symlink), fall back to Abs
+		cacheKey, err = filepath.Abs(targetConfig)
+		if err != nil {
+			cacheKey = targetConfig // final fallback to non-canonical path
+		}
+	}
+
+	// Check cache first to avoid re-parsing the same file
+	if mod.parseCache != nil {
+		if entry, ok := mod.parseCache.get(cacheKey); ok {
+			// Replay cached dependencies onto current module
+			for _, dep := range entry.dependencies {
+				alreadyHas := false
+				for _, existing := range mod.DependsOn {
+					if existing == dep {
+						alreadyHas = true
+						break
+					}
+				}
+				if !alreadyHas {
+					mod.DependsOn = append(mod.DependsOn, dep)
+				}
+			}
+			return entry.value, nil
+		}
+	}
 
 	// We update the ctx of terragruntOptions to the config being read in.
 	clonedOpts := ctx.TerragruntOptions.Clone()
@@ -180,7 +228,21 @@ func readTerragruntConfigImpl(ctx *tgconfig.ParsingContext, tgLogger log.Logger,
 		return cty.NilVal, err
 	}
 
-	return tgconfig.TerragruntConfigAsCty(cfg)
+	val, err := tgconfig.TerragruntConfigAsCty(cfg)
+	if err != nil {
+		return cty.NilVal, err
+	}
+
+	// Store in cache for future use - cache both value and newly discovered dependencies
+	if mod.parseCache != nil {
+		newDeps := mod.DependsOn[depsBefore:]
+		mod.parseCache.set(cacheKey, &tgCacheEntry{
+			value:        val,
+			dependencies: append(project.Paths{}, newDeps...),
+		})
+	}
+
+	return val, nil
 }
 
 // tgReadTFVarsFileFuncImpl reads a *.tfvars or *.tfvars.json file and returns the contents as a JSON encoded string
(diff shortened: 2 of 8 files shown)


mcalhoun:
* **#2228** <a href="https://app.graphite.dev/github/pr/terramate-io/terramate/2228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/terramate-io/terramate/2228?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

soerenmartius:
Really like this contribution @mcalhoun - what I noticed right away is the exceptional PR description which right away gave me a really decent understanding about the scope of this PR.

snakster:
What makes caching problematic is that we rely on the hooks being executed during evaluation to update `DependsOn`.

So if you cache a file like this which is included by multiple modules, only the first one will have a DependsOn containing all transitive dependencies.

Here is a test case to reproduce this:
```
func TestTransitiveDependencies(t *testing.T) {
	t.Parallel()
	type want struct {
		err     error
		modules tg.Modules
	}
	type testcase struct {
		name       string
		layout     []string
		basedir    project.Path
		ignoreDeps bool
		want       want
	}

	for _, tc := range []testcase{
		{
			name: "transitive dependencies",
			layout: []string{
				`f:common/shared.hcl:` + Doc(
					Expr("inputs",
						`{
  						data = jsondecode(read_tfvars_file("${get_repo_root()}/common/shared.tfvars"))
					}`),
				).String(),
				`f:common/shared.tfvars:`,
				`f:common/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/common"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("shared.hcl")`),
					),
				).String(),
				`f:mod1/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/mod1"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("${get_repo_root()}/common/shared.hcl")`),
					),
				).String(),
				`f:mod2/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/mod2"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("${get_repo_root()}/common/shared.hcl")`),
					),
				).String(),
			},
			want: want{
				modules: tg.Modules{
					{
						Path:       project.NewPath("/common"),
						Source:     "https://some.etc/common",
						ConfigFile: project.NewPath("/common/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/common/shared.hcl"),
							project.NewPath("/common/shared.tfvars"),
						},
					},
					{
						Path:       project.NewPath("/mod1"),
						Source:     "https://some.etc/mod1",
						ConfigFile: project.NewPath("/mod1/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/common/shared.hcl"),
							project.NewPath("/common/shared.tfvars"),
						},
					},
					{
						Path:       project.NewPath("/mod2"),
						Source:     "https://some.etc/mod2",
						ConfigFile: project.NewPath("/mod2/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/common/shared.hcl"),
							project.NewPath("/common/shared.tfvars"),
						},
					},
				},
			},
		},
	} {
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			s := sandbox.New(t)
			s.BuildTree(tc.layout)
			basedir := tc.basedir
			if basedir.String() == "" {
				basedir = project.NewPath("/")
			}
			modules, err := tg.ScanModules(s.RootDir(), basedir, !tc.ignoreDeps)
			errtest.Assert(t, err, tc.want.err)
			if err != nil {
				return
			}
			if diff := cmp.Diff(modules, tc.want.modules,
				cmpopts.IgnoreFields(tg.Module{}, "FilesProcessed"),
				cmpopts.EquateComparable(project.Path{}), cmpopts.IgnoreUnexported(tg.Module{})); diff != "" {
				t.Errorf("Diff (want [+], got [-]): %s", diff)
			}
		})
	}
}
```
I don't know if this use case actually makes sense, but with caching, it will miss `/common/shared.tfvars` for `/mod1` and `/mod2`. On `main`, it will pass. I think it would be possible to cache the dependencies as well, but it will be tricky.

Edit: Or maybe it won't be tricky. I tried this before: https://github.com/terramate-io/terramate/compare/main...snk-tg-cache
It didn't lead to any speedup for the customer. Maybe you could do something similar here, store the `DependsOn` with each entry.

Another question: Can we assume that the result of `read_terragrunt_config(x)` is not affected by any other side effects? i.e. any variables in the included file that would get different values depending on the calling context.

mcalhoun:
Good catch @snakster! You're right - the current implementation misses transitive dependencies because we skip hook execution on cache hits. I'll see if I can update it to cache both the parsed value and the dependencies discovered during parsing, then replay those dependencies when using cached results. That should handle the test case you provided while keeping the performance benefits.

snakster:
This is a test for the second case I mentioned:
```
		{
			name: "dynamic dependencies",
			layout: []string{
				`f:base0/common/shared.hcl:` + Doc(
					Expr("inputs",
						`{
  						data = jsondecode(read_tfvars_file(find_in_parent_folders("shared.tfvars")))
					}`),
				).String(),
				`f:base0/common/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/common"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("shared.hcl")`),
					),
				).String(),
				`f:base1/mod1/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/mod1"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("${get_repo_root()}/base0/common/shared.hcl")`),
					),
				).String(),
				`f:base2/mod2/terragrunt.hcl:` + Doc(
					Block("terraform",
						Str("source", "https://some.etc/mod2"),
					),
					Block("locals",
						Expr("a", `read_terragrunt_config("${get_repo_root()}/base0/common/shared.hcl")`),
					),
				).String(),
				`f:base0/shared.tfvars:`,
				`f:base1/shared.tfvars:`,
				`f:base2/shared.tfvars:`,
			},
			want: want{
				modules: tg.Modules{
					{
						Path:       project.NewPath("/base0/common"),
						Source:     "https://some.etc/common",
						ConfigFile: project.NewPath("/base0/common/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/base0/common/shared.hcl"),
							project.NewPath("/base0/shared.tfvars"),
						},
					},
					{
						Path:       project.NewPath("/base1/mod1"),
						Source:     "https://some.etc/mod1",
						ConfigFile: project.NewPath("/base1/mod1/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/base0/common/shared.hcl"),
							project.NewPath("/base1/shared.tfvars"),
						},
					},
					{
						Path:       project.NewPath("/base2/mod2"),
						Source:     "https://some.etc/mod2",
						ConfigFile: project.NewPath("/base2/mod2/terragrunt.hcl"),
						DependsOn: project.Paths{
							project.NewPath("/base0/common/shared.hcl"),
							project.NewPath("/base2/shared.tfvars"),
						},
					},
				},
			},
		},
```
Even if the dependencies are cached as well, I assume it would give the wrong ones.

mcalhoun:
@snakster I think the scenario you brought up is now addressed (and I added a test for it).

snakster:
> @snakster I think the scenario you brought up is now addressed (and I added a test for it).

👍 

But also have a look at the second one, I could run it now that the dependencies are cached, and as suspected it will return 
/base0/shared.tfvars for both modules instead of /baseX/shared.tfvars.

There could be other cases where the result of read_terragrunt_config() is somehow affected by the calling context, so I am not sure if caching is possible.

snakster:
@mcalhoun I've set this PR to draft for now until we figure out how to make it work.

PR #2243, opened 2025-11-05
feat: Add drifts cloud logs sync to TMC and state flow now includes running

sixstone-qq:

This PR makes the following two changes:

* Move the drifts creation flow to a stateful system where the drift check run is running and then it has a status
  compared to before where it was a fire and forget system.
* Sync the drift check command run logs (stdout and stderr) pushed to the cloud.

  (It still has the limitation when the command does not exist, logs are not synchornised)

It is a long outstanding feature for TMC users.

```
no
```

---

> [!NOTE]
> Move drift sync to a stateful v2 flow (start → running → update) with command log streaming; update client/engine/testserver and tests accordingly.
> 
> - **Cloud API (v2)**:
>   - Add drift lifecycle: `POST /v2/drifts/:org` to start (returns `uuid`), `PATCH /v2/drifts/:org/:uuid` to update status/changeset, and `POST/GET /v2/drifts/:org/:uuid/logs` for logs.
>   - Keep legacy drift details on `GET /v1/drifts/...`; add `DriftsV1Path`.
>   - Introduce `drift.Running` status; extend filters/serialization; add new resources: `DriftCheckRunStart*`, `UpdateDriftPayloadRequest`, `DriftWithStack`, `DriftsWithStacks`.
> - **Client/SDK**:
>   - `CreateStackDrift` now returns `DriftCheckRunStartResponse` (uuid); add `UpdateStackDrift`.
>   - Generalize log sync via `cloud.Entity` (deployment/preview/drift) in `SyncCommandLogs`.
>   - Adjust endpoints to `/v2/drifts`; keep `/v1/drifts` for details.
> - **Engine/Cloudsync**:
>   - Start drift run in `BeforeRun`; finalize with status/changeset in `AfterRun`.
>   - Send command logs for drift runs; update `Logs` to receive task and route logs by entity.
>   - Track drift UUID per stack; make run state concurrent-safe.
> - **Test Server & E2E**:
>   - Implement new v2 drift routes, store drifts by UUID, persist/retrieve logs; update drift state mapping and sorting.
>   - Update end-to-end tests to assert new flow, logs, and `running` state.
> - **Docs/Changelog**:
>   - Update CHANGELOG: note protocol change (start/end) and new drift log sync.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e4edc56402df24a62db1dfbe7c89674c9d185802. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/cloudsync/state.go
+++ b/cloudsync/state.go
@@ -5,6 +5,7 @@ package cloudsync
 
 import (
 	"strings"
+	"sync"
 
 	"github.com/terramate-io/terramate/cloud/api/resources"
 	"github.com/terramate-io/terramate/errors"
@@ -21,13 +22,15 @@ const (
 
 // CloudRunState represents the state of the current run.
 type CloudRunState struct {
+	sync.RWMutex
+
 	RunUUID resources.UUID
 
-	StackMeta2ID map[string]int64
-	// StackMeta2PreviewIDs is a map of stack.ID to stackPreview.ID
-	StackMeta2PreviewIDs map[string]string
-	// StackMeta2DriftUUIDs is a map of stack.ID to drift.UUID
-	StackMeta2DriftUUIDs map[string]resources.UUID
+	stackMeta2ID map[string]int64
+	// stackMeta2PreviewIDs is a map of stack.ID to stackPreview.ID
+	stackMeta2PreviewIDs map[string]string
+	// stackMeta2DriftUUIDs is a map of stack.ID to drift.UUID
+	stackMeta2DriftUUIDs map[string]resources.UUID
 	ReviewRequest        *resources.ReviewRequest
 	RREvent              struct {
 		PushedAt  *int64
@@ -38,43 +41,55 @@ type CloudRunState struct {
 
 // SetMeta2CloudID sets the cloud ID of a stack given its metadata ID.
 func (rs *CloudRunState) SetMeta2CloudID(metaID string, id int64) {
-	if rs.StackMeta2ID == nil {
-		rs.StackMeta2ID = make(map[string]int64)
+	rs.Lock()
+	defer rs.Unlock()
+	if rs.stackMeta2ID == nil {
+		rs.stackMeta2ID = make(map[string]int64)
 	}
-	rs.StackMeta2ID[strings.ToLower(metaID)] = id
+	rs.stackMeta2ID[strings.ToLower(metaID)] = id
 }
 
 // StackCloudID returns the cloud ID of a stack given its metadata ID.
-func (rs CloudRunState) StackCloudID(metaID string) (int64, bool) {
-	id, ok := rs.StackMeta2ID[strings.ToLower(metaID)]
+func (rs *CloudRunState) StackCloudID(metaID string) (int64, bool) {
+	rs.RLock()
+	defer rs.RUnlock()
+	id, ok := rs.stackMeta2ID[strings.ToLower(metaID)]
 	return id, ok
 }
 
 // SetMeta2PreviewID sets the cloud preview ID of a stack given its metadata ID.
 func (rs *CloudRunState) SetMeta2PreviewID(metaID string, previewID string) {
-	if rs.StackMeta2PreviewIDs == nil {
-		rs.StackMeta2PreviewIDs = make(map[string]string)
+	rs.Lock()
+	defer rs.Unlock()
+	if rs.stackMeta2PreviewIDs == nil {
+		rs.stackMeta2PreviewIDs = make(map[string]string)
 	}
-	rs.StackMeta2PreviewIDs[strings.ToLower(metaID)] = previewID
+	rs.stackMeta2PreviewIDs[strings.ToLower(metaID)] = previewID
 }
 
 // CloudPreviewID returns the cloud preview ID of a stack given its metadata ID.
-func (rs CloudRunState) CloudPreviewID(metaID string) (string, bool) {
-	id, ok := rs.StackMeta2PreviewIDs[strings.ToLower(metaID)]
+func (rs *CloudRunState) CloudPreviewID(metaID string) (string, bool) {
+	rs.RLock()
+	defer rs.RUnlock()
+	id, ok := rs.stackMeta2PreviewIDs[strings.ToLower(metaID)]
 	return id, ok
 }
 
 // SetMeta2DriftUUID sets the cloud drift UUID of a stack given its metadata ID.
 func (rs *CloudRunState) SetMeta2DriftUUID(metaID string, driftUUID resources.UUID) {
-	if rs.StackMeta2DriftUUIDs == nil {
-		rs.StackMeta2DriftUUIDs = make(map[string]resources.UUID)
+	rs.Lock()
+	defer rs.Unlock()
+	if rs.stackMeta2DriftUUIDs == nil {
+		rs.stackMeta2DriftUUIDs = make(map[string]resources.UUID)
 	}
-	rs.StackMeta2DriftUUIDs[strings.ToLower(metaID)] = driftUUID
+	rs.stackMeta2DriftUUIDs[strings.ToLower(metaID)] = driftUUID
 }
 
 // CloudDriftUUID returns the cloud drift UUID of a stack given its metadata ID.
-func (rs CloudRunState) CloudDriftUUID(metaID string) (resources.UUID, bool) {
-	id, ok := rs.StackMeta2DriftUUIDs[strings.ToLower(metaID)]
+func (rs *CloudRunState) CloudDriftUUID(metaID string) (resources.UUID, bool) {
+	rs.RLock()
+	defer rs.RUnlock()
+	id, ok := rs.stackMeta2DriftUUIDs[strings.ToLower(metaID)]
 	return id, ok
 }
 


cloud/testserver/drifts.go
```diff
@@ -122,14 +133,82 @@ func PostDrift(store *cloudstore.Data, w http.ResponseWriter, r *http.Request, p
...
+		w.WriteHeader(http.StatusNotFound)
+		writeErr(w, fmt.Errorf("stack not found %s (%s) in org %s", driftCheck.StackMetaID, driftCheck.StackTarget, org.UUID))
+		return
+	}
+
+	var ok bool
+	st.State.Status, ok = stateTable()[driftCheck.Status][st.State.DeploymentStatus]
+	if !ok {
+		w.WriteHeader(http.StatusInternalServerError)
+		writeErr(w, errors.E("invalid stack status: %s %s", drift.Running.String(), st.State.DeploymentStatus.String()))
```
mcalhoun: the error message uses drift.Running.String() instead of driftCheck.Status.String()

cloud/api/resources/resources.go
```diff
@@ -750,32 +776,39 @@ func (ds Drifts) Validate() error {
...
 		}
 	}
-
-	if d.Details != nil {
-		err := d.Details.Validate()
-		if err != nil {
-			return err
-		}
+	if len(d.Command) == 0 {
+		return errors.E(`field "command" is required"`)
```
mcalhoun, line 808:
```suggestion
		return errors.E(`field "command" is required`)
```

cloud/testserver/drifts.go
```diff
@@ -122,14 +133,82 @@ func PostDrift(store *cloudstore.Data, w http.ResponseWriter, r *http.Request, p
...
+	err = payload.Validate()
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		writeErr(w, err)
+		return
+	}
+
+	driftCheck, updated := store.UpdateDrift(&org, resources.UUID(driftUUID), payload.Status, payload.Changeset, payload.UpdatedAt)
+	if !updated {
+		w.WriteHeader(http.StatusNotFound)
```
mcalhoun: Stack not found returns StatusNotFound; in PostDrift (line 96) a similar case returns StatusInternalServerError. Prefer StatusInternalServerError for stack lookup failures (data consistency issue).

cloudsync/create_drift.go
```diff
@@ -16,36 +17,105 @@ import (
...
+		if errors.IsAnyKind(err, engine.ErrRunCommandNotExecuted, engine.ErrRunFailed) {
+			doDriftBefore(e, run, state)
+			driftUUID, found = state.CloudDriftUUID(st.ID)
+		}
+
+		if !found {
+			logger.Error().Msg("missing drift UUID for stack ID")
+			return
+		}
+	}
```
mcalhoun, line 97: If doDriftBefore fails or returns an empty UUID, driftUUID could be empty.

cloud/api/resources/resources.go
```diff
@@ -938,6 +971,28 @@ func (r CreatePreviewResponse) Validate() error {
 	return errs.AsError()
 }
 
+// Validate validates the CreateDriftResponse payload
```
snakster:
```suggestion
// Validate validates the DriftCheckRunStartResponse payload
```

cloud/testserver/cloudstore/store.go
```diff
@@ -133,14 +133,21 @@ type (
 	// Drift model.
 	Drift struct {
 		ID          int64                         `json:"id"`
+		UUID        resources.UUID                `json:"uuid"`
 		StackMetaID string                        `json:"stack_meta_id"`
 		StackTarget string                        `json:"stack_target"`
 		Status      drift.Status                  `json:"status"`
-		Details     *resources.ChangesetDetails   `json:"details"`
+		Changeset   *resources.ChangesetDetails   `json:"details"`
```
snakster: json is still `details`. Let's keep them aligned?

cloud/testserver/cloudstore/store.go
```diff
@@ -547,11 +553,73 @@ func (d *Data) InsertDrift(orgID resources.UUID, drift Drift) (int, error) {
...
+
+	drift, found := org.Drifts[driftUUID]
+	if !found {
+		return errors.E(ErrNotExists, "drift uuid %s", driftUUID)
+	}
+
+	drift.State.Logs = append(drift.State.Logs, logs...)
+	d.Orgs[org.Name].Drifts[driftUUID] = drift
+	return nil
+}
```
mcalhoun, line 610: There's a potential race condition here.

The org is fetched outside the lock, then locked. Between these, the org could change. Fetch the org inside the lock or use a read lock, then upgrade. Current behavior may be acceptable if orgs rarely change; document if intentional.

How the org could change:

GetOrg():
Acquires a read lock (RLock)
Returns a clone of the org (org.Clone())
Releases the read lock

Gap: 
No lock is held
Another goroutine could modify d.Orgs
Lock():
Acquires a write lock

cloudsync/create_drift.go
```diff
@@ -16,36 +17,105 @@ import (
...
+		},
+		Metadata:  state.Metadata,
+		StartedAt: &startedAt,
+		Command:   run.Task.Cmd,
+	})
+
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to create drift")
+		e.DisableCloudFeatures(errors.E(errors.ErrInternal, clitest.CloudSyncDriftFailedMessage))
+		return
```
mcalhoun, line 63: doDriftBefore doesn't return an error, so callers can't detect failures. Consider returning an error or at least logging it clearly.

cloud/testserver/cloudstore/store.go
sixstone-qq: Thanks.
I think we should fix it in the whole test server as this pattern is wrong in other places. I think this pattern was used to speed up the org lookup. If we need to upgrade it anyway, the usage of a function is leaking the read lock.

Note: [RWMutex cannot be upgraded](https://pkg.go.dev/sync#RWMutex)

cloud/testserver/drifts.go
sixstone-qq: Fixed in fdfd2792

cloud/api/resources/resources.go
sixstone-qq: Fixed in fdfd2792

cloud/testserver/drifts.go
sixstone-qq: Fixed in fdfd2792

cloud/testserver/cloudstore/store.go
sixstone-qq: Fixed in fdfd2792

cloudsync/create_drift.go
sixstone-qq: That follows the existing pattern for all use cases, so I think it does not surprise the developer here. Do you have any proposal to log it more clear?

cloudsync/create_drift.go
sixstone-qq: yes, that's the intention on line 92 to check there is a driftUUID to use to update the status. I am not sure to follow you there.

cloudsync/cloud_hooks.go
```diff
@@ -49,18 +53,42 @@ func AfterRun(e *engine.Engine, run engine.StackCloudRun, state *CloudRunState,
...
+		entity = cloud.Entity{
+			Kind:     cloud.EntityKindPreview,
+			EntityID: stackPreviewID,
+		}
+	} else if task.CloudSyncDriftStatus {
+		driftUUID, _ := state.CloudDriftUUID(run.Stack.ID)
+		entity = cloud.Entity{
+			Kind:     cloud.EntityKindDrift,
+			EntityID: string(driftUUID),
+		}
```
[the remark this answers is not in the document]
sixstone-qq: Fixed in https://github.com/terramate-io/terramate/pull/2243/commits/cd6f38219eda84e4865411ffec1a1194a4320383

sixstone-qq:
cursor review

cloud/testserver/cloudstore/store.go
```diff
@@ -541,17 +552,82 @@ func (d *Data) FindDeploymentForCommit(orgID resources.UUID, commitSHA string) (
...
+		return errors.E(ErrNotExists, "org uuid %s", orgID)
+	}
+
+	drift, found := org.Drifts[driftUUID]
+	if !found {
+		return errors.E(ErrNotExists, "drift uuid %s", driftUUID)
+	}
+
+	drift.State.Logs = append(drift.State.Logs, logs...)
+	d.Orgs[org.Name].Drifts[driftUUID] = drift
```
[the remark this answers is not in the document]
sixstone-qq: That's not true as the lock was acquired before returning the copy so no other goroutine can modify the organization until the function returns.

cloud/testserver/stacks.go
```diff
@@ -29,6 +29,13 @@ func stateTable() map[drift.Status]map[deployment.Status]stack.Status {
 			deployment.Failed:   stack.Failed,
 			deployment.Canceled: stack.Failed,
 		},
+		drift.Running: {
+			deployment.OK:       stack.OK,
+			deployment.Pending:  stack.OK,
+			deployment.Running:  stack.OK,
+			deployment.Failed:   stack.Failed,
+			deployment.Canceled: stack.Failed,
+		},
```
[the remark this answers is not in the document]
sixstone-qq: it is out of the scope of this PR, furthermore, this is for testing server in e2etests

cloudsync/create_drift.go
```diff
@@ -16,36 +17,105 @@ import (
...
+	}
+
+	state.SetMeta2DriftUUID(st.ID, response.DriftUUID)
```
snakster, line 66: I suspect there is a race condition here for `--parallel`, when this happens:
```
if rs.StackMeta2DriftUUIDs == nil {
		rs.StackMeta2DriftUUIDs = make(map[string]resources.UUID)
	}
```
The before/after hooks are already run in parallel. For previews, the similar setup of the response map happens in the preparation step, which is not parallel yet.

CHANGELOG.md
```diff
@@ -20,6 +20,18 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 - Backward compatibility in versions `0.0.z` is **not guaranteed** when `z` is increased.
 - Backward compatibility in versions `0.y.z` is **not guaranteed** when `y` is increased.
 
+## Unreleased
```
snakster, line 23: It should be
```
- ## 0.15.0
+ ## Unreleased
```
below, and your new entries merged with the existing `0.15.0`, since `0.15.0` was not yet released.

This is a mistake in the current changelog.

cloud/api/resources/resources.go
sixstone-qq: Fixed in bd8e1750

cloud/testserver/cloudstore/store.go
sixstone-qq: I renamed the JSON field because it's not used as far as I see in fixtures for the cloud store.

Changed in bd8e1750

cloudsync/create_drift.go
sixstone-qq: Added a `RWMutex` in `cloudsync.State` to hold the map writes in e27ce084

CHANGELOG.md
sixstone-qq: Fixed in https://github.com/terramate-io/terramate/pull/2243/commits/64ffb33f602421038ece7fcff7adf40961225a9b

cloudsync/create_drift.go
mcalhoun: Sorry, I wasn't clear...

Since UUID is type UUID string, an empty UUID still makes found = true (the key exists). The check if !found passes, and the code proceeds. Was that intended?

cloudsync/create_drift.go
mcalhoun: If that's the existing pattern, then it's fine to keep it this way for now.

cloudsync/state.go
```diff
@@ -36,29 +41,55 @@ type CloudRunState struct {
...
+	defer rs.RUnlock()
+	id, ok := rs.stackMeta2ID[strings.ToLower(metaID)]
 	return id, ok
 }
 
 // SetMeta2PreviewID sets the cloud preview ID of a stack given its metadata ID.
 func (rs *CloudRunState) SetMeta2PreviewID(metaID string, previewID string) {
-	if rs.StackMeta2PreviewIDs == nil {
-		rs.StackMeta2PreviewIDs = make(map[string]string)
+	rs.Lock()
```
snakster, line 62: For the preview/deployment, the lock is not required, as you changed nothing there.
`SetMeta2PreviewID` and `SetMeta2CloudID` will be called before `RunAll()`.

Only for drifts, `SetMeta2DriftUUID` will be called during `RunAll` and it could be parallel there.

Did you add it to all for consistency?

cloudsync/state.go
```diff
@@ -36,32 +41,52 @@ type CloudRunState struct {
...
+	defer rs.Unlock()
 	if rs.StackMeta2PreviewIDs == nil {
 		rs.StackMeta2PreviewIDs = make(map[string]string)
 	}
 	rs.StackMeta2PreviewIDs[strings.ToLower(metaID)] = previewID
 }
 
 // CloudPreviewID returns the cloud preview ID of a stack given its metadata ID.
-func (rs CloudRunState) CloudPreviewID(metaID string) (string, bool) {
+func (rs *CloudRunState) CloudPreviewID(metaID string) (string, bool) {
```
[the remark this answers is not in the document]
snakster: This is wrong, the `SetMeta2PreviewID` will be called before, and then only the reads are parallel, which is safe.

cloudsync/state.go
```diff
@@ -36,32 +41,52 @@ type CloudRunState struct {
...
+	defer rs.Unlock()
 	if rs.StackMeta2ID == nil {
 		rs.StackMeta2ID = make(map[string]int64)
 	}
 	rs.StackMeta2ID[strings.ToLower(metaID)] = id
 }
 
 // StackCloudID returns the cloud ID of a stack given its metadata ID.
-func (rs CloudRunState) StackCloudID(metaID string) (int64, bool) {
+func (rs *CloudRunState) StackCloudID(metaID string) (int64, bool) {
```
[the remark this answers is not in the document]
snakster: Also wrong, see comment below.

cloudsync/state.go
sixstone-qq: Yes, in the future this may change so we could overlook this race condition in the future when protocol changes

cloudsync/create_drift.go
sixstone-qq: Fixed in 456e0f1c

cloudsync/state.go
snakster: Ok, personally I would leave it out, since I don't expect this to change, but it's just a read lock anyway.

CHANGELOG.md
```diff
@@ -20,15 +20,19 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
...
 
+- Change the protocol with Terramate Cloud to send drift check runs to
+  start and end instead of a single fire and forget operation.
 - (**BREAKING CHANGE**) Upgraded Terragrunt dependency from v0.55.21 to v0.82.0. This major version jump may introduce breaking changes in Terragrunt integration, particularly around configuration parsing and hook execution behavior.
 - Deprecated `--include-output-dependencies` and `--only-output-dependencies` flags in favor of the new, more comprehensive dependency management flags
 
 ### Added
 
+- Send command logs (stdout and stderr) to Terramate Cloud when
+  --sync-drift-status is set in terramate run and terramate script run commands.
```
snakster:
```suggestion
  `--sync-drift-status` is set in `terramate run` and `terramate script run` commands.
```
_nitpick_ We format like this in other places.

sixstone-qq: Thanks, fixed in 86e922b2

cloudsync/create_drift.go
```diff
@@ -16,36 +18,105 @@ import (
...
 
+	driftUUID, found := state.CloudDriftUUID(st.ID)
+	if !found || driftUUID == "" || driftUUID == resources.UUID(uuid.Nil.String()) {
+		// That could mean the doDriftBefore was never run as the command never ran.
+		// In this case we create the drift and we create the drift run and immediately failed.
+		if errors.IsAnyKind(err, engine.ErrRunCommandNotExecuted, engine.ErrRunFailed) {
+			doDriftBefore(e, run, state)
+			driftUUID, found = state.CloudDriftUUID(st.ID)
+		}
+
```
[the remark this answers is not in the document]
sixstone-qq: This was a product decision to not send drifts that were not even started.

PR #2230, opened 2025-10-13
feat: add new cli flags for including dependencies

mcalhoun:
## What this PR does / why we need it:

Adds comprehensive dependency filtering capabilities to the `list` command, allowing users to filter stacks based on their dependency relationships. This includes:

- New flags for dependency filtering:
    - `--include-dependencies`: Add stacks that selected stacks depend on
    - `--only-dependencies`: Replace selection with only dependencies
    - `--exclude-dependencies`: Remove dependencies from selection
    - `--include-all-dependents`: Add all stacks that depend on selected stacks
    - `--include-direct-dependents`: Add only direct dependents
    - `--only-direct-dependents`: Replace selection with only direct dependents
    - `--only-all-dependents`: Replace selection with only all dependents
    - `--exclude-dependents`: Remove all dependents from selection
- Support for both Terramate native dependencies and Terragrunt dependencies
- Comprehensive test coverage for all filtering options
- Backward compatibility with deprecated `--include-output-dependencies` and `--only-output-dependencies` flags

## Which issue(s) this PR fixes:

Fixes #

## Special notes for your reviewer:

This implementation builds a dependency graph from both Terramate native dependencies (stack.after, input.from_stack_id) and Terragrunt dependencies (dependencies.paths blocks). The graph is then used to apply various filtering options to the stack selection.

The implementation includes extensive tests for both Terramate native dependencies and Terragrunt dependencies, ensuring all filtering options work correctly.

## Does this PR introduce a user-facing change?

Yes, this PR adds new command-line flags for dependency filtering to the `list` command. The existing `--include-output-dependencies` and `--only-output-dependencies` flags are now deprecated but still functional for backward compatibility.

---

> [!NOTE]
> Adds graph-based dependency/dependent selection across list/run/script (Terragrunt + Terramate), deprecates legacy output-dependencies flags, and updates formatting, tests, and versioning.
>
> - **CLI**:
>     - Add new dependency selection flags to `list`, `run`, and `script run` (`--include/--only/--exclude` for both dependencies and dependents; direct/all variants).
>     - Keep `--include-output-dependencies`/`--only-output-dependencies` as deprecated aliases.
> - **Engine**:
>     - Introduce `DependencyFilters` and `ApplyDependencyFilters` with a graph built from data deps (`input.from_stack_id` and Terragrunt `dependency` blocks); exclude ordering-only deps (`after` / `dependencies.paths`).
[...]

PR #2246, opened 2025-11-12
refactor: don't use ScanModules for TG dependency tracking

snakster:

The flags introduced in #2230 use `tg.ScanModules()` to load the Terragrunt dependencies. Before, this function is used by `create --all-terragrunt`, but the Terragrunt loading that happens as part of change detection uses a different mechanism that is (1) more tightly integrated with the normal config loading and (2) parallelized.

As a result, the Terragrunt files were loaded twice when both these flags were used, thus taking about twice as long:
```
time terramate list --changed
13.92 real        18.19 user        74.27 sys

time terramate list --changed --include-all-dependencies
23.44 real        25.78 user        76.77 sys
```

The parallel loading was already the result of past initiatives to make the loading faster, so ideally, this is the method we want to use for the new flags.

The problem is that for `--changed`, this only works, because it a global flag, not a command flag (it should be a command flag - this was a mistake of the past). Normally, the config is already loaded before a command is selected, so by the time we know if this is a command that has the a dependency flag, it is too late.

The approach taken to fix this is to refactor the ordering of command selection and config loading.

The old method worked like this:
```
* beforeConfigHandler(): select and run commands that don't need the config
* <load the config>
* afterConfigHandler(): select and run commands that need the config
```

The new method works like this:
```
* select the command
* check requirements of the command, load the config if required etc.
* run the command
```

It is done in two commits:
* [refactor: command handler selection](https://github.com/terramate-io/terramate/commit/70ae2ce880dbcdf6287ed79a285c55975982a1c1) is the first step of implementing this new method, also cleaning up the CLI initialisation process a bit.
*  [refactor: don't use ScanModules for TG dependency tracking](https://github.com/terramate-io/terramate/commit/4be0d8205449d2a9a3ea0eeebdd93f44b606f1c2) then changes the commands that use the new dependency flags to trigger terragrunt dependency loading during config loading, and changing the dependency analysis code to make use of the previously loaded data instead of calling ScanModules().

Resulting time shows the performance improvement:
```
time terramate list --changed --include-all-dependencies
15.76 real        18.92 user        81.69 sys
```

I added some comments to help with the review.

```
no
```

---

> [!NOTE]
> Refactors CLI to select commands before loading config and introduces per-command requirements, while reworking dependency analysis to use preloaded Terragrunt data instead of rescanning.
> 
> - **CLI/Core**:
>   - Introduces `commands.Command` with `Requirements(ctx, cli)` and `Exec(ctx, cli)`; adds `commands.CLI` facade (version/product/io/engine/config access).
>   - Adds `EngineRequirement` (`RequireEngine`, `WithTerragrunt`, `WithExperiments`) to load engine/config conditionally per command.
>   - Changes execution flow: select command → satisfy requirements (load engine, experiments) → run.
> - **Engine/Config**:
>   - Plumbs `loadTerragruntModules` through project/config loaders; spins workers to load TG modules during config load.
>   - Rewrites `NewDependencyGraph` to consume TG `DependencyBlocks` from config and `input.from_stack_id`; removes `tg.ScanModules` usage.
>   - Adds `HasDependencyFilters`, refines `ApplyDependencyFilters`, cycle detection, and keeps ordering-only deps out of scope; exposes `CloudRegion`.
> - **Commands**:
>   - Migrates commands to new interfaces, acquiring context via `cli.Engine()/Printers()/WorkingDir()`; removes in-struct Engine/Printers wiring.
>   - Enforces experiments/terragrunt loading via `Requirements` (e.g., scripts, list/run when dependency flags used).
>   - Minor adjustments across cloud login/info, fmt, generate, clone, stack create/list, trigger, debug, experimental eval/rungraph/vendor.
> - **UI (tui)**:
>   - Replaces before/after handlers with `SelectCommand`; initializes logging/checkpoint, sets working dir, conditionally inits engine, sets analytics.
> - **Tests**:
>   - Adds/updates e2e and unit tests to assert dependency behavior (TG `dependencies.paths` do not widen scope) and new graph logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cfae79da6c6ac250f6fb5e89f7387dedc273e5c2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -20,7 +20,7 @@ type Command interface {
 	// Requirements is a generic interface to query a command for common requirements
 	// that will be fulfilled by the CLI before executing it.
 	//
-	// Examples for this include cloud setup, config loading, experiments etc.
+	// The result can either by a single requirement, or a RequirementsList.
 	Requirements(context.Context, CLI) any
 
 	// Exec executes the command.
@@ -48,8 +48,12 @@ type CLI interface {
 	Engine() *engine.Engine
 }
 
+// RequirementsList allows to return multiple requirements from Command.Requirements().
 type RequirementsList []any
 
+// HasRequirement checks if the given command has requirement of type T, and returns it if found.
+// If the command returns a single requirement, it will be checked against T.
+// If it returns a RequirementsList, each element will be checked.
 func HasRequirement[T any](ctx context.Context, cli CLI, cmd Command) (*T, bool) {
 	r := cmd.Requirements(ctx, cli)
 	switch r := r.(type) {
--- a/commands/requirements.go
+++ b/commands/requirements.go
@@ -3,10 +3,14 @@
 
 package commands
 
+// EngineRequirement is a requirement for commands that need the configuration to be loaded.
+// Commands that don't have this requirement set may not call CLI.Engine().
 type EngineRequirement struct {
-	LoadTerragrunt bool
+	LoadTerragruntModules bool
+	Experiments           []string
 }
 
+// RequireEngine creates a new EngineRequirement with the given options.
 func RequireEngine(opts ...RequireEngineOpt) *EngineRequirement {
 	req := &EngineRequirement{}
 	for _, opt := range opts {
@@ -15,18 +19,21 @@ func RequireEngine(opts ...RequireEngineOpt) *EngineRequirement {
 	return req
 }
 
+// RequireEngineOpt is the option type for RequireEngine().
 type RequireEngineOpt func(*EngineRequirement)
 
-func WithLoadTerragrunt() RequireEngineOpt {
+// WithTerragrunt an option for RequireEngine() to control loading of Terragrunt modules
+// for change detection and dependency tracking.
+func WithTerragrunt(loadModules bool) RequireEngineOpt {
 	return func(req *EngineRequirement) {
-		req.LoadTerragrunt = true
+		req.LoadTerragruntModules = loadModules
 	}
 }
 
-type ExperimentsRequirement struct {
-	Names []string
-}
-
-func RequireExperiments(names ...string) *ExperimentsRequirement {
-	return &ExperimentsRequirement{Names: names}
+// WithExperiments an option for RequireEngine() to require given experiments to be enabled.
+// This is part of the engine requirement, because experiments are specified in the config.
+func WithExperiments(names ...string) RequireEngineOpt {
+	return func(req *EngineRequirement) {
+		req.Experiments = names
+	}
 }
--- a/config/config.go
+++ b/config/config.go
@@ -47,8 +47,8 @@ type Root struct {
 
 	// hasTerragruntStacks tells if the repository has any Terragrunt stack.
 	hasTerragruntStacks *bool
-	// changeDetectionEnabled tells if change detection is enabled.
-	changeDetectionEnabled bool
+
+	loadTerragruntModules bool
 
 	maxTgWorkers int
 	tgTaskChan   chan *Tree
@@ -118,7 +118,7 @@ type List[T DirElem] []T
 // the config in fromdir and all parent directories until / is reached.
 // If the configuration is found, it returns the whole configuration tree,
 // configpath != "" and found as true.
-func TryLoadConfig(fromdir string, changeDetectionEnabled bool, hclOpts ...hcl.Option) (tree *Root, configpath string, found bool, err error) {
+func TryLoadConfig(fromdir string, loadTerragruntModules bool, hclOpts ...hcl.Option) (tree *Root, configpath string, found bool, err error) {
 	for {
 		parser, err := hcl.NewTerramateParser(fromdir, fromdir, hclOpts...)
 		if err != nil {
@@ -141,7 +141,7 @@ func TryLoadConfig(fromdir string, changeDetectionEnabled bool, hclOpts ...hcl.O
 			rootTree := NewTree(fromdir)
 			rootTree.Node = *cfg
 			root := NewRoot(rootTree, hclOpts...)
-			root.changeDetectionEnabled = changeDetectionEnabled
+			root.loadTerragruntModules = loadTerragruntModules
 			root.initTgWorkers()
 			err = root.loadTree(rootTree, fromdir, hclOpts...)
 			if err != nil {
@@ -174,7 +174,7 @@ func NewRoot(tree *Tree, hclOpts ...hcl.Option) *Root {
 }
 
 func (root *Root) initTgWorkers() {
-	if !root.changeDetectionEnabled {
+	if !root.loadTerragruntModules {
 		return
 	}
 	root.tgTransientErrs = map[string]error{}
@@ -208,7 +208,7 @@ func (root *Root) tgWorker() {
 }
 
 // LoadRoot loads the root configuration tree.
-func LoadRoot(rootdir string, changeDetectionEnabled bool, hclOpts ...hcl.Option) (*Root, error) {
+func LoadRoot(rootdir string, loadTerragruntModules bool, hclOpts ...hcl.Option) (*Root, error) {
 	rootcfg, err := hcl.ParseDir(rootdir, rootdir, hclOpts...)
 	if err != nil {
 		return nil, err
@@ -216,7 +216,7 @@ func LoadRoot(rootdir string, changeDetectionEnabled bool, hclOpts ...hcl.Option
 	rootTree := NewTree(rootdir)
 	rootTree.Node = *rootcfg
 	root := NewRoot(rootTree, hclOpts...)
-	root.changeDetectionEnabled = changeDetectionEnabled
+	root.loadTerragruntModules = loadTerragruntModules
 	root.initTgWorkers()
 	err = root.loadTree(rootTree, rootdir, hclOpts...)
 	if err != nil {
@@ -646,7 +646,7 @@ func (root *Root) loadTree(parentTree *Tree, cfgdir string, hclOpts ...hcl.Optio
 		parentTree.TgRootFile = filesResult.TgRootFile
 	}
 
-	if filesResult.TgRootFile != "" && root.changeDetectionEnabled {
+	if filesResult.TgRootFile != "" && root.loadTerragruntModules {
 		parentTree.terragruntModuleLoadFinished = make(chan struct{})
 		root.tgTaskChan <- parentTree
 	}
--- a/engine/dependencies.go
+++ b/engine/dependencies.go
@@ -7,17 +7,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/terramate-io/terramate/config"
 	"github.com/terramate-io/terramate/errors"
-	"github.com/terramate-io/terramate/project"
-	"github.com/terramate-io/terramate/tg"
 )
-
-// Modules is an alias for tg.Modules to avoid import cycles
-type Modules = tg.Modules
-
-// LoadTerragruntModules is a helper to load Terragrunt modules
-func LoadTerragruntModules(rootdir string, dir project.Path) (Modules, error) {
-	return tg.ScanModules(rootdir, dir, true)
-}
 
 // DependencyGraph represents the dependency relationships between stacks.
 type DependencyGraph struct {
@@ -35,7 +25,7 @@ type DependencyGraph struct {
 //
 // NOTE: stack.After and stack.Before are NOT included because they contain ordering-only
 // dependencies (e.g., Terragrunt dependencies.paths) which should NOT widen scope.
-func (e *Engine) NewDependencyGraph(stacks config.List[*config.SortableStack], tgModules tg.Modules, target string) (*DependencyGraph, error) {
+func (e *Engine) NewDependencyGraph(stacks config.List[*config.SortableStack], target string) (*DependencyGraph, error) {
 	logger := log.With().
 		Str("action", "engine.NewDependencyGraph()").
 		Logger()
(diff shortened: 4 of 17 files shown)


commands/clone/clone.go
```diff
@@ -18,47 +19,52 @@ import (
 
 // Spec is the command specification for the clone command.
 type Spec struct {
```
snakster, line 21: Just taking this an example:

Before, `afterConfigHandler` would set `Spec.Engine`, `WorkingDir` - all things that are available after loading the config.

The new method is to create Spec (which is the command handler) as early as possible, as soon as the command line arguments are parsed.  The remaining exported fields are either arguments, or fields only derived from arguments.

Fields like `engine` etc. will be set when `Exec` is run. They have been kept as members to keep the changes in the command function itself minimal.

This pattern of renaming `Engine, Printers, WorkingDir` (all non-argument fields) as unexported fields and setting them in Exec() is repeated for all commands.

The [SelectCommand](https://github.com/terramate-io/terramate/pull/2246/files#diff-b3af08c1aec5b8961d5a93f207b11b05476f1b0a116b20f3d64a8c6b5171c173R70) is where all these command handlers are created.

commands/commands.go
```diff
@@ -3,12 +3,66 @@
...
+// Command is an interface for a command handler.
+type Command interface {
 	// Name of the comamnd.
 	Name() string
+
+	// Requirements is a generic interface to query a command for common requirements
+	// that will be fulfilled by the CLI before executing it.
+	//
+	// The result can either by a single requirement, or a RequirementsList.
+	Requirements(context.Context, CLI) any
```
snakster, line 24: This indicates if a command requires the engine or not. This reason this generic signature was chosen is for extensibility in the pro version.

Previous `beforeConfigHandler` commands will return `nil`.
Previous `afterConfigHandler` commands will return `RequireEngine(...)`.

commands/commands.go
```diff
@@ -3,12 +3,66 @@
...
+
 	// Exec executes the command.
-	Exec(ctx context.Context) error
+	Exec(context.Context, CLI) error
+}
+
+// CLI is the interface for common CLI data required by commands.
+type CLI interface {
```
snakster, line 31: This is to break the dependency cycle between
```
ui/tui.CLI -> commands
ui/tui.CLI -> commands/<handlers>
commands/<handlers> -> commands
```
It's also nice to refine what information from the CLI is available to commands.

commands/commands.go
```diff
@@ -3,12 +3,66 @@
...
+	Engine() *engine.Engine
+}
+
+// RequirementsList allows to return multiple requirements from Command.Requirements().
+type RequirementsList []any
+
+// HasRequirement checks if the given command has requirement of type T, and returns it if found.
+// If the command returns a single requirement, it will be checked against T.
+// If it returns a RequirementsList, each element will be checked.
+func HasRequirement[T any](ctx context.Context, cli CLI, cmd Command) (*T, bool) {
```
snakster, line 57: It's a bit arcane, but I really wanted to avoid having to do
```
return []any{
   ...
}
```
in all `Requirements()` functions even if they all just have a single requirement for now.

commands/requirements.go
```diff
@@ -0,0 +1,39 @@
...
+// for change detection and dependency tracking.
+func WithTerragrunt(loadModules bool) RequireEngineOpt {
+	return func(req *EngineRequirement) {
+		req.LoadTerragruntModules = loadModules
+	}
+}
+
+// WithExperiments an option for RequireEngine() to require given experiments to be enabled.
+// This is part of the engine requirement, because experiments are specified in the config.
+func WithExperiments(names ...string) RequireEngineOpt {
```
snakster, line 35: This is not used for all experiment checks yet, just for scripts for now. But would be nice to use this generic approach instead of slightly different output everywhere.

config/config.go
```diff
@@ -47,8 +47,8 @@ type Root struct {
 
 	// hasTerragruntStacks tells if the repository has any Terragrunt stack.
 	hasTerragruntStacks *bool
-	// changeDetectionEnabled tells if change detection is enabled.
-	changeDetectionEnabled bool
+
+	loadTerragruntModules bool
```
snakster, line 51: Has been renamed, as this was really only used control loading of the Terragrunt modules on this level.

Before it was based on `parsedArgs.Changed`, now it is based on if the command requirements requested terragrunt loading.

engine/cloud.go
```diff
@@ -266,6 +265,15 @@ func (e *Engine) CloudOrgName() string {
 	return ""
 }
 
+// CloudRegion returns the cloud region from configuration, defaulting to EU.
```
snakster, line 269: This was moved here, as there was the `CloudOrgName()` in this file already, and these two functions should be close together. No idea why they were in different places before.

engine/dependencies.go
```diff
@@ -35,7 +25,7 @@ type DependencyGraph struct {
 //
 // NOTE: stack.After and stack.Before are NOT included because they contain ordering-only
 // dependencies (e.g., Terragrunt dependencies.paths) which should NOT widen scope.
-func (e *Engine) NewDependencyGraph(stacks config.List[*config.SortableStack], tgModules tg.Modules, target string) (*DependencyGraph, error) {
+func (e *Engine) NewDependencyGraph(stacks config.List[*config.SortableStack], target string) (*DependencyGraph, error) {
```
snakster, line 28: The terragrunt modules are no longer loaded and passed in separately. Instead, each stack config node may contain a Terragrunt module, accessed via `IsTerragruntModule`/`TerragruntModule`.

tg/tg_module.go
```diff
@@ -191,57 +189,56 @@ func LoadModule(rootdir string, dir project.Path, fname string, trackDependencie
...
-					Msg("dependency ConfigPath is not a string, skipping")
-				continue
-			}
-
-			depConfigPath := dep.ConfigPath.AsString()
-			depAbsPath := depConfigPath
-			if !filepath.IsAbs(depConfigPath) {
-				depAbsPath = filepath.Join(tgMod.Path, depConfigPath)
-			}
+	// "dependency" block is TerragruntDependencies
```
snakster, line 192: This one is a bit tricky... this `trackDependencies` was set to false for `--changed` Terragrunt loading, but true for the new dependency flags. So what if both are set? The new dependency flags already use `mod.DependencyBlocks` to not interfere with `mod.DependsOn`, which is used by `--changed`. We must make sure, that even when `trackDependencies` is false, `mod.DependencyBlocks` are still set.

So:
* Always parse DependencyBlock
* Always set mod.DependencyBlocks, but mod.DependsOn only if trackDependencies is set, to not break `--changed`

There are several existing tests for this, which is how I figured out the expected behaviour.

ui/tui/cli.go
```diff
@@ -164,6 +160,11 @@ func NewCLI(opts ...Option) (*CLI, error) {
 	c.printers.Stdout = printer.NewPrinter(c.state.stdout)
 	c.printers.Stderr = printer.NewPrinter(c.state.stderr)
 	c.state.uimode = engine.HumanMode
+
+	if val := os.Getenv("CI"); envVarIsSet(val) {
```
snakster, line 169: This was done somewhere later before, but it is just based on an env var, so lets do it where the value is initially set.

ui/tui/cli.go
```diff
@@ -189,15 +190,184 @@ func (c *CLI) Version() string { return c.version }
...
+// Stdout returns the stdout writer.
+func (c *CLI) Stdout() io.Writer { return c.state.stdout }
+
+// Stderr returns the stderr writer.
+func (c *CLI) Stderr() io.Writer { return c.state.stderr }
+
+// Stdin returns the stdout writer.
+func (c *CLI) Stdin() io.Reader { return c.state.stdin }
+
+func (c *CLI) initLogging(parsedArgs *FlagSpec) error {
```
snakster, line 222: I extracted the code from CLI.Exec(), beforeConfigHandler, afterConfigHandler into smaller functions with a clear responsiblity. It was a bit unstructured before and especially having init code in the before/after handlers is problematic as these are supposed to be extended by the pro version to add new commands.

ui/tui/cli.go
```diff
@@ -405,22 +559,6 @@ func (c *CLI) isTelemetryEnabled() bool {
 	return *cfg.Terramate.Config.Telemetry.Enabled
 }
 
-func (c *CLI) cloudOrgName() string {
```
snakster, line 421: This function seemed to exist twice? See Engine.CloudOrgName().

ui/tui/cli_handler.go
```diff
@@ -56,10 +46,6 @@ import (
 // ErrSetup is the error returned when the CLI fails to setup its initial values.
 const ErrSetup errors.Kind = "failed to setup Terramate"
 
-type handlerState struct {
```
snakster, line 59: This was created, tags were set, but then it was never used for anything. Probably a leftover, so I removed it.

ui/tui/cli_handler.go
```diff
@@ -80,193 +66,55 @@ func DefaultRootFlagHandlers() []RootFlagHandlers {
...
-				cliconfig.DirEnv,
-				cliconfig.Filename),
-				err)
-			return nil, false, false, errors.E(ErrSetup)
-
-		}
-		c.clicfg.UserTerramateDir = homeTmDir
-	}
+// SelectCommand selects the command handler and extracts the parameters.
+func SelectCommand(ctx context.Context, c *CLI, command string, flags any) (cmd commands.Command, err error) {
```
snakster, line 79: This is a single selection handler now for all commands, instead of having it split between beforeConfigHandler and afterConfigHandler.

After removing the command selection code here, and moving all the other init code to new functions, before/AfterConfig handler were empty, so they have been removed.

New hooks might be added if the pro version requires them, but for now what the pro version does is just add new commands, which is now covered by the selectCommandHandler.

snakster (commented):
Some remarks to help with the review 🦆

engine/dependencies.go
```diff
@@ -105,30 +95,33 @@ func (e *Engine) NewDependencyGraph(stacks config.List[*config.SortableStack], t
...
+				if !stackPaths[depStackPath] {
+					logger.Debug().
+						Str("stack", modStackPath).
+						Str("dependency", depStackPath).
+						Msg("dependency from Terragrunt dependency block not in current stack set, skipping")
+					continue
+				}
+
+				graph.addDependency(modStackPath, depStackPath)
+			}
```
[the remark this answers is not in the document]
snakster: Fixed in [6711da5](https://github.com/terramate-io/terramate/pull/2246/commits/6711da5c531fb3233bbbdeec0ef2345818bd566e)

commands/fmt/fmt.go
```diff
@@ -13,29 +13,43 @@ import (
...
+// Requirements returns the requirements of the command.
+func (s *Spec) Requirements(context.Context, commands.CLI) any { return commands.RequireEngine() }
+
 // Exec executes the fmt command.
-func (s *Spec) Exec(_ context.Context) error {
+func (s *Spec) Exec(_ context.Context, cli commands.CLI) error {
+	telemetry.DefaultRecord.Set(
+		telemetry.Command("fmt"),
+		telemetry.BoolFlag("detailed-exit-code", s.DetailedExitCode),
+	)
```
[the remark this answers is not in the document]
snakster: Fixed in [8a7c7be](https://github.com/terramate-io/terramate/pull/2246/commits/8a7c7beaad55c9af469d8d454c71edf6598fdaa2)

I was experimenting to put the telemetry in Exec, decided to leave it where it was, but forgot to remove this code again.

commands/fmt/fmt.go
```diff
@@ -13,29 +13,43 @@ import (
...
+
+	workingDir string
+	printers   printer.Printers
 }
 
 // Name returns the name of the command.
 func (s *Spec) Name() string { return "fmt" }
 
+// Requirements returns the requirements of the command.
+func (s *Spec) Requirements(context.Context, commands.CLI) any { return commands.RequireEngine() }
```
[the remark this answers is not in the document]
snakster: This may be true, but it works how it was before. Currently, `c.setWorkingDirectory` is only called when the engine requirement is set, maybe in the future it could be a separate requirement, but not for now.

ui/tui/cli_handler.go
```diff
@@ -390,46 +226,46 @@ func DefaultAfterConfigHandler(ctx context.Context, c *CLI) (commands.Executor,
...
-			Source:     parsedArgs.Experimental.Vendor.Download.Source,
-			Reference:  parsedArgs.Experimental.Vendor.Download.Reference,
-			Dir:        parsedArgs.Experimental.Vendor.Download.Dir,
-		}, true, false, nil
+			//WorkingDir: c.state.wd,
+			//Engine:     c.state.engine,
+			//Printers:   c.printers,
+			Source:    parsedArgs.Experimental.Vendor.Download.Source,
+			Reference: parsedArgs.Experimental.Vendor.Download.Reference,
+			Dir:       parsedArgs.Experimental.Vendor.Download.Dir,
```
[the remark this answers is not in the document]
snakster: Fixed in [5ab0eb7](https://github.com/terramate-io/terramate/pull/2246/commits/5ab0eb79f7f0c78f56c471b6fcf4b61264ff547d)

commands/requirements.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	return req
+}
+
+// RequireEngineOpt is the option type for RequireEngine().
+type RequireEngineOpt func(*EngineRequirement)
+
+// WithTerragrunt an option for RequireEngine() to control loading of Terragrunt modules
+// for change detection and dependency tracking.
+func WithTerragrunt(loadModules bool) RequireEngineOpt {
```
sixstone-qq, line 27: I don't see the need to pass the argument, it's confusing to use `WithTerragrunt(false)` to me, am I missing any use case?

snakster: I wrote it like this in the beginning, but then I had this case:
```
func (s *Spec) Requirements(context.Context, commands.CLI) any {
	return commands.RequireEngine(
		commands.WithTerragrunt(s.GitFilter.IsChanged || s.HasDependencyFilters()),
	)
}
```
which I preferred over building the option list based on conditions.

PR #2253, opened 2025-11-26
refactor: add post engine loading hook point

snakster:

The refactor done as part of #2246 moved command selection to a function that specifically does that, and removed the afterConfigHandler hook point. This hook point was used by catalyst for both adding new commands, but also to run general post config loading code.

This PR re-introduces a hook point to run non-command specific code after the engine was loaded.

It also fixes direct type cast of (*FlagSpec), instead of using AsFlagSpec, so the new code will work with catalyst. 

```
no
```

---

> [!NOTE]
> Introduces post-init engine hooks executed after engine initialization and replaces direct FlagSpec cast with AsFlagSpec safety; exposes new WithPostInitEngineHooks option.
> 
> - **CLI (`ui/tui/cli.go`, `ui/tui/cli_options.go`)**:
>   - **Hooks**:
>     - Add `postInitEngineHooks []PostInitEngineHook` and type `PostInitEngineHook`.
>     - Execute hooks after engine initialization and bindings setup in `Exec()`.
>     - Add option `WithPostInitEngineHooks(...)` to register hooks.
>   - **Flag parsing**:
>     - Replace direct `(*FlagSpec)` cast with `AsFlagSpec[FlagSpec](...)` and nil check.
>   - **Docs**:
>     - Update `WithAfterConfigSetup` comment to note availability relative to post-init hooks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d7e2548ef95db095d56253c6abd80ded56f520ec. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ui/tui/cli.go
+++ b/ui/tui/cli.go
@@ -456,7 +456,11 @@ func (c *CLI) Exec(args []string) {
 		}
 	}
 
-	parsedArgs := c.input.(*FlagSpec)
+	parsedArgs := AsFlagSpec[FlagSpec](c.input)
+	if parsedArgs == nil {
+		panic(errors.E(errors.ErrInternal, "please report this as a bug"))
+	}
+
 	migrateFlagAliases(parsedArgs)
 
 	// profiler is only started if Terramate is built with -tags profiler


ui/tui/cli_options.go
```diff
@@ -159,10 +159,18 @@ func WithBeforeConfigSetup(handlers ...BindingsSetupHandler) Option {
 }
 
 // WithAfterConfigSetup is an option to setup handlers that run after the config is loaded.
-// Bindings set here will available in afterConfigHandlers and during command execution.
+// Bindings set here will available in postInitEngineHooks and during command execution.
```
sixstone-qq, line 162: I feel this is not clear to me. Why is this option available?

snakster: It is not used here, but provides a method for catalyst to inject behaviour like yaml config loading, or virtual stack metadata.

PR #2254, opened 2025-11-28
fix: tokenizing tm_slug(null) in collection types

snakster:

The change done in #2206 will cause an expression with any collection type (object, map, list, tuple) that contains an unknown value to be tokenized as `null`. This affects more than just `tm_slug(null) -> null`.

The PR relaxes this to unknown values themselves rather than their containing expression and also refactors the respective tests so that they test whats closer to the actual usage scenario.

```
yes, see changelog
```

---

> [!NOTE]
> Ensure `tm_slug(null)` no longer turns enclosing HCL collections into `null` during codegen by treating only unknown values as `null`, and update tests accordingly.
> 
> - **HCL AST/tokenization**:
>   - Use `val.IsKnown()` instead of `IsWhollyKnown()` to detect unknowns and render only unknown values as `null`.
>   - Preserve collection structures when elements are dynamic/null (e.g., `{a="hello", b=tm_slug(null)}` tokenizes to `b=null`).
> - **Stdlib tests (`tm_slug`)**:
>   - Refactor eval tests to assert `cty.Value` equality and add cases for `tm_slug(null)` and within objects.
>   - Add tokenization assertion showing objects with dynamic/null members render correctly.
> - **Changelog**:
>   - Document fix for `tm_slug(null)` no longer turning containing HCL collection types to `null` during code generation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cb8675fa83bb0a49460b4ff4c0e4d34e6a246fbf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ Given a version number `MAJOR.MINOR.PATCH`, we increment the:
 
 - Add support for multi-project workspaces to Terramate language server.
 
+### Fixed
+
+- Fix `tm_slug(null)` expressions turning their containing HCL collection types to `null` during code generation.
+
 ## 0.15.0
 
 ### Changed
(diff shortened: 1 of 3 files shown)


mcalhoun (approved):
LGTM, thanks!

PR #2291, opened 2026-02-26
fix: Recognize generated files converted to CRLF

### LINKED ISSUE #2290: [BUG] Terramate fails to recognize autogenerated files when line endings are converted into CRLF
**Describe the bug**
Terramate fails to recognize its own automatically generated files, if a generated file's line endings are converted into CRLF.

An error message like this is thrown when trying to run `terramate generate`:

```
C:\Users\PervonZweigbergk\a\terramate-crlf-bug-demo>terramate generate
Code generation report

Failures:

- /
        error: manually defined code found: saving file "generated_file.tf": check file "C:\\Users\\PervonZweigbergk\\a\\terramate-crlf-bug-demo\\generated_file.tf"

Hint: '+', '~' and '-' mean the file was created, changed and deleted, respectively.

C:\Users\PervonZweigbergk\a\terramate-crlf-bug-demo>
```

In this case this is the contents of generated_file.tf:

```
PS C:\Users\PervonZweigbergk\a\terramate-crlf-bug-demo> format-hex .\generated_file.tf


           Path: C:\Users\PervonZweigbergk\a\terramate-crlf-bug-demo\generated_file.tf

           00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F

00000000   2F 2F 20 54 45 52 52 41 4D 41 54 45 3A 20 47 45  // TERRAMATE: GE
00000010   4E 45 52 41 54 45 44 20 41 55 54 4F 4D 41 54 49  NERATED AUTOMATI
00000020   43 41 4C 4C 59 20 44 4F 20 4E 4F 54 20 45 44 49  CALLY DO NOT EDI
00000030   54 0D 0A 0D 0A 6C 6F 63 61 6C 73 20 7B 0D 0A 20  T....locals {..
00000040   20 67 72 65 65 74 69 6E 67 20 3D 20 22 68 65 6C   greeting = "hel
00000050   6C 6F 2C 20 77 6F 72 6C 64 21 22 0D 0A 7D 0D 0A  lo, world!"..}..
```

Notice that the line breaks after the header are 0D 0A 0D 0A (CR LF CR LF).

This can happen in the following scenario:

- A project using Terramate has opted to check in certain types of generated code into a git repository for whatever reason (as opposed to just generating the code fresh locally.)
- Git is running on Windows with the setting core.autocrlf=true (default on Windows version of Git)
- This setting is not overridden by `.gitattributes` or any other mechanism similar in the repository.

**To Reproduce**

1. Run on Windows (or another hypotetical OS broken enough to use CRLF line endings)
2. Run `git config core.autocrlf` to ensure the setting is set to "true" (the default on Windows). If not, temporarilly change it with `git config core.autocrlf true` (remember what it was set to before so you can change it back.
3. Clone my demo repository with git clone `https://github.com/pv2b/terramate-crlf-bug-demo` (this project does not contain any .gitattributes, and does contain the generated code)
4. Run `terramate generate` and observe the error.

Alternatively, the error should be reproducible on any platform by using a text editor or other tool to change the line endings of the file to CRLF. (The above case is just the typical path where a user might likely run into this unintentionally.)

 - Windows 11 21H2 (22000.3260)
 - git version 2.35.1.windows.2
 - Terramate Version 0.15.5

**Additional context**
Possible workarounds:

- Remove generated files and re-run terramate generate to regenerate them.
- At the individual user level: Make sure to `set git config core.autocrlf input`. This will keep your checked out files in whatever line ending format they were in the repo (LF hopefully), while still making sure to convert CRLF to LF when committing.
- At the repository level, use .gitattributes to tell git to always use LF linebreaks for *.tf files with this example (assuming *.tf is the extension you use for generated files) - this is the solution I ended up going with for my own project.
```
*.tf text eol=lf
```

The bug appears to be here:

https://github.com/terramate-io/terramate/blob/dad99f01b050d7618d0fba8283ecd7ede53cdac8/generate/generate.go#L1271-L1280

The above code does a string.HasPrefix on the contents of the file (which doesn't try to do any CRLF handling against strings defined in these two places:

https://github.com/terramate-io/terramate/blob/dad99f01b050d7618d0fba8283ecd7ede53cdac8/generate/genhcl/genhcl.go#L62-L63

https://github.com/terramate-io/terramate/blob/dad99f01b050d7618d0fba8283ecd7ede53cdac8/generate/genhcl/genhcl.go#L149-L152

For files with the legacy v0 header, this problem should not occur because the prefix check doesn't have a newline as part of the prefix.

However, for the more modern genhcl.Header style, notice that there is a hardcoded \n\n in there which assumes the file always has LF file endings.

As for a suggested fix, I would say probably the most straight forward version is to add a third check, converting the desired prefix to CRLF format. (This should be faster than converting the whole file.)

I've submitted a PR for with a fix for this.

pv2b:

terramate generate would refuse to overwrite files generated by terramate, incorrectly flagging them as manually generated files, if the file had been converted from LF to CRLF linebreaks.

This fix will allow terramate to still recognize its own headers even if the linebreaks have been mangled.

Fixes #2290

First time programming in Go, so I've tried to keep the code as boring/simple as possible

```
no
```

---

> [!NOTE]
> **Low Risk**
> Low risk: a small change to generated-header prefix detection plus a regression test; the main risk is only in potentially misclassifying a file as generated, but matching is still limited to Terramate’s known headers.
> 
> **Overview**
> Prevents `terramate generate` from treating previously generated files as “manual code” when those files had their line endings converted to CRLF (common on Windows checkouts).
> 
> Updates `hasGenHCLHeader()` to also accept CRLF-normalized variants of the known Terramate headers, and adds a regression test ensuring generation succeeds (overwrites) when an existing generated Terraform file contains a CRLF header.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0cae42835f846e90a9b8c8fa3587ff86404c7d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -1275,6 +1275,16 @@ func hasGenHCLHeader(commentStyle genhcl.CommentStyle, code string) bool {
 		if strings.HasPrefix(code, header) {
 			return true
 		}
+		// Make sure to recognize headers also if the generated file has been converted
+		// to CRLF headers. Terramate itself never generates a file with CRLF headers, but
+		// this can happen if a tool, such as Git on Windows, converts the file endings
+		// to CRLF on checkout, e.g. if core.autocrlf.input is set to true, which is the
+		// default on Windows.
+		// Fixes bug: https://github.com/terramate-io/terramate/issues/2290
+		header_crlf := strings.ReplaceAll(header, "\n", "\r\n")
+		if strings.HasPrefix(code, header_crlf) {
+			return true
+		}
 	}
 	return false
 }
--- a/generate/generate_hcl_test.go
+++ b/generate/generate_hcl_test.go
@@ -2121,6 +2121,35 @@ func TestWontOverwriteManuallyDefinedTerraform(t *testing.T) {
 	assert.EqualStrings(t, manualTfCode, actualTfCode, "tf code altered by generate")
 }
 
+func TestWillOverwriteGeneratedTerraformConvertedToCrlf(t *testing.T) {
+	t.Parallel()
+
+	const (
+		genFilename  = "test.tf"
+		generatedTfCode = "// TERRAMATE: GENERATED AUTOMATICALLY DO NOT EDIT\r\n\r\nlocals {}\r\n"
+	)
+
+	generateHCLConfig := GenerateHCL(
+		Labels(genFilename),
+		Content(
+			Terraform(
+				Str("required_version", "1.11"),
+			),
+		),
+	)
+
+	s := sandbox.NoGit(t, true)
+	s.BuildTree([]string{
+		fmt.Sprintf("f:%s:%s", terramate.DefaultFilename, generateHCLConfig.String()),
+		"s:stack",
+		fmt.Sprintf("f:stack/%s:%s", genFilename, generatedTfCode),
+	})
+
+	generateAPI := newGenerateAPIForTest(t)
+	report := generateAPI.Do(s.Config(), project.NewPath("/"), 0, project.NewPath("/modules"), nil)
+	assert.EqualInts(t, 1, len(report.Successes), "want success")
+}
+
 func TestGenerateHCLStackFilters(t *testing.T) {
 	t.Parallel()
 


mariux:
Thanks for reporting the bug in #2290 and for putting together this fix, @pv2b — especially as a first-time Go contribution. We appreciate you taking the time.

We've recently updated our [contribution policy](https://github.com/terramate-io/terramate/blob/main/CONTRIBUTING.md) and are no longer accepting external code contributions to the core repository. We're sorry we didn't get back to you sooner on this PR.

We took a closer look at the CRLF issue while reviewing your fix and realized the problem runs deeper than the header detection. When `git autocrlf` converts line endings, the content comparison in our outdated detection (`DetectOutdated`) would also be affected — generated content uses `\n` but the file on disk has `\r\n`, causing files to always appear outdated even when nothing actually changed. This means the right fix is probably normalizing line endings when reading files from disk rather than patching individual checks. Your PR correctly identified the most visible symptom, but we'll need a broader fix to handle CRLF consistently across the generate and outdated detection paths.

We've updated #2290 with these findings. We can't commit to a timeline as we're currently prioritizing work for Cloud customers, which unfortunately means community-reported bugs take longer than we'd like — but this is a real issue and your analysis gave us a head start.

Happy to credit you as a co-author when we ship the fix — just let us know on #2290.

Thanks for your patience, and sorry for the delay.

